PR #29, opened 2018-10-23
Feature/mom6 ecbuild gv rm

aerorahul:
1. clean up ecbuild of SOCA
2. Remove dependency on SIS2, but keep sea-ice in the interface, until a strategy to "couple" models is ready at the OOPS level.

This pull request is part of the effort to enable building MOM and its dependencies as part of the soca-bundle.


guillaumevernieres (approved):
All good.

PR #43, opened 2019-02-27
Feature/godas d

guillaumevernieres:
Implementation of GODAS D. No comments on the norms please! Fixed in a different branch.


travissluka (changes requested):
file ./Utils/soca_omb_stats_mod.F90 is missing from repo

guillaumevernieres:
oops, sorry. Just pushed.

travissluka:
I'm seeing a weird pattern in the bkg error along a single stripe (look at cold tongue below the surface) when running **test_soca_3dvar**. It looks fine after **test_soca_bkgerr** though
 
bkgerr

guillaumevernieres:
Thanks for catching that @travissluka . I also see that the bkg error changes with the number of cpu's ... Interesting feature, but maybe not what we want! I'm looking into it.

travissluka:
that pattern is gone, but I'm seeing a difference between the **test_soca_bkgerr** (left) and 
**test_soca_3dvar** (right) @guillaumevernieres  is it supposed to be that way?
image

travissluka (approved):
seems to work now.

PR #45, opened 2019-03-06
switch to 75 lvl hybrid vertical coordinates for MOM

travissluka:
switched from the 63 level sigma configuration to the 75 level hybrid vertical configuration. Configuration in the vertical should now be identical to what EMC will be using for their 1/4 degree model. The OM4_05 MOM_input was used as a starting point, so we'll have to take a closer look at the actual parameters used whenever we get to doing real science 

New restart files were spun up from climatology (and no longer have tropical ice, or Indonesian tsunamis!)


guillaumevernieres (approved):
Slightly closer to science! Thanks @travissluka .

PR #46, opened 2019-03-11
Feature/ensvariance

guillaumevernieres:
This branch needs the soca-bundle branch feature/marine-cs-2019 to build. 
Addition of the possibility of reading the variance of an ensemble to initialize the "bkgerror" linear change of variable class, plus a few fixes in the model class.


aerorahul:
`src/Utils/soca_layer_remap_mod.F90` file is missing from `soca::feature/ensvariance`

aerorahul:
The following tests FAILED:
```
	  1 - soca_coding_norms (Failed)
	  3 - test_soca_forecast_mom6 (Failed)
	  7 - test_soca_enspert (Failed)
	  8 - test_soca_ensvariance (Failed)
	 13 - test_soca_model (SEGFAULT)
	 14 - test_soca_increment (SEGFAULT)
	 22 - test_soca_3dvarfgat_mom6 (Failed)
	 23 - test_soca_4dhybrid (Failed)
```
Ignore the norms test, but is this expected? If so, I will continue reviewing.

guillaumevernieres:
@aerorahul : Nope. The only test that fails for me is the coding norm.

guillaumevernieres:
@aerorahul : You need to use the branches specified in the soca-bundle.

aerorahul:
@guillaumevernieres yes. I am using the branches from the `soca-bundle::feature/marine-cs-2019`
I will do a clean clone, build, test and dig.

travissluka:
all tests pass for me... haven't had a chance to look through the code yet though

aerorahul:
I am still seeing failures on my Mac
```
The following tests FAILED:
	  1 - soca_coding_norms (Failed)
	  3 - test_soca_forecast_mom6 (Failed)
	  7 - test_soca_enspert (Failed)
	 13 - test_soca_model (Failed)
	 22 - test_soca_3dvarfgat_mom6 (Failed)
	 23 - test_soca_4dhybrid (Failed)
```
I re-ran `test_soca_model` and it runs to completion from the model perspective, but fails to cleanly exit, because of an MPI_Abort:
```
13: OOPS_STATS ------------------------- Timing Statistics -------------------------
13:
13: Run: Finishing oops::Test running test::Model<SOCA> with status = 0
13:
13: FATAL: MPP_OPEN: error in OPEN for /proc/self/status.
13:
13:
13: FATAL: MPP_OPEN: error in OPEN for /proc/self/status.
13:
13: --------------------------------------------------------------------------
13: MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
13: with errorcode 1.
13:
13: NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
13: You may or may not see output from other processes, depending on
13: exactly when Open MPI kills them.
13: --------------------------------------------------------------------------
2/2 Test #13: test_soca_model ..................***Failed   27.54 sec
```

My guess is, someplace there is not a clear destructor/finalize for MPI.

I am going to keep looking.

guillaumevernieres:
@aerorahul : Are the test Ok with develop?

aerorahul:
Yes, the same tests fail with `develop`.

src/Fields/soca_fields_mod.F90
```diff
@@ -645,17 +662,41 @@ subroutine read_file(fld, c_conf, vdate)
                   &fld%geom%ocean%lat(i,j))
           end do
        end do
+
+       ! Remap layers if needed
+       if (vert_remap) then
+          call initialize_remapping(remapCS,'PCM')
```
travissluka, line 668: I'm not entirely clear why vertical remapping is required on a file read.

Also, I'm sure the impact is minimal, but this should probably be set to the same scheme used in the MOM_input file. PCM is only 1st order accurate, PPM_H4 is used in the default GFDL configurations.

guillaumevernieres: All good points @travissluka . The vertical interpolation was implemented to allow for the computation of ensemble perturbations, where each members have different layer thicknesses. That's the only application that I had in mind, 1st order is probably good enough and more stable for that, why the choice. It wasn't meant as a way to restart the model with a different vertical geometry. Happy to pass the method through the config, but the remapping will eventually be implemented as a change of variable (and removed from read_file). It's not done yet because we are missing the implementation of he variable change in a few oops base classes. We'll tackle the issue after the code sprint.

travissluka: ok, sounds convincing enough

travissluka (approved):

PR #55, opened 2019-03-15
Improved analysis. 

guillaumevernieres:
Better bkgerror, gets rid of some of the questionable values in the 3dvar increment.


src/Transforms/Balance/soca_balance_mod.F90
```diff
@@ -84,6 +86,7 @@ subroutine soca_balance_setup(c_conf, self, traj)
              end if
           end do
           self%kst%jacobian(i,j,:) = jac(:)
+          self%kst%jacobian(i,j,1:self%kst%nlayers) =  0.0_kind_Real
```
aerorahul: replace kind_Real with kind_real

src/Covariance/soca_covariance_mod.F90
```diff
@@ -300,13 +301,15 @@ subroutine soca_2d_convol(dx, horiz_convol, geom)
 
     real(kind=kind_real), allocatable :: tmp_incr(:,:,:,:)
     ! Allocate unstructured tmp_increment and make copy of dx    
-    call soca_struct2unstruct(dx(:,:), geom, tmp_incr)
+    !call soca_struct2unstruct(dx(:,:), geom, tmp_incr)
```
aerorahul: please remove commented code.

src/Covariance/soca_covariance_mod.F90
```diff
@@ -300,13 +301,15 @@ subroutine soca_2d_convol(dx, horiz_convol, geom)
...
+    call geom%ocean%struct2unstruct(dx(:,:), tmp_incr)    
 
     ! Apply 2D convolution
     call horiz_convol%apply_nicas(tmp_incr)
 
-    ! copy unstructured tmp_incr to structured dx    
-    call soca_unstruct2struct(dx(:,:), geom, tmp_incr)
+    ! copy unstructured tmp_incr to structured dx
+    !call soca_unstruct2struct(dx(:,:), geom, tmp_incr)
```
aerorahul: same as above. also everywhere else in this PR.

aerorahul (changes requested):
The usual tests fail (for me on my Mac), so thats OK.
changes look good.
minor updates requested.

guillaumevernieres:
@aerorahul : Removed the commented out code. 
@travissluka , @aerorahul : Could you make sure the increment for the 3dvar is reasonable?

aerorahul:
looking at the 3dvar increment now.

aerorahul:
There is something bizarre in the maritime continent in the T, S increments. basically, looks fixed.

travissluka:
@aerorahul plots? everything looks far more reasonable on my runs. 

@guillaumevernieres I'm going to do a quick test with the hgodas SST/ADT obs , at smaller hz correlation scale, to see how it looks now

guillaumevernieres:
@travissluka : That would be interesting, just remember that the qc does not work with sst yet ...
@aerorahul : There are still issues, but it should not be as unrealistic as what you were getting with what's in develop.

aerorahul (approved):

travissluka:
note: i would highly recommend we get in the habit of doing a "squash and merge" for PRs like this (lots of commits but with very little code actually changed)

aerorahul:
@travissluka Yes. That has been usually true, except for trigger-happy committers.

travissluka (changes requested):
adt assimilation looks much better now.
BUT sst assimilation is no good. Looked fine before, but now SST obs have no impact due to a surface T bkg error of order 0.01C in the open ocean. Fix my SST DA !

guillaumevernieres:
@travissluka : The background error is coming from a 9 member ensemble of stand alone mom6, nudged to some climatological sst's. Is there a way to change the nudging time scale? If you use the sst obs file commented out in the yml, you will see a surface increment.

travissluka:
@guillaumevernieres could we just replace the T bkg err in your variance file with the bkg error that was generated with the old code using the godas method. I'll send some plots this weekend of what happens when I do that with the hgodas obs. It should be good enough that way to carry on and deal with doing things better after the sprint.

guillaumevernieres:
@travissluka  sure.

guillaumevernieres:
Closing this PR. I will submit a new one that addresses the same issues as well as the splitting of the bkgerr change of variable in several factories.

PR #60, opened 2019-03-18
Feature/cycle obs

guillaumevernieres:
This branch needs the bugfix/marine-altimeter branch of the ioda-converter repo to work.
The cycling of the 3dvar fgat works but is not part of the ctests. The rads adt obs are not in the git-lfs and need to be downloaded.


exp/cycle.py
```diff
@@ -5,7 +5,17 @@
 This script cycles mom6 through multiple DA windows.
 
 Example:
-     $  ./cycle.py --ic 2018071912 --window 24 --num_cycles 10
+
+     <---------------------- window_length (2 hours) ---------------------->
```
travissluka: "2 hours"?

exp/cycle.py
```diff
@@ -36,10 +46,18 @@
...
-        
-        # Prepare obs files
-        # TODO
 
+        # Get date for the middle of the window
+        ymdh_mid = ymdh + timedelta(hours=dt)
+        
+        # Prepare ADT observation files
+        adt_script = '../../bin/altimeter2ioda.py'
+        adt_path = '../../../../../../Data/nesdis/rads/'
```
travissluka: make the obs source path configurable

exp/cycle.py
```diff
@@ -36,10 +46,18 @@
...
+        mid_date = str(ymdh_mid.year).zfill(4)+str(ymdh_mid.month).zfill(2)+str(ymdh_mid.day).zfill(2)+str(ymdh_mid.hour).zfill(2)
+        command=adt_script+' --path '+adt_path+' --ic '+yyyy+mm+dd+hh+' --date '+mid_date+' --window '+str(dt)
+        print command
+        os.system(command)
```
travissluka, line 59: not at all important here, but it would be good to get in the habit of using 
`from __future__ import print_function` and the `print()` syntax, so scripts will work with python 3.x whenever people realize python 2.x is ancient

exp/cycle.py
```diff
@@ -36,10 +46,18 @@
...
-        # Prepare obs files
-        # TODO
 
+        # Get date for the middle of the window
+        ymdh_mid = ymdh + timedelta(hours=dt)
+        
+        # Prepare ADT observation files
+        adt_script = '../../bin/altimeter2ioda.py'
+        adt_path = '../../../../../../Data/nesdis/rads/'
+        mid_date = str(ymdh_mid.year).zfill(4)+str(ymdh_mid.month).zfill(2)+str(ymdh_mid.day).zfill(2)+str(ymdh_mid.hour).zfill(2)
```
travissluka:
```suggestion
        mid_date = ymdh_mid.strftime("%Y%m%d%H")
```

exp/cycle.py
```diff
@@ -36,10 +46,18 @@
...
+        command=adt_script+' --path '+adt_path+' --ic '+yyyy+mm+dd+hh+' --date '+mid_date+' --window '+str(dt)
+        print command
+        os.system(command)
+        
         # Run DA
         command='cycle.sh '+yyyy+' '+mm+' '+dd+' '+hh+' '+str(dt)
```
travissluka:
```suggestion
        command='./cycle.sh '+yyyy+' '+mm+' '+dd+' '+hh+' '+str(dt)
```

exp/cycle.sh
```diff
@@ -1,18 +1,18 @@
 #!/bin/sh
 
 # Set date for IC
-YYYY=$1 #2018
-MM=$2   #07
-DD=$3    #22
-HH=$4    #12 
+YYYY=$1  # 2018
```
aerorahul: I recommend we pass YYYYMMDDHH as the argument, instead of 4 arguments.
The string can be subset inside the script where necessary.

exp/cycle.sh
```diff
@@ -1,18 +1,18 @@
...
 WINDOW_LENGTH=PT$5H # DA window
 
 
 
 # Initialize B
-INIT_B=false 
+INIT_B=false
 if ( ${INIT_B} ) then
-   mpirun -np 4 --oversubscribe ../../bin/soca_staticbinit.x ./expinput/static_SocaError_init.yml
+   mpirun ../../bin/soca_staticbinit.x ./expinput/static_SocaError_init.yml
```
aerorahul: Relative paths are dangerous.  We should make a habit of copying/linking all necessary files into a run directory and locally execute operations.
Creating a run directory will also aid in debugging.

exp/cycle.sh
aerorahul: e.g. CDATE=YYYYMMDDHH

exp/cycle.sh
```diff
@@ -33,20 +33,20 @@ sed -i "s/HH/${HH}/g" input.nml
 #Jason-3
 OBS_ADT_J3=ADT-J3-${YYYY}${MM}${DD}${MM}${HH}
```
aerorahul: Is this really correct?
YYYY MM DD MM HH?

exp/cycle.sh
```diff
@@ -33,20 +33,20 @@ sed -i "s/HH/${HH}/g" input.nml
...
 sed -i "s/OBS_ADT_J3/${OBS_ADT_J3}/g" 3dvarfgat-${WINDOW_BEGIN}.yml
-cp /home/gvernier/Data/FNMOC/ioda-converters/src/marine/altimeter/adt-ioda-j3-$YYYY$MM$DD$HH.nc ./Data/${OBS_ADT_J3}.nc
+cp adt-ioda-j3-$YYYY$MM$DD$HH.nc ./Data/${OBS_ADT_J3}.nc
```
aerorahul: $YYYY$MM$DD$HH would then become $CDATE

aerorahul (dismissed):

exp/cycle.sh
```diff
@@ -1,18 +1,18 @@
...
-HH=$4    #12 
+YYYY=$1  # 2018
+MM=$2    # 04
+DD=$3    # 15
+HH=$4    # 00 
 WINDOW_LENGTH=PT$5H # DA window
 
 
 
 # Initialize B
```
travissluka, line 15: it chokes on my machine unless number of threads set:
```suggestion
export OMP_NUM_THREADS=1
# Initialize B
```

travissluka:
other than the `./cycle.sh` and `OMP_NUM_THREADS` mentioned above... it seems to be working for me

aerorahul (approved):
approved.

guillaumevernieres:
@aerorahul : Let me push my changes!

exp/cycle.py
```diff
@@ -51,17 +53,14 @@
...
+        mid_date = ymdh_mid.strftime("%Y%m%d%H")
+        #mid_date = str(ymdh_mid.year).zfill(4)+str(ymdh_mid.month).zfill(2)+str(ymdh_mid.day).zfill(2)+str(ymdh_mid.hour).zfill(2)
+        command=args.obs2ioda+' --path '+args.obs_path+' --ic '+yyyy+mm+dd+hh+' --date '+mid_date+' --window '+str(dt)
         os.system(command)
         
         # Run DA
-        command='cycle.sh '+yyyy+' '+mm+' '+dd+' '+hh+' '+str(dt)
-        print command
-        os.system(command)
+        #command='./cycle.sh '+yyyy+' '+mm+' '+dd+' '+hh+' '+str(dt)
```
travissluka: uncomment?

guillaumevernieres: oops. Just pushed.

travissluka (approved):

PR #63, opened 2019-03-21
Feature/3dvar dumprst

guillaumevernieres:


travissluka:
I'll go ahead and be the jerk who demands non-blank pull request descriptions in the future.

src/Fields/Fields.cc
```diff
@@ -217,14 +217,14 @@ namespace soca {
     int ncat = -1;
     int nf = -1;
     soca_field_sizes_f90(keyFlds_, nx, ny, nzo, nzi, ncat, nf);
-    // os << std::endl << "  Resolution = " << nx << ", " << ny
-    //   << ", Fields = " << nf;
+    os << std::endl << "  Resolution = " << nx << ", " << ny
```
aerorahul: I don't think this is allowed.
Use Log::debug instead or other ways of displaying information, unless this was part of debugging, in which case, needs clean-up.

src/Model/Model.cc
```diff
@@ -15,7 +15,7 @@
 #include "oops/util/Logger.h"
 
 #include "src/Fields/Fields.h"
-#include "src/Fortran.h"
+#include "ModelFortran.h"
```
aerorahul, line 18: Why breaking from convention here?

src/Model/soca_model_mod.F90
```diff
@@ -104,17 +127,21 @@ subroutine soca_propagate(self, flds, fldsdate)
...
+       call step_MOM(self%mom6_config%forces, &
+                     &self%mom6_config%fluxes, &
+                     &self%mom6_config%sfc_state, &
+                     &self%mom6_config%Time, &
+                     &real(self%mom6_config%dt_forcing, kind=8), &
+                     &self%mom6_config%MOM_CSp,&
+                     &start_cycle=.false.,&
+                     &cycle_length=self%mom6_config%MOM_CSp%dt)
+    else
+       !TODO: Read file
```
aerorahul: Is this valid?
An empty else block?

src/Model/soca_model.interface.F90
```diff
@@ -52,12 +52,7 @@ subroutine c_soca_setup(c_confspec, c_key_geom, c_key_model) bind (c,name='soca_
   model%advance_mom6 = config_get_int(c_confspec,"advance_mom6")
 
   ! Initialize mom6
-  if (model%advance_mom6==1) then
-     call soca_create(model)
-     model%integration_initialized = .true.
-  else
-     print *,"Not initializing MOM6"
-  end if
+  call soca_setup(model)
```
aerorahul, line 55: Ooh, I like this!

src/Model/soca_model.interface.F90
```diff
@@ -76,10 +71,7 @@ subroutine c_soca_delete(c_key_conf) bind (c,name='soca_delete_f90')
   type(soca_model), pointer :: model
 
   call soca_model_registry%get(c_key_conf, model)
-  if (model%advance_mom6==1) then
-     call soca_delete(model)
-     model%integration_initialized = .false.     
-  end if  
+  call soca_delete(model)
```
aerorahul, line 73: and this.

src/Model/soca_model.interface.F90
```diff
@@ -109,9 +101,39 @@ subroutine c_soca_prepare_integration(c_key_model, c_key_state) &
   call soca_field_registry%get(c_key_state,flds)
   call soca_model_registry%get(c_key_model, model)
 
+  call soca_prepare_integration(model, flds)
```
aerorahul: much cleaner.
Should this be soca_initialize_integration to correspond with soca_finalize_integration?

aerorahul (commented):
First glance comments.
Will run tests now.

src/Model/Model.cc
guillaumevernieres: cleaning up the Fortran.h that's in src

src/Model/soca_model_mod.F90
guillaumevernieres: I wondered too, gcc didn't complain. Will try with intel.

src/Model/soca_model.interface.F90
guillaumevernieres: Yes, left over from copy paste of qg ...

guillaumevernieres:
@travissluka : Sorry, I forgot:
Addition of an application to checkpoint the model, which cannot be done anymore when running the 3dvar. Our JEDI friends decided to not call the model in the 3dvar cost function anymore. So while a model object is still being constructed, there are no calls to init and finalize (and currently the only way to checkpoint with the analysis in the state). There are other ways to do this, like having state.save() actually dump a full mom6 restart ... But that's not how it is currently done in soca.

src/Model/Model.cc
guillaumevernieres: @aerorahul : Do you mean just sticking to Fortran.h?

src/Model/soca_model_mod.F90
guillaumevernieres: @aerorahul : Added log comment to fill the block.

src/Model/soca_model_mod.F90
```diff
@@ -104,17 +127,23 @@ subroutine soca_propagate(self, flds, fldsdate)
...
+       WRITE(buf,*) 'IO Advance of MOM6: NOT IMPLEMENTED'
```
travissluka: write and print statements that are executed on all cores need to go away, eventually . I'm already getting swamped with repetitive log lines on my 8 core machine, and dread the day I do a 720 core run on gaea.

mains/CheckpointModel.h
```diff
@@ -0,0 +1,67 @@
...
+#include "oops/runs/Application.h"
+#include "oops/util/DateTime.h"
+#include "oops/util/Duration.h"
+#include "oops/util/Logger.h"
+
+namespace soca {
+
+  class CheckpointModel : public oops::Application {
+   public:
+    static const std::string classname() {return "soca::ErrorCovariance";}
```
travissluka: soca::ErrorCovariance ?

guillaumevernieres:
I'm finding quite a few issues when building with intel17, I'm closing the PR and will resubmit later today. Hopefully.

guillaumevernieres:
@travissluka , @aerorahul :  The intel fix will be a different pr, since the problem also exist in develop.

test/testinput/checkpointmodel.yml
```diff
@@ -0,0 +1,22 @@
...
+  read_from_file: 1
+  date: 2018-04-15T00:00:00Z
+  basename: ./INPUT/
+  ocn_filename: MOM.res.nc
+  ice_filename: ice_model.res.nc
+analysis:
+  read_from_file: 1
+  date: 2018-04-15T00:00:00Z
+  basename: ./Data/
+  ocn_filename: ocn.3dvar.an.2018-04-15T00:00:00Z.nc
```
travissluka:
```suggestion
  ocn_filename: ocn.3dvarsoca.an.2018-04-15T00:00:00Z.nc
```

test/testinput/checkpointmodel.yml
```diff
@@ -0,0 +1,22 @@
...
+  ice_filename: ice.3dvar.an.2018-04-15T00:00:00Z.nc
```
travissluka:
```suggestion
  ice_filename: ice.3dvarsoca.an.2018-04-15T00:00:00Z.nc
```

travissluka (changes requested):

test/testinput/checkpointmodel.yml
travissluka: you should remove your test directory every time before a final make/ctest/push

guillaumevernieres: yes, I should ...

travissluka (approved):

PR #65, opened 2019-03-22
fix rossby radius based hz correlation length option

### LINKED ISSUE #3: Clean up implementation of gpnorm
Finish implementation of gpnorm in Field. Some fields are missing.

### LINKED ISSUE #66: Resurrect the dirac test
I missed a change in bump that I might have been able to notice had I kept the dirac test as part of our ctest.

travissluka:
bump ignores your efforts to set `cor_rh` unless `cor_rv` and `var` are passed in as well.
`ocean_corr_scale` and `ice_corr_scale` in the yml files should work as expected now

* dirac test working now closes #66
* to get dirac test working, gpnorm was cleaned up, output is sent to the correct iostream now.  closes #3 
* randomly generated fields use a fixed seed for reproducability

with the "TEST" output now being populated by the gpnorm output, other "compare.sh" tests can be enabled.


guillaumevernieres:
I did notice a speed up of the initialization of bump when running with smaller scales a few weeks ago. I must have missed a BUMP update. Time to resurrect the "dirac" test, we should then be able to visually inspect our static B applied to dirac delta functions, and make this kind of plots (just an example, figure is quite old):
dirac-test-example

travissluka:
Ah, I was wondering how to make those plots!

travissluka:
@guillaumevernieres  this PR should be good to look at now

src/Fields/Fields.cc
```diff
@@ -217,8 +218,17 @@ namespace soca {
...
     int nf = -1;
     soca_field_sizes_f90(keyFlds_, nx, ny, nzo, nzi, ncat, nf);
+    // TODO: fix this so that resolution is the global resolution
+    // os << std::endl << "  Resolution = " << nx << ", " << ny
+    //    << ", Fields = " << nf;
     std::vector<double> zstat(3*nf);
-    soca_field_gpnorm_f90(keyFlds_, nf, zstat[0]);
+    soca_field_gpnorm_f90(keyFlds_, nf, zstat[0]);    
+    for (int jj = 0; jj < nf; ++jj) {      
+      os << std::endl << std::right << std::setw(7) << vars_[jj]
```
guillaumevernieres, line 227: OK for this PR, but @aerorahul is right, we need to use log to better control the output.

src/Fields/soca_fields_mod.F90
```diff
@@ -758,51 +757,65 @@ end subroutine write_file
...
+        tmp(3) =    sum(fld%cicen(is:ie,js:je,:))
+      end select
+
+      ! calculate global min/max/mean
+      call f_comm%allreduce(tmp(1), pstat(1,jj), fckit_mpi_min())
+      call f_comm%allreduce(tmp(2), pstat(2,jj), fckit_mpi_max())
+      call f_comm%allreduce(tmp(3), pstat(3,jj), fckit_mpi_sum())
+      pstat(3,jj) = pstat(3,jj)/ocn_count
+    end do
   end subroutine gpnorm
```
guillaumevernieres, line 820: :+1:

src/Increment/Increment.cc
```diff
@@ -180,8 +180,8 @@ namespace soca {
   }
   // -----------------------------------------------------------------------------
   void Increment::print(std::ostream & os) const {
-    // os << std::endl << "  Valid time: " << validTime();
-    // os << *fields_;
+    os << std::endl << "  Valid time: " << validTime();
+    os << *fields_;
   }
   // -----------------------------------------------------------------------------
```
guillaumevernieres, line 187: OK for now, but I'll create an issue for the output, use of log instead of std. My bad!

test/CMakeLists.txt
```diff
@@ -291,20 +293,27 @@ ecbuild_add_test( TARGET test_soca_bkgerrgodas
...
-		
+                  LIBS    soca )
+
+ecbuild_add_test( TARGET test_soca_dirac
+                  TYPE SCRIPT
+                  COMMAND "./compare.sh"
+                  ARGS "${CMAKE_BINARY_DIR}/bin/soca_dirac.x testinput/dirac.yml"
+                       testoutput/dirac.test
+                  DEPENDS soca_dirac.x )
+
```
guillaumevernieres, line 309: Excelen!t :+1: I should have never dropped that test.

guillaumevernieres (approved):
All good.

src/Fields/Fields.cc
travissluka: no, @aerorahul is wrong, this is the correct way of doing it. Look at QG. you don't know what log stream the output is supposed to go to. `os` is set to whichever log stream by the calling application. This way, output goes to `debug` or `test` log streams correctly

travissluka: .. but you are right in that there are plenty of _other_ places that need to use use Log instead

guillaumevernieres:
So much for a healthy work/life balance ;)

travissluka:
.... My wife was still asleep, doesn't count!

PR #74, opened 2019-04-02
Added sea_water_salinity to getvalues

guillaumevernieres:
The generic ufo will query for sea_water_salinity (also the obs name) instead of practical salinity, thus the addition ...


travissluka (approved):

aerorahul (approved):
approved.

PR #75, opened 2019-04-02
add microwave SST observation downloads

travissluka:
added download scripts for microwave SST (AMSR2, GMI, WindSat) and goes-16


aerorahul (approved):
approved.

PR #84, opened 2019-04-10
Feature/checkpoint

aerorahul:
Apply cleaning of unsavoury increments (temperature and salinity) inside the interface.
If no limits are specified, clamping is not applied.
This will be replaced by computing balanced values from GSW in the future, and perhaps be moved elsewhere.


src/Model/soca_model_mod.F90
```diff
@@ -4,181 +4,197 @@
...
+  integer :: year, month, day, hour, minute, second
+  character(len=20) :: strdate
+  character(len=1024) :: buf
+
+  ! Update halo
+  call mpp_update_domains(flds%tocn, flds%geom%ocean%G%Domain%mpp_domain)
+  call mpp_update_domains(flds%socn, flds%geom%ocean%G%Domain%mpp_domain)
+
+  ! Checkpoint SOCA fields before jamming into MOM restarts
+  if (self%checkpoint == 1) then
```
guillaumevernieres: In my jargon, checkpointing = writing restarts. What we do here is more of a cleanup.

test/testinput/checkpointmodel.yml
```diff
@@ -7,6 +7,13 @@ model:
   tstep: "PT6H"
   advance_mom6: 0
   variables: [cicen, hicen, socn, tocn, ssh, hocn]
+  checkpoint: 1
+#  tocn_minval: -2.0
+#  tocn_maxval: 34.0
+#  socn_minval: 0.0
+#  socn_maxval: 45.0
+  tocn_minmax: [-2.0, 34.0]
```
guillaumevernieres: We only need to specify a max value, we'll get the min from the salinity.

src/Model/soca_model_mod.F90
guillaumevernieres: @aerorahul : Are there good reasons for the changes in indentation? Genuine question. What's on the left is coming from my editor's auto indentation (most likely).

aerorahul: @guillaumevernieres I was simply cleaning up and making indentation, commenting, etc. consistent across the fortran file.

travissluka:
I would suggest using different min/max values for the purposes of the ctest, because right now clamping the values is not getting tested. 

e.g. 
```
  tocn_minmax: [-1.8, 32.0]
  socn_minmax: [0.1, 38.0]
```

travissluka:
In `CheckpointModel.h` the final print to the `Log::test()` stream needs to be moved to after `model.finalize()`. Otherwise the Test output stream does not show the output with the clamped state values.

src/Model/soca_model.interface.F90
```diff
@@ -40,17 +40,28 @@ subroutine c_soca_setup(c_confspec, c_key_geom, c_key_model) bind (c,name='soca_
...
+  ! Setup time step
   ststep = config_get_string(c_confspec,len(ststep),"tstep")
   dtstep = trim(ststep)
   model%dt0 = duration_seconds(dtstep)
 
   ! Setup mom6 advance or identity model
   model%advance_mom6 = config_get_int(c_confspec,"advance_mom6")
 
+  ! Setup for checkpointing the model
+  model%checkpoint = 0
```
guillaumevernieres: I would suggest we remove model%checkpoint altogether, I don't think there would be an instance where we would not want to check for out of bound values. That would also resolve the confusion between the "checkpoint" application and the checkpoint in this pr.

src/Model/soca_model.interface.F90
```diff
@@ -40,17 +40,28 @@ subroutine c_soca_setup(c_confspec, c_key_geom, c_key_model) bind (c,name='soca_
...
+  if (config_element_exists(c_confspec,"checkpoint")) then
+    model%checkpoint = config_get_int(c_confspec,"checkpoint")
+  endif
+
+  if (model%checkpoint == 1) then
```
guillaumevernieres: As above, but have a default minmax for t and s.

travissluka: instead of setting a default minmax, the code should check to see if `tocn_minmax` and/or `socn_minmax` are present in the yaml file, and if not then no minmax is applied

aerorahul:
good point. Done.

src/Model/soca_model.interface.F90
aerorahul: done.

src/Model/soca_model.interface.F90
aerorahul: done.

travissluka:
Why are we doing work at 10:40 pm?

guillaumevernieres:
Better than reshaping your fingers with your favorite chisel.

guillaumevernieres (approved):

travissluka:
sorry, I'm vetoing this. @guillaumevernieres  did you run the ctests??? `test_soca_checkpointmodel` fails because the .test file was not updated

aerorahul:
@travissluka The test fails for me anyways, because of precision. GCC 7.3 or Intel.
But, I agree, the test file needs to be updated. If I updated it with my run, your tests will definitely fail.  Do you want me to do that?

travissluka:
@aerorahul i can add the updated test file, so that at least the tests will pass in the container on "normal" PCs. I'm still trying to get things working on Gaea, I'm hoping that once I do the answers will be different 
and I can spend some time getting tolerances added to the tests.

Also, ignore the "revert branch" that was created, I got click-happy and haven't had my coffee yet this morning.

guillaumevernieres:
Sure did, but I probably didn't update my sandbox.

PR #91, opened 2019-04-22
Feature/efficient ctest

### LINKED ISSUE #85: Simplified model config for ctests
ctests are too slow,

I'm currently looking at creating 2 or 4 degree horizontal config files with reduced vertical resolution

travissluka:
ctests have been cleaned up a bit to 1) run faster and 2) check the norms for nearly every application. 
closes #85
- a 4 degree, 25 vertical level model configuration is added. Grid is only 1% the size of the previously used 1deg configuration.
- all tests are set to use 2 procs, for reproducibility
- smaller observation files are used (~100 observations per obs type). Tests use all obs types by default now.
- the above changes result in a MUCH faster ctest runtime
- `Log::Test()` output stream added to the checkpointing app.
- `compare.sh` is used to check the  `Log:Test()` stream for every application. Ctests should break easier now when something changes (and I mean that in a good way)
- dirac location specification cleaned up, yml file doesn't care about which process a gridpoint is on (dirac locations on any PE other than 0 was actually wrong)

```
The following tests FAILED:
         10 - test_soca_model (Failed)
         13 - test_soca_linearmodel (Failed)
         20 - test_soca_parameters_bump_cov_lct (Timeout)
         21 - test_soca_parameters_bump_cov_nicas (Timeout)
         22 - test_soca_dirac_bump_cov (Failed)
         24 - test_soca_hofx (Failed)
         28 - test_soca_3dvarfgat (Timeout)
         29 - test_soca_4dhybrid (Failed)
         30 - exp_soca_cycle (Failed)
```

test failures given above should be addressed outside this PR. Test failures are the same as before this PR, with the exception of:
- `3dvarfgat` works, but only when ADT  and SSS obs are commented out. `OOPS_TRACE` shows that it is hanging on `State<MODEL>::getValues`. A timeout of 6 seconds is set to make it not hang, for now
- the same is true for `hofx`, however it fails anyway if ADT or SSS is removed. This failed anyway before this PR, something changed in ufo/ioda
- `test_soca_parameters_bump_cov_lct` actually works now, but `test_soca_parameters_bump_cov_nicas` and `test_soca_dirac_bump_cov` do not, so there is no point letting it run for now.

any test file in `testinput/` that is not currently being used has been moved to `testinput/unused`. @guillaumevernieres  please remove any test files that are actually no longer needed.


aerorahul:
@travissluka can you create a branch in `soca-bundle` containing the corresponding `ufo`, `oops`,`ioda` branches that you used to test this branch.  I suspect, they are not all `develop`

travissluka:
@aerorahul I was using the bundle as it was before you had made changes to MOM6/FMS. I pushed what I was using to feature/efficient_ctest. For some reason, with the MOM6 in the latest develop bundle, the ctests are going painfully slow, even if I make sure build type is set to Release. Something's not right. The ctests in soca should take about 30 seconds to finish.

aerorahul:
@travissluka I’ll take a look. I have a couple more changes in fms::dev/master-ecbuild-IntelFlags, mom6::bugfix/openmp and soca::bugfix/intel. 
The soca-bundle::develop should build without openmp and in release mode with the above branches.

test/CMakeLists.txt
```diff
@@ -127,208 +106,220 @@ ecbuild_add_resources( TARGET   soca_test_scripts
...
+                  COMMAND compare.sh
+                  ARGS
+                      "${CMAKE_BINARY_DIR}/bin/soca_ensvariance.x
+                      testinput/ensvariance.yml"
+                      testoutput/ensvariance.test
+                  DEPENDS soca_ensvariance.x)
+
 
+# The following two tests are slow, how can we speed it up?
+# Use another model grid configuration with even fewer grid cells?
```
guillaumevernieres, line 282: Increasing the length scale should speed up the test.

travissluka: i'm not familiar with the settings for bump running in this mode. My usual "randomly change everything until I get the desired result" method shows that changing nc1, nc2,nc3 to smaller numbers makes it go faster... but I really have no idea what i'm doing :)

aerorahul:
@travissluka  I have a longer list of failed tests with Intel17
```
	  5 - test_soca_socaerror_init (Failed)
	  6 - test_soca_enspert (Failed)
	 10 - test_soca_model (Failed)
	 12 - test_soca_errorcovariance (Failed)
	 13 - test_soca_linearmodel (Failed)
	 17 - test_soca_bkgerrgodas (Failed)
	 19 - test_soca_ensvariance (Failed)
	 20 - test_soca_parameters_bump_cov_lct (Failed)
	 21 - test_soca_parameters_bump_cov_nicas (Failed)
	 22 - test_soca_dirac_bump_cov (Failed)
	 23 - test_soca_dirac_soca_cov (Failed)
	 24 - test_soca_hofx (Timeout)
	 25 - test_soca_3dvarsoca (Failed)
	 26 - test_soca_3dvargodas (Failed)
	 27 - test_soca_checkpointmodel (Failed)
	 28 - test_soca_3dvarfgat (Failed)
	 29 - test_soca_4dhybrid (Failed)
	 30 - exp_soca_cycle (Failed)
```

But the model test runs fast as expected:
```
      Start  4: test_soca_forecast_mom6
 4/30 Test  #4: test_soca_forecast_mom6 ...............   Passed    2.23 sec
```

The `socaerror_init` test failure is a result of `ncwrite` and probably `numprocs` related.
```
5: --- Block: 01_01_01_01
5:        Compute adaptive sampling
5: !!! Error in io_fld_write: dimension nc0 has a different size in file
5: !!! ABORT in io_fld_write on task #0001: dimension nc0 has a different size in file
5: application called MPI_Abort(MPI_COMM_WORLD, 1) - process 0
1/1 Test #5: test_soca_socaerror_init .........***Failed    3.39 sec
```

travissluka:
@aerorahul did you try removing the soca/test directory in the build dir? I think i've seen that error before if the bump files already exist, but had been created with different PE count.

aerorahul:
@travissluka you are right. I was reusing the build with a previous test.

travissluka:
@aerorahul BTW, for the newer MOM/FMS compiling with GCC, the `MOM6::bugfix/openmp` branch does the trick for me. Ctests run at full speed when using that branch with `FMS::dev/master-ecbuild`

aerorahul:
@travissluka please try `FMS::dev/master-ecbuild-IntelFlags`

travissluka:
@aerorahul it works fine with `FMS::dev/master-ecbuild-IntelFlags`

guillaumevernieres (approved):
Outstanding @travissluka ! Seems fine on gcc, will test on intel later but I'd be happy to merge that work and move on to the other pr's, then address the intel issues in a different pr.

travissluka:
@guillaumevernieres I would say merge.
We can work on the known bugs, and adding tolerances to the Test outputs, afterward.

PR #92, opened 2019-04-23
bugfix for intel

aerorahul:
Intel compiler demands to be a pointer instead of a target.

Change-Id: I6570360d757b2ae03c69f74af761bd5c21873e69


guillaumevernieres:
@aerorahul : This was addressed in an other PR ... I think we talked about it a while ago.

aerorahul:
@guillaumevernieres you are correct. There is a branch feature/intel-fix, but no PR.
I'll issue a PR from there and close this, if that is preferred.

guillaumevernieres:
@aerorahul : Never mind, I don't see the fix in any of the other pr ...

guillaumevernieres:
@aerorahul : nope, that pr is fine! keep it.

travissluka (approved):

PR #95, opened 2019-04-24
Resurrection of 3dvarfgat and hofx tests (although that later one is just a zombie)

guillaumevernieres:
All I did was to move the burden of handling the zero obs case downstream to bump interp. 
3dvarfgat passes, hofx doesn't hang but fails for issues pointed out by @danholdaway .


travissluka:
great, one bug down. 

BTW, I am getting failures as well in anything that uses BkgErrGODAS. No idea why, it worked yesterday and I don't see anything that changed.... separate problem

travissluka (approved):

PR #97, opened 2019-04-24
domain pointer in BkgErrGODAS was uninitialized

### LINKED ISSUE #96: BkgErrGODAS broken
PR #92 to make intel happy has angered the GCC gods and now all tests using BkgErrGODAS fail.
I'm looking into a fix for both

travissluka:
Made into a non-pointer type. closes #96

illegal use of target/pointer might also be why bkgerrgodas was acting weird on Theia

### DIFF
--- a/src/Transforms/BkgErrGodas/soca_bkgerrgodas_mod.F90
+++ b/src/Transforms/BkgErrGodas/soca_bkgerrgodas_mod.F90
@@ -118,7 +118,7 @@ contains
     type(soca_bkgerrgodas_config),     intent(inout) :: self
 
     real(kind=kind_real), allocatable :: sig1(:), sig2(:)
-    type(soca_domain_indices), pointer :: domain
+    type(soca_domain_indices) :: domain
     integer :: is, ie, js, je, i, j, k
     integer :: ins, ns = 1, iter, niter = 1
     type(soca_omb_stats) :: sst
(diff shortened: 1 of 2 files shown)


aerorahul:
@travissluka all tests using BkgErrGODAS now run to completion on both Intel and GCC.
The tests still fail to compare. e.g.
test_soca_enspert tries to compare testinput/genenspert.test with the output of the ctest.
The ctest, does not write out the Initial State to output.
If the reference testinput/genenspert.test is edited to remove Initial state, the test passes.
I am looking into why the ctest is not writing out the Initial state.  It spits out the Initial state to stdout, but not to the file.

The missing part is this:
```
6: Test     : Initial state:
6: Test     :   Valid time: 2018-04-15T00:00:00Z
6: Test     :   cicen   min=    0.000000   max=    1.000000   mean=    1.515270
6: Test     :   hicen   min=    0.000000   max= 2667.837408   mean=  493.034857
6: Test     :    socn   min=    0.000000   max=   39.037978   mean=  862.697603
6: Test     :    tocn   min=   -1.883330   max=   31.008851   mean=  153.848302
6: Test     :     ssh   min=   -2.103854   max=    0.912143   mean=   -0.297046
6: Test     :    hocn   min=    0.001000   max= 1345.640000   mean= 3267.054300
```

aerorahul (approved):
all tests run to completion with this fix, on both Intel and GCC.

aerorahul:
On GCC, I am still getting this annoying message at the end:
```
6: Primary job  terminated normally, but 1 process returned
6: a non-zero exit code. Per user-direction, the job has been aborted.
6: -------------------------------------------------------
6: --------------------------------------------------------------------------
6: mpirun detected that one or more processes exited with non-zero status, thus causing
6: the job to be terminated. The first process to do so was:
6:
6:   Process name: [[30201,1],0]
6:   Exit code:    1
```

I nuked the soca directory, but still.

PR #98, opened 2019-04-24
fix MPI_Finalize() being called in soca

### LINKED ISSUE #93: MPI_Finalize issue with model
`test_soca_model` and `test_soca_linearmodel` both "pass" except that the programs terminate with the error

```
10: *** The MPI_Comm_f2c() function was called after MPI_FINALIZE was invoked.
10: *** This is disallowed by the MPI standard.
10: *** Your MPI job will now abort.
```

travissluka:
closes #93
this is a temporary fix, to make `test_soca_model` and `test_soca_linearmodel` work again, and should be examined more closely in the future to be done cleaner.

### DIFF
--- a/src/Model/soca_mom6.F90
+++ b/src/Model/soca_mom6.F90
@@ -310,7 +310,10 @@ contains
     type(soca_mom6_config), intent(inout) :: mom6_config
 
     ! Finalize fms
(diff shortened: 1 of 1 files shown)


guillaumevernieres (approved):
Does the trick.

PR #100, opened 2019-04-24
ice concentration obs added to relevant ctests

travissluka:
a reduced set of 100 ice concentration observations are added to the var and hofx ctests. Ignore the science for now, these are added to pick up on any unintended ctest answer changes in the future.

fun observations:
- ice concentration in the analysis below 0.0 and above 1.0... sure why not
- 3dvarfgat is not making use of ice concentration obs. Perhaps expected given no advancing ice model? I would have thought it would at least use a persistent background ice state throughout the window.


guillaumevernieres:
@travissluka : having the ice analysis out of bounds is a sign of a great analysis. The same goes for our negative salinity analysis.  
Nothing to do with the sea-ice (no)advance. The time stamp for the emc ice fraction is 0 ... so what you see makes sense. Unless you are not using the emc-ice and the obs time stamp is not 0.

travissluka:
@guillaumevernieres all the obs are 0 hours from the window center, so there should be 1 time step where they are used in fgat. plain 3dvar makes use of them.

travissluka:
ah, nevermind, the fgat doesn't run long enough to get to 12Z. I'll randomize the hours of the ice obs so that fgat has something to work with

guillaumevernieres (approved):
All good, or as good as it can currently get.

PR #102, opened 2019-04-26
update .test file for latest change to checkpoint

travissluka:


guillaumevernieres (approved):
I must have forgotten to fetch the latest changes of @aerorahul 's branch, that test was passing for me. Thanks for the fix.

aerorahul (approved):
approved.

PR #107, opened 2019-04-29
updated answer files for tests

travissluka:
the test answer files were regenerated after 
1) a ufo/ioda/oops fix was made to hofx
2) fix a bug where gpnorm was not printing out the correct mean for 3d state variables

assuming you're not @aerorahul , we should be down to just the following ctest failures

```
Total Test time (real) =  31.26 sec

The following tests FAILED:
         20 - test_soca_parameters_bump_cov_lct (Timeout)
         21 - test_soca_parameters_bump_cov_nicas (Timeout)
         22 - test_soca_dirac_bump_cov (Failed)
         31 - test_soca_4dhybrid (Failed)
```


aerorahul (approved):
looks good.

PR #111, opened 2019-05-01
dot slash in front of exec.

guillaumevernieres:
Quick fix for users without the current  dir in PATH. This PR only addresses some of the failure on Theia/intel.


aerorahul (approved):
should be harmless.

PR #115, opened 2019-05-03
fix some uninitialized memory access found with valgrind

travissluka:
I found some instances of  usage of uninitialized values in soca. There was a small, but non-zero, chance that this could have been causing machine/compiler/random answer changes. More importantly though this is clearing up valgrind complaints so that future valgrind debugging will be easier.

other than errors in ioda (fixed with https://github.com/JCSDA/ioda/pull/215) and oops (https://github.com/JCSDA/oops/issues/238 , though I don't _think_ this one impacts answers in soca, errors seem localized to print statements for us), the only remaining valgrind complaints are within MOM6 advance... and I don't think any of us wants to fix MOM6 valgrind complaints.

### DIFF
--- a/src/Fields/soca_fields_mod.F90
+++ b/src/Fields/soca_fields_mod.F90
@@ -116,6 +116,7 @@ contains
 
     ! Allocate and copy fields
     call soca_field_alloc(self, rhs_fld%geom)
+    call zeros(self)
     !call copy(self,rhs_fld)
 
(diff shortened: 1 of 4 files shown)


guillaumevernieres (approved):
Thanks for the fix. Tests all ok, but I realized I forgot to remove the "ones" method or implement it in ocnsfc in the last pr ...

PR #124, opened 2019-05-13
Added bounds for T&S in init integration

### LINKED ISSUE #122: Analysis state "clean-up"
Simple fix while waiting for the implementation of change of variables for the ocean tracers.
T and S also need to be clamped in the model init phase, this is currently only done in the finalize phase.

guillaumevernieres:
This is a follow up of @aerorahul 's "checkpointing" PR, which checked for out of bound values when writing a model restart. This PR does the same thing but at the level of the initialization of the model integration, and will, for example check for out of bound values at the start of each outer loop or after adding perturbations to the state.

closes #122


travissluka (approved):
all tests pass once `develop` is merged in

aerorahul (approved):
changes look good.

PR #130, opened 2019-05-14
feature/marineobs2ioda

aerorahul:
convert a batch of marine obs to ioda.
controlled via a yaml file.

will need more improvements, suggestions welcome.


guillaumevernieres:
I'm OK with no test for now, but how about moving the script to /bin and call it soca_...?

aerorahul:
@guillaumevernieres I agree with the name change.
As for tests, I intend to add tests in following PR with a cycling ability.

I will also add hgodas use case as well.

aerorahul:
@guillaumevernieres script has been renamed.
@travissluka hgodas_*2ioda.py probably already reads the windowed data; i.e. it does not handle multiple input files for each day or hour for e.g.  I assume you are creating the files a-priori and this script simply converts format to ioda.

guillaumevernieres (approved):
All good.

PR #136, opened 2019-06-10
IO for CICE

### LINKED ISSUE #104: sea-ice IO for CICE (5, 6 and 4?)
We might eventually develop a separate CICE interface, but in the meantime, we need to develop the necessary tools to:

1. Read in ice thickness/fraction from a background cice file. 

2. Save the sea ice analysis in a background restart file for CICE. This work includes checking for out of bound values and adjusting other ice parameters (energy/enthalpy/tracer volume, ...) necessary to restart the model. 

@aerorahul : I'll start with CICE5, but do you need the same for CICE4? Should we care about SIS2?

guillaumevernieres:
This PR addresses issue #104 
What was done:
- Modularized seaice field (moved seaice fields out of field).
- Convert sis2 "h_ice" from [kg/m2] to [m] when reading sis2 file
- Added IO for CICE, should work-ish for 4, 5 and 6, but some scripting will be required to rename  dimensions since I am using fms to read/write.
- Added aggregate variables for ice fraction and thickness
- Updated out test files to reflect changes in ice thickness units
closes #104


guillaumevernieres:
@aerorahul , @travissluka : I'll be merging by the end of the day, with or without review! But I'd rather you have a look at it. I need that work in develop for the cycling of cice at EMC. There are also more sea-ice related pr waiting on the pipe line ...

src/Fields/soca_seaice_mod.F90
```diff
@@ -0,0 +1,379 @@
...
+     procedure :: sub => soca_seaice_sub
+     procedure :: mul => soca_seaice_mul
+     procedure :: axpy => soca_seaice_axpy
+     procedure :: diff_incr => soca_seaice_diff_incr
+     procedure :: read_restart => soca_seaice_read_rst
+     procedure :: read_diag => soca_seaice_read_diag
+     procedure :: write_restart => soca_seaice_write_rst     
+  end type soca_seaice_type
+
+  real(kind=kind_real), parameter :: soca_rho_ice = 905.0 ! [kg/m3]
```
aerorahul, line 49: should we have a constants file dedicated for constants.
Scattered constants can be hidden and dangerous!
Just a thought.

src/Fields/soca_seaice_mod.F90
```diff
@@ -0,0 +1,379 @@
+!
+! (C) Copyright 2017-2019 UCAR
```
aerorahul: Should be 2017-
instead of 2017-2019

src/Fields/soca_seaice_mod.F90
```diff
@@ -0,0 +1,379 @@
...
+    deallocate(self%cicen)
+    deallocate(self%hicen)
+
+  end subroutine soca_seaice_delete
+
+  ! ------------------------------------------------------------------------------  
+  subroutine soca_seaice_zeros(self)
+    class(soca_seaice_type), intent(inout) :: self
+
+    self%cicen      = 0.0_kind_real
```
aerorahul: why the unnecessary space?
Here and all through the file.

aerorahul (approved):
looks good.
minor comments.

src/Fields/soca_seaice_mod.F90
guillaumevernieres: copy paste from fields, previously aligned with something longer I guess, will fix.

src/Fields/soca_seaice_mod.F90
guillaumevernieres: I'm following what is done in the other repos. Are you sure this is wrong? Every file I touch, I update the year if need be.

aerorahul: I think in general is year begin  -  year end.
I have seen the other elsewhere in code too.
may be leave it.

src/Fields/soca_seaice_mod.F90
guillaumevernieres: You are right, but I should probably pass rho_ice through the config ...

aerorahul: ok. for now.
can cleanup later.

PR #137, opened 2019-06-19
NSIDC ice concentration obs

guillaumevernieres:
Added script to download the nsidc sea-ice fraction retrieval.


aerorahul (approved):
the download script works.
Note, the NCEI data is available till 2015, so 2018-04-15 (default date) will fetch nothing

guillaumevernieres:
We should probably add error traps for out of bound requests. I'll create an issue ...

PR #141, opened 2019-06-21
Forecast with 1deg setup crashes after 23z 

### LINKED ISSUE #140: Model crashing due to issue in diag manager
It use to work as is, but not anymore. The 1deg setup, used in "exp", crashes when calling the diag manager in mom6.

guillaumevernieres:
closes #140 
This issue only affects the 1deg setup, which isn't tested anymore, so this is not affecting the tests.


guillaumevernieres:
@aerorahul : First step towards fixing our issue with the forcing not being updated. I'll have the second part ready by the end of the day, I think.

aerorahul (approved):
works.

PR #142, opened 2019-06-21
feature/geomcleanup

aerorahul:
Clean-up of the SOCA geometry interface.  Specifically;
- `soca_geom_mod.f90` is renamed to `soca_geom.interface.F90` to be consistent across SOCA and other projects. `soca_model_geom_type.f90` is renamed to `soca_geom_mod.f90` to be consistent across SOCA objects e.g. fields etc.
- Listed type `soca_geom` is now accessed from `soca_geom_mod` directly.
- Above 2 changes, required changes in affected routines.

All tests pass (as expected).


guillaumevernieres (approved):
All good. Thanks for the clean up.

PR #144, opened 2019-06-21
Check if the forcing needs to be updated at every time step.

### LINKED ISSUE #143: Surface forcing not updated
Reading of surface forcing is not implemented properly.

guillaumevernieres:
closes #143
Reading of the forcing was only initialized at the model init step. This PR implements the reading and update of the forcing at every time step, using mom6's "set_forcing".


src/Model/soca_model_mod.F90
```diff
@@ -139,6 +142,17 @@ subroutine soca_propagate(self, flds, fldsdate)
   ocean_time = self%mom6_config%Time
 
   if (self%advance_mom6==1) then
+     ! Set the forcing for the next steps.
+     call fms_io_init()
```
aerorahul, line 146: We need to figure out a way to do `fms_io_init` and `fms_io_exit` once.  Init at the beginning as part of setup, and exit at the end as part of the cleanup.

aerorahul: Not suggesting it has to be now, or in this PR.  Just an FYI

guillaumevernieres: I knew you were going to comment on this! Certainly annoying to have to call these routines, I'm guessing they are called somewhere in mom and override whatever we would do in run int/end.

src/Model/soca_mom6.F90
```diff
@@ -212,7 +213,7 @@ subroutine soca_mom6_init(mom6_config)
 
     type(tracer_flow_control_CS), pointer :: &
          tracer_flow_CSp => NULL()  !< A pointer to the tracer flow control structure
-    type(surface_forcing_CS),  pointer :: surface_forcing_CSp => NULL()
+    !type(surface_forcing_CS),  pointer :: surface_forcing_CSp => NULL()
```
aerorahul: can be removed!

src/Model/soca_mom6.F90
```diff
@@ -104,13 +104,14 @@ module soca_mom6
                                      !< at the ocean surface.
     type(surface)      :: sfc_state  !< Pointers to the ocean surface state fields.
     real               :: dt_forcing !< Coupling time step in seconds.
+    type(time_type)    :: Time_step_ocean !< time_type version of dt_forcing
```
aerorahul, line 107: L200 can be removed as well.

aerorahul (approved):
minor changes requested, but approved!

PR #146, opened 2019-06-25
Removed multiple public declaration of soca_geom

guillaumevernieres:
Multiple public declaration for the same type fails to compile with intel.


aerorahul (approved):
thanks for the catch @guillaumevernieres

PR #147, opened 2019-06-25
CICE initialization

### LINKED ISSUE #148: Create necessary balanced restart fields for CICE
Simple adjustment of ice and snow volume to make the integration of cice stable.

guillaumevernieres:
Added the possibility of creating the necessary fields (fraction, ice and snow thickness) to restart CICE (5&6, not 4). This was going to live inside a script, but I changed my mind and the PR is now a bit more involved than expected. Simple methods, the fancy stuff will go in the cice dedicated interface. 
New features:
- Added back snow depth
- The add increment method for sea-ice will require some open-mindedness from the reviewer ... The method discards the per category analysis and replaces it by a re-scaling of the background. Typical way to create "balanced" initial condition for CICE.
- Adjusted output comparison files to reflect the changes.
closes #148


src/Fields/Fields.cc
```diff
@@ -239,7 +239,7 @@ namespace soca {
     int ncat = -1;
     int nf = -1;
     soca_field_sizes_f90(keyFlds_, nx, ny, nzo, nzi, ncat, nf);
-    bool ok = (nf == 11);    // <---- HARD CODED STUFF ... NEED TO CHANGE
+    bool ok = (nf >= 3);    // <---- HARD CODED STUFF ... NEED TO CHANGE
```
travissluka, line 242: eh, yeah, I don't like this either. we need a better way at some point of marking which fields are optional and which are required

test/testoutput/genenspert.test
```diff
@@ -65,8 +65,8 @@ Test     :      lw   min=   -0.000000   max=   89.266090   mean=   47.830237
 Test     :      us   min=    0.004396   max=    0.021537   mean=    0.011925
 Test     : Member 4 final state: 
 Test     :   Valid time: 2018-04-15T06:00:00Z
-Test     :   cicen   min=   -0.654142   max=    1.519502   mean=    0.254830
-Test     :   hicen   min=  -71.710943   max=   81.252156   mean=    0.611073
+Test     :   cicen   min=    0.000000   max=    1.000000   mean=    0.252399
+Test     :   hicen   min=   -1.697620   max=    3.085750   mean=    0.121511
```
travissluka, line 69: I'm sad to see the negative sea ice fractions disappear, but negative thickness is still here, yay!
Should thickness values be clamped positive the same way concentration is?

guillaumevernieres: Yes, but next pr, hicen and hsnon are not passed back to the model yet. I could zero out the perturbations for hicen, but having it out of bound in the test is a good reminder of the issue.  I did reduce the scales, -1.69m vs -71.7m :)

travissluka (approved):

PR #149, opened 2019-07-01
Optional saving of local geometry

### LINKED ISSUE #139: Make the writing of the geometry optional.
Currently writes the global and local geometry.

guillaumevernieres:
closes #139 
The geometry constructor was saving the local domain for each pe, useful for debugging but cumbersome the rest of the time. It is now added as an option, the default does not save the local geometry.

### DIFF
--- a/src/Geometry/soca_geom_mod.F90
+++ b/src/Geometry/soca_geom_mod.F90
@@ -5,6 +5,7 @@
 !
 
 module soca_geom_mod
+  use config_mod  
   use MOM_grid,                  only : ocean_grid_type
   use MOM_verticalGrid,          only : verticalGrid_type
   use soca_mom6
@@ -44,6 +45,7 @@ module soca_geom_mod
      integer,              allocatable :: ij(:,:)        ! index of ocean+shore line in compute grid
      real(kind=kind_real), allocatable :: cell_area(:,:)
      real(kind=kind_real), allocatable :: rossby_radius(:,:)
+     logical :: save_local_domain ! If true, save the local geometry for each pe.
    contains
      procedure :: init => geom_init
      procedure :: end => geom_end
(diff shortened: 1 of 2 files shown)


src/Geometry/soca_geom_mod.F90
```diff
@@ -68,7 +70,21 @@ subroutine geom_init(self, c_conf)
...
+
     call soca_geom_init(self%G, self%GV, self%ice_column, c_conf)
+
+    ! Set output option for local geometry
+    self%save_local_domain = .false.
+    if (config_element_exists(c_conf,"save_local_domain")) then
+       isave = config_get_int(c_conf,"read_from_file")
+       if (isave.eq.1 ) then
+          self%save_local_domain = .true.
+       else
```
aerorahul: Default is set to `.false.`  An else block is not needed.

src/Geometry/soca_geom_mod.F90
```diff
@@ -294,22 +310,24 @@ subroutine geom_infotofile(self)
...
+    if (self%save_local_domain) then
+       ! Save local compute grid
+       pe = f_comm%rank()
+
+       write (strpe,fmt) pe
+       geom_output_pe='geom_output_'//trim(strpe)//'.nc'
+
+       call geom_get_domain_indices(self, "compute", is, ie, js, je)
+       ns = (ie - is + 1) * (je - js + 1 )
+       varname='shoremask'
```
aerorahul: Can we not just remove varname? The write2pe has self%varname in it anyways.

e.g.
`call write2pe(reshape(self%shoremask,(/ns/)),'shoremask',geom_output_pe,.false.)`

src/Geometry/soca_geom_mod.F90
```diff
@@ -272,8 +288,8 @@ end subroutine geom_validindex
   ! ------------------------------------------------------------------------------
   !> Write geometry info to file
   subroutine geom_infotofile(self)
-    class(soca_geom), intent(in) :: self
-
+    class(soca_geom),  intent(in) :: self
```
aerorahul: delete white space, here and L292

aerorahul (approved):
few comments. but looks good.

src/Geometry/soca_geom_mod.F90
guillaumevernieres: We could, but we'd get a warning at compile time (string length not matching).

src/Geometry/soca_geom_mod.F90
guillaumevernieres: oops ...

src/Geometry/soca_geom_mod.F90
aerorahul: even if you turn L123 in soca/src/Utils/soca_utils.F90 to:
```
character(len=*), intent(in) :: varname
```

guillaumevernieres: nope, that works, can't seem to see past the tip of my nose this am ...

travissluka (approved):

PR #154, opened 2019-07-11
cleanup of geometry data structure.

aerorahul:
Removes calls to getting compute and data domain indices.  These are now part of geometry data structure.
next PR will focus on updating the interface to pass geometry instead of keeping it as a pointer in various other data structures.

This PR is zero-diff.

### DIFF
--- a/src/Fields/soca_seaice_mod.F90
+++ b/src/Fields/soca_seaice_mod.F90
@@ -8,17 +8,17 @@
 module soca_seaice_mod
 
   use config_mod
-  use datetime_mod  
+  use datetime_mod
   use fms_io_mod, only : fms_io_init, fms_io_exit,&
        &register_restart_field, restart_file_type,&
        &restore_state, free_restart_type, save_restart
-  use fms_mod,    only: read_data  
-  use iso_c_binding  
+  use fms_mod,    only: read_data
+  use iso_c_binding
   use kinds
-  use MOM_forcing_type,    only : forcing  
+  use MOM_forcing_type,    only : forcing
   use random_mod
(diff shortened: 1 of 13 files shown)


src/Fields/soca_seaice_mod.F90
```diff
@@ -52,36 +52,35 @@ module soca_seaice_mod
...
-    self%jed = jed
+    isd = geom%isd ; self%isd = isd
+    ied = geom%ied ; self%ied = ied
+    jsd = geom%jsd ; self%jsd = jsd
+    jed = geom%jed ; self%jed = jed
 
     ! Get number of categories
-    self%ncat = geom%ncat
-    
+    ncat = geom%ncat ; self%ncat = ncat
```
guillaumevernieres, line 74: doesn't look like ncat is used

aerorahul: @guillaumevernieres It is actually used in L180, L235 in [src/Fields/soca_seaice_mod.F90](https://github.com/JCSDA/soca/pull/154/files/2f62b87302f35f180bce78ebd9eb7e831f840ffc#diff-537450855ac66c90b20a24b65b6f736c)

travissluka (approved):

src/Fields/soca_seaice_mod.F90
guillaumevernieres: That's a different routine. Why not self%ncat = geom%ncat? that seems to be the only use of ncat here ...

guillaumevernieres (approved):

PR #162, opened 2019-07-24
FTP service is no longer available at PO.DAAC, the scripts using it fail to download the obs

flampouris:
Since the 3rd June 2019, FTP service is no longer available at PO.DAAC

All the scripts that used it (source.ghrsst_sst.sh, source.jpl_smap.sh, source.nesdis_sst_viirs.sh) have been updated to use the PO.DAAC Drive instead.

The PO.DAAC is password-protected, the user has to register at the PO.DAAC Drive server and create at the working machine ~/.netrc file as following:

```
touch ~/.netrc
chmod 600 ~/.netrc
echo 'machine podaac-tools.jpl.nasa.gov login USERNAME_from_PO.DAAC password PASSWORD_from_PO.DAAC' >> ~/.netrc
```

To download data, use the get_obs.sh as usual.

### DIFF
--- a/scripts/obs/README.md
+++ b/scripts/obs/README.md
@@ -1,36 +1,34 @@
 # Download observations from public servers
-The scripts in the current directory (soca/scripts/obs) can be used to download observations from public servers.
+The scripts in the current directory (soca/scripts/obs) can be used to download oceanographic observations from public servers.
 
 ## Getting Started
-The scripts are written in bash. They could be executed to any machine with linux OS.
+The scripts are written in bash. They could be executed to any machine with *nix OS which supports Bash shell. 
 
 ### Prerequisites
-For the data hosted at the Physical Oceanography Distributed Active Archive Center (PO.DAAC), the access to the observations is two step process:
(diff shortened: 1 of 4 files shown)


guillaumevernieres:
> create at the working machine ~/.netrc file as following:
> 
> ```
> touch ~/.netrc
> chmod 600 ~/.netrc
> echo 'machine podaac-tools.jpl.nasa.gov login USERNAME_from_PO.DAAC password PASSWORD_from_PO.DAAC' >> ~/.netrc
> ```
> 
> To download data, use the get_obs.sh as usual.

@flampouris : Could you add that info somewhere? Maybe in a readme?

flampouris:
@guillaumevernieres The README.md was added with additional information.

And I tested also on my laptop and the available observations were downloaded.

travissluka:
@flampouris awesome, thanks for doing this, I didn't know about the .netrc trick!
 The password should be the "Drive API Credentials" password, not the general PODAAC website password, correct? If so could you clarify that in the readme?

travissluka (approved):

PR #164, opened 2019-07-31
added option to not call mom6 setup

### LINKED ISSUE #163: Add the possibility to by-pass the model setup
This would be useful to reduce the memory over-heads when running 3dvar for example.

guillaumevernieres:
closes #163 
This feature adds the option to by-pass mom6 setup and reduce the associated memory overheads. This can be useful when running 3DVAR, for example, where the model is not used, but still constructed ([L68 of Variational.h](https://github.com/JCSDA/oops/blob/a9be4065788a8f016e0d23a0b851fd8948f9a2bf/src/oops/runs/Variational.h#L68)).

### DIFF
--- a/src/soca/Model/Model.cc
+++ b/src/soca/Model/Model.cc
@@ -28,18 +28,25 @@ namespace soca {
   static oops::ModelMaker<Traits, Model> makermodel_("SOCA");
   // -----------------------------------------------------------------------------
   Model::Model(const Geometry & resol, const eckit::Configuration & model)
-    : keyConfig_(0), tstep_(0), geom_(resol), vars_(model)
+    : keyConfig_(0), tstep_(0), geom_(resol), vars_(model), setup_mom6_(true)
   {
     Log::trace() << "Model::Model" << std::endl;
     Log::trace() << "Model vars: " << vars_ << std::endl;
     tstep_ = util::Duration(model.getString("tstep"));
-    const eckit::Configuration * configc = &model;
-    soca_setup_f90(&configc, geom_.toFortran(), keyConfig_);
-    Log::trace() << "Model created" << std::endl;
+    setup_mom6_ = model.getBool("setup_mom6", true);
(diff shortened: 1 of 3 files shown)


src/soca/Model/Model.cc
```diff
@@ -28,18 +28,25 @@ namespace soca {
...
     Log::trace() << "Model::Model" << std::endl;
     Log::trace() << "Model vars: " << vars_ << std::endl;
     tstep_ = util::Duration(model.getString("tstep"));
-    const eckit::Configuration * configc = &model;
-    soca_setup_f90(&configc, geom_.toFortran(), keyConfig_);
-    Log::trace() << "Model created" << std::endl;
+    setup_mom6_ = model.getBool("setup_mom6", true);
+    if (setup_mom6_)
+      {
+        const eckit::Configuration * configc = &model;
```
aerorahul: This line is best outside of the if block. It does not cost anything.

test/testinput/3dvar_godas.yml
```diff
@@ -18,8 +18,10 @@ resolution:
 
 model:
   name: SOCA
+  setup_mom6: false
   tstep: PT1H
   advance_mom6: 0
+  setup_mom6: false
```
aerorahul, line 23: Why repeated?

aerorahul (approved):
I suppose we always need to create the model object, and cannot get away without adding an if-block inside the constructor/destructor.

Code changes look good.

ytremolet:
I'm working on removing the model entirely from 3D-Var but for now the object is still created. So in the short term this is a good way to do it.

guillaumevernieres:
@ytremolet : Since we have your attention, could we put [CheckpointModel.h](https://github.com/JCSDA/soca/blob/develop/src/mains/CheckpointModel.h) in oops, instead of soca? Poor choice of name, but all it does is dump the analysis into model restarts. This is something that is needed for cice as well.

Issue #37, opened 2019-02-04
Geometry/Model consolidation


guillaumevernieres:
The model geometry is initialized in Model and Geometry, remove redundancy and consolidate.

PR #167, opened 2019-08-01
update MOM6 APIs required for MOM6 update

aerorahul:
Update soca interface to bring in MOM6 API changes.

The soca-bundle builds, but soca tests are currently all failing.  I suspect a change is required in `input.nml` and may be yml files.

Use following branches for FMS and MOM6
```
fms::dev/master-ecbuild-sync
mom6::dev/master-ecbuild-sync
```

Note: this is a **draft** PR. **Do not merge!**

Change-Id: Ic5bd9410b7da356a2d305d2623c46beb389b3146


aerorahul:
@guillaumevernieres @travissluka 
Any idea what might be wrong?
This has to do with the new MOM6 update and an inconsistency in our setup (namelist most likely).

```
❯❯❯ mpirun -np 4 ../../bin/mom6.x                                                                      
NOTE from PE     0: MPP_DOMAINS_SET_STACK_SIZE: stack size set to    32768.
&MPP_IO_NML
 HEADER_BUFFER_VAL=      16384,
 GLOBAL_FIELD_ON_ROOT_PE=T,
 IO_CLOCKS_ON=F,
 SHUFFLE=          0,
 DEFLATE_LEVEL=         -1,
 CF_COMPLIANCE=F,
 /
NOTE from PE     0: MPP_IO_SET_STACK_SIZE: stack size set to     131072.
NOTE from PE     0: MPP_DOMAINS_SET_STACK_SIZE: stack size set to  2000000.
NOTE from PE     0: ======== Model being driven by MOM_driver ========
NOTE from PE     0: callTree: o Program MOM_main, MOM_driver.F90
&OCEAN_SOLO_NML
 DATE_INIT=       2018,          4,         15, 3*0          ,
 CALENDAR="NOLEAP          ",
 MONTHS=          0,
 DAYS=          1,
 HOURS=          0,
 MINUTES=          0,
 SECONDS=          0,
 OCEAN_NTHREADS=          1,
 NCORES_PER_NODE=         36,
 USE_HYPER_THREAD=F,
 /
NOTE from PE     0: open_param_file: MOM_input has been opened successfully.

FATAL from PE     3: MPP_DEFINE_DOMAINS: number of points in X must be even when there is a fold in Y for MOMc

--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 3 in communicator MPI_COMM_WORLD
with errorcode 1.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
--------------------------------------------------------------------------

FATAL from PE     2: MPP_DEFINE_DOMAINS: number of points in X must be even when there is a fold in Y for MOMc


FATAL from PE     1: MPP_DEFINE_DOMAINS: number of points in X must be even when there is a fold in Y for MOMc

[gs6101-ice.local:55398] 2 more processes have sent help message help-mpi-api.txt / mpi-abort
[gs6101-ice.local:55398] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

guillaumevernieres:
@aerorahul : Looking at it ...

guillaumevernieres:
@aerorahul : It's a bit more involved than the last update! It's possible that we (as in @travissluka) will have to generate a grid with even number of cells (low-res GOM?), but that doesn't seem like the biggest issue. The 1deg setup uses even numbers, but the geometry test doesn't work (the test run but doesn't generate the grid), and mom6.x segfaults with a fresh copy of ".datasets" and the mom6 examples. 
Digging ...

guillaumevernieres:
@aerorahul @travissluka 
There's something fishy with our mom6 build:
-  soca test and mom6.x "work" (1 deg setup, running tests manually) in debug but segfaults in release.
- MOM6 exec built with mkmf (recent clone of MOM6-example) works fine with -O3 (same 1deg setup as above)
- The odd global NJ is probably a bug in soca, caught by a new error trap in mom6? Not that we really care since we weren't panning to do any science at 4 deg res. I'm assuming we'll use that opportunity to coarsen the grid some more and speedup our testing.
 
Looks like we'll be busy next week!

guillaumevernieres:
I'll have a separate PR addressing issue #37 , which I should have done ages ago. That will address some of the issues of this draft PR.

travissluka:
I'm looking into the "number of points must be even in X" issue, I'm suspecting a bug in MOM, because our NX is even for the 4 deg grid and breaking with standalone MOM

guillaumevernieres:
> I'm looking into the "number of points must be even in X" issue, I'm suspecting a bug in MOM, because our NX is even for the 4 deg grid and breaking with standalone MOM

:+1: 

I'm getting rid of the mom6 calls in the geometry init, full of memory leaks, overkill and the grid is generated a second time in the model class. 
I'm writing an application that will generate a grid and write it to file, the grid constructor will simply read that file. @travissluka @aerorahul , if I don't hear from you, I'll assume you are OK with the idea.

travissluka:
using the almighty "git bisect" I was able to trace back to a change to MOM6 that is causing part of the problems for us
https://github.com/NOAA-GFDL/MOM6/commit/7c4adcb96685db1bbb205a111677ccab9b832f90
I'm looking to see if there is a way around it, otherwise we'll have to make our grid's NI divisible by 4

travissluka:
@guillaumevernieres I might be okay with that, I'm assuming in that app you'll be making the calls to the usual MOM6  calls?

guillaumevernieres:
> @guillaumevernieres I might be okay with that, I'm assuming in that app you'll be making the calls to the usual MOM6 calls?

Yes, we'll have the same flexibility of setup, but we won't be carrying the memory burden ...

guillaumevernieres:
> using the almighty "git bisect" I was able to trace back to a change to MOM6 that is causing part of the problems for us
> [NOAA-GFDL/MOM6@7c4adcb](https://github.com/NOAA-GFDL/MOM6/commit/7c4adcb96685db1bbb205a111677ccab9b832f90)
> I'm looking to see if there is a way around it, otherwise we'll have to make our grid's NI divisible by 4

Ouch, that's a little limiting!

travissluka:
so, yeah, a change was made to `MOM_grid_init()` in which a second downsampled grid is constructed  HId2 (for reduced resolution diagnostics with 1/8th global models, from what I can gather from the git logs). Unless we can convince GFDL to do a PR making that optional, we'll have to make tripolar grids divisible in the x direction by 4. regional grids *should* be unaffected. This extra grid being carried around might be the cause of some of our other bugs now.

guillaumevernieres:
OK.
As an update, I can run the model with the 1 deg setup (mom6.x and model test), so we should be back up and running after our geometry fixes. Not exactly what we were planning to do this week!

travissluka:
I have a branch in soca (`feature/5deg_grid`) with a new grid that works with the latest MOM6 and with the current version of soca.
However, if I merge that branch into `feature/mom6-update` I get the following error when I try to run soca tests
```
FATAL from PE     0: compute_global_grid_integrals: zero ocean area (check topography?)
```

The model runs fine with `mom6.x` so I'm guessing there's a problem with initialization in soca somewhere. I give up for today if someone else wants to take a look at it with the new grid.

guillaumevernieres:
@travissluka : Can you branch out of develop? My geometry stuff isn't ready yet, but that will fix the issue you are seeing (your current branch should work in debug).

guillaumevernieres:
@travissluka @aerorahul 
Virtual high-five, I just did a quick test merging our branches and it works. Modulo the usual snafu in my branch, we should be done with that issue once @aerorahul and my branch are merged (@travissluka 's fix was merged last night). How's that for team work?

guillaumevernieres:
@aerorahul @travissluka I'm merging this branch with develop, I need to go through a few minor conflicts related to the last pr.

guillaumevernieres (approved):
Amazing work @aerorahul !!!

travissluka (approved):
...well that was fun

guillaumevernieres:
> ...well that was fun

Agreed. I'll make sure I'm on PTO next time we sync with GFDL.

aerorahul:
thanks guys!

PR #168, opened 2019-08-03
Interactive view of observation space

guillaumevernieres:
Added a script to view some obs stats interactively. Recycled from older work, a bit of a hack but works and can be quite useful. Based on wxpython.
oceanview
Left click of the mouth on one of the observations will show a profile plots of oma/omf/obs/fcst/ana ... 
Other options documented in the script.


aerorahul (approved):
so long as it works.

PR #169, opened 2019-08-06
Change grid resolution from 4deg to 5deg

travissluka:
The test grid files have been changed from 4 degrees to 5 degrees, and answers updated accordingly Most importantly this makes the number of grid points in the x axis divisible by 4, which is required for tripolar grids with the upcoming MOM6 update.


guillaumevernieres (approved):
All good.

PR #170, opened 2019-08-06
Grid "generation" application and some refactoring of geometry

### LINKED ISSUE #37: Geometry/Model consolidation

guillaumevernieres:
- Addition of an application that uses mom6 to generate a grid and write it to file
- The above grid is used in all/most of the test, the geometry will not be complete without reading this file
- Simplification of the geometry constructor
closes #37

### DIFF
--- a/src/soca/Geometry/soca_geom.interface.F90
+++ b/src/soca/Geometry/soca_geom.interface.F90
@@ -77,9 +77,6 @@ contains
 
     print *,'geom key = ',c_key_self
     call self%gridgen()
-    !call self%get_rossby_radius()
-    !call self%validindex() !BUG: Needs a halo of 2 to work
-    !call self%infotofile()
 
   end subroutine c_soca_geo_gridgen
 
--- a/src/soca/Geometry/soca_geom.interface.F90
+++ b/src/soca/Geometry/soca_geom.interface.F90
@@ -46,9 +46,6 @@ contains
     call soca_geom_registry%get(c_key_self,self)
 
     call self%init(c_conf)
-    call self%get_rossby_radius()
-    call self%validindex() !BUG: Needs a halo of 2 to work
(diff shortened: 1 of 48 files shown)


src/soca/Geometry/soca_geom.interface.F90
```diff
@@ -68,6 +65,24 @@ subroutine c_soca_geo_clone(c_key_self, c_key_other) bind(c,name='soca_geo_clone
...
+    integer(c_int), intent(inout) :: c_key_self
+    type(c_ptr),       intent(in) :: c_conf
+
+    type(soca_geom),      pointer :: self
+
+    call soca_geom_registry%get(c_key_self,self)
+
+    print *,'geom key = ',c_key_self
+    call self%gridgen()
+    !call self%get_rossby_radius()
```
travissluka: remove commented out code

src/soca/Run/soca_setup.f90
```diff
@@ -21,7 +21,7 @@ subroutine soca_setup(c_conf) bind(c,name='soca_setup_f')
   type(c_ptr), intent(in) :: c_conf
   type(fckit_mpi_comm) :: f_comm
 
-  f_comm = fckit_mpi_comm()
+  !f_comm = fckit_mpi_comm()
```
travissluka: what is the point of this file now, its essentially empty

travissluka (approved):

src/soca/Run/soca_setup.f90
guillaumevernieres: soca::Run doesn't do anything anymore. I'll remove the comment, but removal or adding features to the run class should not be part of this pr. Or should it?

PR #181, opened 2019-08-21
Activate Travis-CI for soca

mer-a-o:
this PR adds the files required by Travis-CI to the repo.


travissluka:
It looks like the only thing failing is the soca coding norms, which is odd because that passes on every other machine. I turned on verbose output for the build
https://travis-ci.com/JCSDA/soca/builds/124951063#L3437-L3483

@aerorahul any idea what's going on?

travissluka:
@mer-a-o It looks like the coding norm test passed with Travis-CI build number 3, but is failing as of build number 10. I'm not sure what it is in the travis config file that would cause that, but I don't think anything changed in the other repos during that time

mer-a-o:
Coding norm test passes now.  I changed .travis.yml back to cloning soca repo within yml file rather than using what was cloned automatically by Travis-CI.

travissluka (approved):
thanks @mer-a-o ! I'm assuming we're good to go now with this?

mer-a-o:
@travissluka, yes! ready to merge.

PR #183, opened 2019-08-29
updates for recent UFO/OOPS changes

travissluka:
This catches soca up with updates made to ufo/oops for the code sprint. This should work with oops/ufo develop branches.
After being merged into feature/removed_config_mod, https://github.com/JCSDA/soca/pull/172 should be able to be merged into develop... finally

### DIFF
--- a/src/soca/Traits.h
+++ b/src/soca/Traits.h
@@ -27,6 +27,7 @@
 #include "ufo/LinearObsOperator.h"
 #include "ufo/ObsBias.h"
 #include "ufo/ObsBiasCovariance.h"
+#include "ufo/ObsBiasIncrement.h"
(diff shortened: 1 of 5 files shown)


src/soca/Traits.h
```diff
@@ -26,8 +26,8 @@
 #include "ufo/GeoVaLs.h"
 #include "ufo/LinearObsOperator.h"
 #include "ufo/ObsBias.h"
-#include "ufo/ObsBiasIncrement.h"
```
shlyaeva, line 29: might still need this include

shlyaeva (commented):
thanks for catching up and sorry for breaking soca (especially for silently changing filters)

aerorahul (approved):
@travissluka thanks for the patch.

PR #184, opened 2019-08-29
increase tolerance on SOCA 3dvar and 3dhyb tests

aerorahul:
These tests fail on a Mac.
Increasing the tolerance on 3dvar and 3dhyb tests from 3e-5 to 2.3e-4 and 4…e-4 to 5e-4 respectively allows these tests to "pass" on a Mac. 
Default (stricter) tolerances will be used when issuing PR via TravisCI.

Change-Id: I095e10bcbdca483889baeb0818ca1e934bb0c262

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -417,7 +417,7 @@ soca_add_test( NAME enshofx
 # variational methods
 soca_add_test( NAME 3dvar_soca
                EXE  soca_3dvar.x
-               TOL  3.0e-5 0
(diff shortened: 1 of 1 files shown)


travissluka (approved):
great.
@aerorahul @guillaumevernieres Whenever we want to make sure the answers exactly match the reference answers run ecbuild with` -DSOCA_TESTS_FORC_DEFAULT_TOL=ON`, fortunately it seems travisCI produces identical answers to the ubuntu installations we are using on our own computers.

PR #185, opened 2019-08-30
ctest cleanup

### LINKED ISSUE #176: Remove 1 deg grid files from git-lfs
keep only the 5 deg grid files in github.
Remove the higher resolution 1 deg grid (to be placed on S3  at some point, along with the 1/4 deg and regional grid configs)

travissluka:
small cleanup of the ctests. 
* the data files for 1 deg configuration have been removed. They are not needed by the ctests. In the future, configurations higher than the current default 5 deg will be hosted directly on amazon. closes #176
* most of the remaining netcdf data files have been compressed. This, along with the above point, shrinks our git-lfs footprint from 1.7GB to ~ 12 MB... so when we run out of git-lfs bandwidth again there is no way they can blame us!
* travisCI tests will now ignore the compare.py tolerances we have specified and use a very tight tolerance of 1e-12 (by using the `-DSOCA_TESTS_FORC_DEFAULT_TOL=ON` cmake flag). This way we can still pick up on unexpected changes to answer, while having large tolerances that allow soca to pass on other machines.  This now means that the reference test files will need to be generated on travisCI, or on our linux workstations that produce identical answers to travisCI


aerorahul (approved):
@travissluka  This indeed speeds up the clone.  Thanks!
My only suggestion is we should be able to run any desired resolution.  To that end, either we update the CMakeLists.txt or add a script to stage the files in a runtime directory.
This will enable us to test different resolutions, on different machines quickly without having to hack through existing CMakeLists.txt files.

test/testref/3dvar_soca.test
```diff
@@ -22,11 +22,11 @@ Test     :     shf   min=  -58.949293   max=  201.374248   mean=    9.318213
 Test     :      lw   min=    0.000000   max=   88.036094   mean=   48.153834
 Test     :      us   min=    0.004396   max=    0.019669   mean=    0.013335
 Test     : CostJb   : Nonlinear Jb = 2.528383
-Test     : CostJo   : Nonlinear Jo(CoolSkin) = 19969.905666, nobs = 168, Jo/n = 118.868486, err = 0.359475
-Test     : CostJo   : Nonlinear Jo(SeaSurfaceTemp) = 655.918499, nobs = 113, Jo/n = 5.804588, err = 0.349115
+Test     : CostJo   : Nonlinear Jo(CoolSkin) = 19969.905667, nobs = 168, Jo/n = 118.868486, err = 0.359475
```
aerorahul, line 25: @travissluka why does removal of files, change the number?  Also, why should this matter, if we are adding a tolerance to the comparison test?  The change is in the 6th decimal place, and the tolerance is e-4.

travissluka: this is due to enabling tight tolerances on travisCI. Shastri generated that file when he added a qc test

travissluka:
@aerorahul I agree with having a cmake flag to let us set the desired resolution, and then pull the required model config files from S3 automatically, would be nice. The only annoyance is that we would have another set of yml and ref files we would have to keep up with.

aerorahul:
> @aerorahul I agree with having a cmake flag to let us set the desired resolution, and then pull the required model config files from S3 automatically, would be nice. The only annoyance is that we would have another set of yml and ref files we would have to keep up with.

@travissluka I agree that keeping another set of yml and ref files will be annoying.
We can use the `configure_file` feature of `cmake` to generate yml files based on resolution.
I don't think we should keep another set of ref files for different resolutions.  The point is not to test whether all resolutions produce identical answers (wrt a reference), but to be able to test different resolutions quickly.

guillaumevernieres (commented):
@travissluka : I'm seeing a bunch of ctest failures, despite the check on travis-ci. Likely a problem on my side, but don't merge that branch just yet!

travissluka:
@aerorahul sounds good. In that case, I don't think we would need to make any changes to the yaml files so long as we skip the interface tests for geometry/state/etc.. 

@guillaumevernieres ugh, welcome back to work

PR #186, opened 2019-09-03
remove soca::instantiateObsFilterFactory and use the ufo one

travissluka:
obs filter factory instantiation will be handled by ufo::instantiateObsFilterFactory<...>() , instead of by soca, once https://github.com/JCSDA/ufo/pull/567 is merged

requires the `feature/instantiateFilters` branch of ufo


aerorahul (approved):
@travissluka Thanks for the consolidation of the instantiation of observation filters.

PR #190, opened 2019-09-09
don't recreate bump geom on every getvalue call

travissluka:
This is a bit of a quick hack to get the getValues() times more reasonable.

On Discover, getValues() was taking **460s** with the 1/4 deg grid and about 300k observations from SST/ADT/insituT/S. The geometry in bump was being reinitialized every time getValues() was called (for a total of 12 times), and this is currently very slow un-parallelized code in bump. Now, the bump geometry is only initialized once, and a separate `type_obsop` class is carried around instead for the interpolation. The same configuration now only takes **50s** for getValues().

No answers should change 

Again, this is a bit of a quick hack.(there is a global variable in the fortran code, gasp!) We can get back to this and do it in a cleaner way once some of the interpolation is cleaned up in oops (e.g https://github.com/JCSDA/oops/pull/357)

### DIFF
--- a/src/soca/Utils/soca_bumpinterp2d_mod.F90
+++ b/src/soca/Utils/soca_bumpinterp2d_mod.F90
@@ -79,7 +79,7 @@ subroutine interp_init(self, mod_lon, mod_lat, mod_mask, obs_lon, obs_lat, bumpi
     bump%nam%obsop_interp = 'bilin'     ! Interpolation type (bilinear)
     bump%nam%default_seed = .true.
     bump%nam%new_obsop = .true.
-    bump%nam%verbosity = 'main'
+    bump%nam%verbosity = 'none'
     bump%nam%write_obsop = .false.
     bump%nam%sam_write = .false.
     bump%nam%strategy = 'specific_univariate'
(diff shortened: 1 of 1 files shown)


src/soca/Utils/soca_bumpinterp2d_mod.F90
```diff
@@ -32,6 +30,13 @@ module soca_bumpinterp2d_mod
    procedure :: finalize => interp_exit
 end type soca_bumpinterp2d
 
+! The bump grid here is stored separately from the interpolation weights
+! (obsop inside soca_bumpinterp2d) because the unstructured grid only needs
+! to be setup once
+! TODO: these shouldn't be globals, put them somewhere correct
+type(bump_type) :: bump
+logical         :: bump_initialized = .false.
+
```
guillaumevernieres, line 39: :+1: 
Nice @travissluka ! What's next? Common blocks? block data?  
:)

guillaumevernieres (approved):
Looks good, thanks.

src/soca/Utils/soca_bumpinterp2d_mod.F90
travissluka: yeah, but first step is to revert back to 6 character long variable names and put in a whole bunch of equivalence statements...

I could have done things cleanly, but I wanted to keep these temporary patches localized to a single file so that they're easier to rip out in the future.

PR #192, opened 2019-09-17
Use the partial initialization feature of mom6

guillaumevernieres:
Partial initialization was implemented in the mom6 initialization, but not used. This is required when using the "checkpoint" application and no forcing is present, which is he case when running the EMC coupled model.

### DIFF
--- a/test/testinput/checkpointmodel.yml
+++ b/test/testinput/checkpointmodel.yml
@@ -8,6 +8,7 @@ model:
   name: SOCA
   tstep: PT6H
   advance_mom6: 0
+  partial_update: 1
   variables: [cicen, hicen, hsnon, socn, tocn, ssh, hocn, sw, lhf, shf, lw, us]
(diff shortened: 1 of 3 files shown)


test/testinput/checkpointmodel.yml
```diff
@@ -8,6 +8,7 @@ model:
   name: SOCA
   tstep: PT6H
   advance_mom6: 0
+  partial_update: 1
```
travissluka, line 11: umm... soca is looking for `partial_init` but you put `partial_update` in the yml here

guillaumevernieres:
@travissluka , sorry, looks like that pr is a waste of time! I was trying to bypass the setup of the forcing, but that doesn't work when fixing the bug in the yaml that you mention. I may revisit in the future, but since we are going to have a model in the emc workflow next week, this shouldn't be an issue anymore.
Closing for now.

PR #195, opened 2019-09-26
revive develop

### LINKED ISSUE #194: Current develop branch broken
... because no one kept up with pending changes due to https://github.com/JCSDA/oops/pull/358

travissluka:
I'm able to get soca develop to compile again after the latest PR to oops (https://github.com/JCSDA/oops/pull/358) but I still can't get it to run due to socaerror_init failing with 

` !!! ABORT in io_fld_write on task #0002: NetCDF: Parallel operation on file opened for non-parallel access`

My gut is telling me that MOM6 is opening the restart file in non-parallel mode, which is preventing BUMP from opening it in parallel mode. Help would be appreciated in tracking this one down.

closes #194


travissluka:
... ok, so this branch will work once I update the compare tests.... so long as you have parallel netcdf working on your computer. @aerorahul do we have this on Discover?

danholdaway:
Discover is all good. Have been testing this latest saber there.

aerorahul:
> ... ok, so this branch will work once I update the compare tests.... so long as you have parallel netcdf working on your computer. @aerorahul do we have this on Discover?

@travissluka netCDF on Discover is built with the parallel option turned ON.

aerorahul (approved):
looks good.

guillaumevernieres:
@travissluka getting quite a few failures here, will dig.

travissluka:
@guillaumevernieres Most of the test failures are compare issues, but I'm getting test_soca_increment failing on one test for reasons I can't understand, and 3dvarfgat is hanging

guillaumevernieres:
@travissluka : Looks like the issue is in add incr, for some reason, saving the increment changes the results of the tlm test ...
if you comment out the saving of the incr, the increment test is OK
[comment out that stuff](https://github.com/JCSDA/soca/blob/2eb0260faa623796a7d934f4a021e6d18db46f30/src/soca/Fields/soca_fields_mod.F90#L538)

Still get failures elsewhere.

guillaumevernieres:
@travissluka : 
Reverting #190 and removing the save increment fixes the issues in increment and fgat.

travissluka:
I don't think the "save increment" was the problem, but you're right about PR #190 . I'll try tomorrow to revive the changes that were made in #190 , otherwise I'll revert it back to all of its gloryful slowness

guillaumevernieres:
> I don't think the "save increment" was the problem, but you're right about PR #190 . I'll try tomorrow to revive the changes that were made in #190 , otherwise I'll revert it back to all of its gloryful slowness

Sounds good. The "save incr" fixed the interp tlm test for me, I don't know why, maybe something fishy with fms.

travissluka:
This branch works now on my machine in the container. It unfortunately is killing Travis CI due to `Allocation would exceed memory limit`
@mer-a-o do you know what the memory limits are on TravisCI? I have a hard time believing we are suddenly bumping up against the limits without there being some other bug in the code i'm not aware of.

travissluka:
finally...
passes on travisci, desktop in container, and on discover/intel.
does not pass yet on discover/gcc due to needing much larger tolerances on the tests, I'll update those later after investigating #197

guillaumevernieres (approved):
Some issues on Hera that we'll sort out next week. All good otherwise. Thanks.

PR #196, opened 2019-09-27
Remove BUMP MPI calls

benjaminmenetrier:
See https://github.com/JCSDA/saber/pull/28#issuecomment-535855089 for details.


travissluka:
@benjaminmenetrier  we can't test this yet, we're still recovering from the previous saber PR. Our test references have increased to requiring suspiciously large tolerances on different machines, and given valgrind output i'm seeing there seems to be a "`Conditional jump or move depends on uninitialised values`" bug in either soca or saber. I'm trying to track down the problem

travissluka (approved):

guillaumevernieres (approved):

PR #198, opened 2019-09-27
enlarge stack size for tests

mmiesch:
@travissluka - hopefully this will solve the test failures you were seeing with Travis.  If you need to change any other system settings you can do it in a similar way.


travissluka:
@mmiesch turns out the root issue was the bug in https://github.com/JCSDA/saber/pull/31

Since the bug wasn't causing a problem on HPCs or my desktop, only travisci, perhaps it might be better to actually leave the stack a fixed size to keep catching those types of issues

mmiesch:
Ok - feel free to delete this PR & branch if you wish - I'll leave it up to you

travissluka:
closing. Thanks @mmiesch  for looking into it

PR #200, opened 2019-09-28
Cousin PR to UFO feature/refactor_locs_ctor

### LINKED ISSUE #199: prep soca for changes in ufo PR #654
need to change `increment_test.yml` for pending ufo PR (https://github.com/JCSDA/ufo/pull/654)

naive attempt following @shlyaeva 's recommendation isn't working. Test failure possibly related to https://github.com/JCSDA/soca/issues/197 ?  since i'm seeing:

```
16: Running case "interface/Increment/testIncrementInterpAD" ...                                                                                                                                                   
16: soca_fields:read_file: Not reading var cicen                                                                                                                                                                   
16: soca_fields:read_file: Not reading var hicen                                                                                                                                                                   
16: soca_fields:read_file: Not reading var cicen                                                                                                                                                                   
16: soca_fields:read_file: Not reading var hicen                                                                                                                                                                   
16: Random Locations vars: 0 variables:                                                    
16: Random Locations :  no output        
16: difference between 7.4888505246580905 and -28.053696064497402 exceeds 2.8053696064497402e-11 (FAIL)
```

guillaumevernieres:
closes #199 
I could be myopic (I am), but It doesn't look like we can specify locations anymore. I picked points in the tropical pacific, which featured an abundance of sea-ice in earlier version of soca.

### DIFF
--- a/test/testinput/increment_test.yml
+++ b/test/testinput/increment_test.yml
@@ -12,8 +12,20 @@ TestDate: &date 2018-04-15T00:00:00Z
 InterpTest:
   tolerance: 1e-12
   Locations:
-    lats: [  38.5,   0.0,   0.0,   2.0,  -2.0,  -2.0,   2.0, 89.0, -67.0]
-    lons: [ -14.5, -15.0, -25.0, -15.0, -25.0, -15.0, -25.0,  0.0,   0.0]
+    window_begin: 2018-04-14T00:00:00Z
+    window_end: 2018-04-15T00:00:00Z
+    ObsSpace:
+      name: Random Locations
+      simulate:
+        variables: []
+      Generate:
+        nobs: 10
+        lat1: -20
+        lat2: 20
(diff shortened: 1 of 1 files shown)


travissluka (approved):
Fine for now so that the UFO PR can go ahead.
However, it make me uncomfortable testing such a small area, I foresee uncaught bugs in the future. Ideally we would have global coverage of locations, including spots near the coast, where there is ice, etc... Either by using 
1) a random distribution that can throw out locations that are on land or (preferable)
2) being able to hard code a list of lat, lons again

@shlyaeva do you know if there is any easy way to do this?

shlyaeva:
@travissluka: you can use an existing observation file that has good enough coverage (if you have one). lats/lons from that file would be used. If you don't have one, you could put your hardcoded values in the obs-like file (it would need to have: lon, lat, datetime) and use it.

PR #208, opened 2019-10-03
add mpi comm to geometry

travissluka:
Add mpi communicator to the geometry constructor, in preparation for oops PR https://github.com/JCSDA/oops/pull/425

needs to be tested with oops branch of matching name


guillaumevernieres (approved):
Thanks @travissluka , needs to be completed by #210 ... one day.

aerorahul (approved):
ok!

PR #211, opened 2019-10-05
Removed empty Run class

ytremolet:
No need for classes that do nothing, and that might cause problems down the road. 

And I've always dreamt I would be able to contribute to soca! :-)


aerorahul (approved):
Thanks for that work @ytremolet.
Finally the marine side gets some JEDI-love.

guillaumevernieres (approved):
Good job @ytremolet !

Student certificate

ytremolet:
Very proud. I will frame this one!

shlyaeva:
I want one too!

PR #218, opened 2019-10-10
Add floating point error trapping to soca tests

### LINKED ISSUE #217: Enable sigfpe trapping for some soca tests
https://github.com/JCSDA/oops/pull/420 will enable optional trapping of floating point errors. Once merged, we should start enabling this for soca tests that are currently okay with that (interface tests?). Then we can note where there are bugs for us to fix, and where the MOM6 model is just being grumpy and should be ignored.

travissluka:
Adds the floating point error trapping that was just added to OOPS to some of our soca tests. Tested with both GCC and on Discover with Intel17
closes #217 

* The one glaring FPE in soca in the balance operator is fixed, even though this was not having any impact on answers
* the interface tests that were using `ecbuild_add_test` were moved to using the `soca_add_test` wrapper. Test yaml files for those tests were renamed to fit into the standard we were imposing of `config_file = testinput/<test_name>.yaml`
* by default tests will enable TRAPFPE, unless the `NOTRAPFPE` flag is used, or the 
`-DSOCA_TESTS_FORC_TRAPFPE=ON` cmake flag is set, trapping is off for executables run outside of ctest

Trapping is enabled for all tests that were passing on both intel17 and gcc (intel seems to cause more problems inside MOM than gcc does). In a perfect Utopian world the rest will be fixed and re-enabled... by someone.. maybe... or not... probably not... we'll see


guillaumevernieres (approved):
:+1:

PR #219, opened 2019-10-11
discover workflow

travissluka:
These are the scripts I've been using to do cycling experiments on Discover. Just wanted to get these checked in (since i've been promising to do that for a month) before I use them to do any actual science. I'm sure things will change slightly once I start really using them.

These are the job subscripts, and slurm cycling script. The configuration files are in the `soca-config` repo in a related PR (https://github.com/JCSDA/soca-config/pull/2). The configuration binary files are available on the shared soca google drive, link provided in the readme.

Up to you if you actually want to test it on Discover, or you can just take my word for it.


guillaumevernieres (approved):

aerorahul (approved):
approving now.
will test later.

PR #220, opened 2019-10-11
disable git lfs for saber

mer-a-o:


travissluka:
@mer-a-o `git lfs install --skip-smudge` might do the trick, instead of `git lfs uninstall`

aerorahul:
@mer-a-o 
Also, 
`export GIT_LFS_SKIP_SMUDGE=1`
before a `git clone` also skips downloading lfs files.

To re-enable
`export GIT_LFS_SKIP_SMUDGE=0`

travissluka:
it turns out soca tests dont depend on binary files from any other repo. So we should be able to skip the lfs for all non-soca jedi repos (oops/crtm/ioda/ioda-converters/ufo)
Should be a big speedup!

mer-a-o:
Thanks @travissluka and @aerorahul ! Looks like it worked. Can we disable git lfs for any other repo in soca-bundle?

travissluka:
awesome :+1: That should help out quite a bit with the git-lfs bandwidth issues if done to all the other projects.

mer-a-o:
@travissluka: good idea. This PR is ready to merge then.

travissluka (approved):

aerorahul (approved):
not sure how to test this, but they look right.

PR #223, opened 2019-10-18
Cached build of repos when doing TravisCI build

travissluka:
_Pay no attention to this for now, testing ideas for doing caching on TravisCI_

but if you must know...

Previously, running the tests in debug and with code coverage was timing out because it took so long (well over an hour). Now a TravisCI test take ~7 minutes if the cache is already populated, ~25 minutes otherwise for a full rebuild. :tada: 

Instead of running the tests on TravisCI by compiling everything that would normally be in the bundle, each upstream repository is built separately, and then installed to a cached location doing `make install`. (This required changes to quite a few repos to make sure the install works correctly). On subsequent builds this cached install directory is used instead of rebuilding. The repos are rebuilt  if the scripts detect any of several changes (different repo branch name, different git hash tag, an upstream repo was rebuilt, change in order of repo dependencies.... ). The soca repo is always rebuilt every time.

The `ccache` utility is used to greatly accelerate compiling the C files even when an entire repo needs to be rebuilt. This of course only accelerates the C files, not Fortran, but it's still useful.

By doing the compilation separately this way, the build type can be changed for each repo. Now, the upstream repos are all built with `RelWithDebInfo` to help run the tests faster, and soca is built with `Debug` to catch more errors. Code coverage instrumentation usually slows down the code, but now it can be enabled for soca only, and not all the upstream repos. We can choose to instead build all repos, or a subset of the repos, in debug for a longer nightly/weekly cron run.

Requires:
* https://github.com/JCSDA/CVMix-src/pull/1
* https://github.com/JCSDA/geoKdTree/pull/1
* https://github.com/JCSDA/MOM6_DA_hooks/pull/1
* https://github.com/JCSDA/GSW-Fortran/pull/7
* https://github.com/JCSDA/MOM6/pull/7
* https://github.com/JCSDA/ufo/pull/717
* https://github.com/JCSDA/oops/pull/441


guillaumevernieres (approved):
Nice!

PR #226, opened 2019-10-22
Cached build of repos when doing TravisCI build 

travissluka:
Enable caching for the TravisCI builds, as well as some general cleanups

* code coverage is added
* codecov and travisci badges added to the main soca page
* `ccache` tool is used to accelerate CXX files between builds
* upstream repos are built separately, and cached, reducing the code that needs to be built if dependencies don't change between builds. Repos that rarely change (model, fckit) are first in the list (so, for example, if OOPS changes, we don't have to rebuild all the repos upstream of it)
* upstream repos are compiled in `RelWithDebInfo`, soca repo in `Debug`
* PR tests are handled differently now. Before, the soca branch was used in the tests. Now, the branch _after_ it has been merged with develop is used. This is the default behavior of TravisCI if we build with the branch it automatically checks out. So now if the PR tests pass, we can assume develop will be okay after the merge, we don't have to keep merging develop into the test branch, just rerun the test.
* some of the scripting in `.travis.yml` has been pulled out into individual scripts in `.travis/`, these scripts _should_ be generalized enough to work for any project, so long as the relevant variables in the `env:` section of `.travis.yml` are changed. I'll be adding more scripts in this directory in the near future for features such as using the nightly cron run to update the `relase/stable-nightly` branch of the bundle
* a couple of tiny changes to SOCA so that it will work when upstream repos are compiled outside the main bundle.
* in addition to testing the PR branches and `develop`, travisCI will also test any branch that starts with the name `travisci_` (useful to test changes to TravisCI without having to do a draft PR)


travissluka:
this PR is ready to look at now, I'm done making changes.

guillaumevernieres (approved):
Thanks @travissluka ! 
Having the automated merge of develop before running the test is pretty cool.  I assume this is only done on the testing side, not on the branch itself?

travissluka:
I'm going to go ahead and merge this, mainly because the other soca TravisCI tests are currently failing this morning because the pesky lcov.tar.gz file can't be found. 

@mer-a-o, @aerorahul  If you have any issues or comments about this PR, let me know and we can address afterward

PR #228, opened 2019-10-28
fix thinning filters

### LINKED ISSUE #227: thinning, difference check and gaussian thinning filters yaml changes
I've just merged an update to ufo that will break experiments where Thinning, Gaussian Thinning or Difference Check filters are used.
Yamls for those filters now require explicitly specifying which variables are filtered out.
If you use any of those filters, please change the yamls. See examples in https://github.com/JCSDA/ufo/pull/718/files

travissluka:
small changes to yaml due to change in ufo
closes #227

### DIFF
--- a/test/testinput/3dvar_godas.yml
+++ b/test/testinput/3dvar_godas.yml
@@ -132,6 +132,7 @@ cost_function:
         - variable: sea_area_fraction@GeoVaLs
           minvalue: 0.5
       - Filter: Thinning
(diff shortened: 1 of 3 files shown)


aerorahul (approved):
Looks good.

travissluka:
going ahead and merging, test is failing because TravisCI is being cranky (will merge the updated TravisCI stuff right after this)

PR #235, opened 2019-10-31
Apply landmask to fields for gpnorm() and random()

### LINKED ISSUE #114: gpnorm with mask
The current gpnorm is only taking account of the mask when counting the # of grid points. We need to constrain the gpnorm stats to the unmasked values.

### LINKED ISSUE #231: mask land in Fields::random()
Currently, randomly generated fields from `Fields::random()` have random values everywhere. We should be masking out the land and setting those to 0. This will be needed so that the `GeometryIterator` test being added in #225 passes if the `GeometryIterator` skips over land (which we'll want to do, since there is no point doing the LETKF on land for the ocean analysis)

related somewhat to #114

travissluka:
* applies the land mask for `Increment::random()`, land locations are set to zero for the random increment. This is needed so that the `GeometryIterator` test will pass if the soca iterator skips over land points. This currently does not change the answers of any of our tests
* instead of doing the same thing for `ones()`, I just removed that method because it was not being used and is not required by the oops interface. (Free bonus code coverage points!!)
* the same masking is applied to `gpnorm()`, min/max/mean values now exclude land points (e.g. look at the minimum salinity, which is no longer 0.0). Test reference answers are updated because of this.

closes #114 
closes #231

### DIFF
--- a/src/soca/Fields/soca_fields_mod.F90
+++ b/src/soca/Fields/soca_fields_mod.F90
@@ -799,7 +799,9 @@ subroutine gpnorm(fld, nf, pstat)
   integer,                 intent(in) :: nf
   real(kind=kind_real), intent(inout) :: pstat(3, nf) !> [min, max, average]
 
+  logical :: mask(fld%geom%isc:fld%geom%iec, fld%geom%jsc:fld%geom%jec)
   real(kind=kind_real) :: ocn_count, tmp(3)
+  real(kind=kind_real) :: local_ocn_count
   integer :: jj,  myrank
   integer :: isc, iec, jsc, jec
   type(fckit_mpi_comm) :: f_comm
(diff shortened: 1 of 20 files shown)


guillaumevernieres (approved):
Approving but also noting that we lost our edge on the development of 1's Id and nothing classes ... sad.

src/soca/Fields/soca_fields_mod.F90
```diff
@@ -130,8 +130,8 @@ end subroutine create_copy
 ! ------------------------------------------------------------------------------
 
 subroutine soca_field_alloc(self, geom)
```
frolovsa, line 132: doesn't look like input geom is used in this subroutine

frolovsa (commented):
Looks good.

src/soca/Fields/soca_fields_mod.F90
guillaumevernieres: Welcome @frolovsa !
Not sure what you mean? see L141-142, 154, 157

frolovsa: @guillaumevernieres . Ok. i see now. Sorry about that. Using this new environment is like diving in a muddy river. You remember how to breath underwater but you can only see 3 inches in front of your face.  Thank you guys for your patience.

guillaumevernieres: No worries @frolovsa , thanks for looking at the code. It's not bug-free for sure, so don't be shy and ask/post if something looks fishy.

PR #236, opened 2019-11-01
fix interpolation using the kdtree

### LINKED ISSUE #197: enspert: difference answers on different machines
The tolerance needed on answers for enspert seems suspiciously large. Valgrind is throwing memory errors on `soca_enspert.x` when attempting to do an increment on the ice fields due to uninitialized variables.

```
==2842== Conditional jump or move depends on uninitialised value(s)
==2842==    at 0x4F2E26F: __soca_seaice_mod_MOD_soca_seaice_add_incr (soca_seaice_mod.F90:203)
==2842==    by 0x4EB50E1: __soca_fields_mod_MOD_add_incr (soca_fields_mod.F90:535)
==2842==    by 0x4E88F2C: soca_field_add_incr_f90 (soca_fields.interface.F90:253)
==2842==    by 0x4E849A6: soca::Fields::add(soca::Fields const&) (Fields.cc:174)
==2842==    by 0x4F44618: soca::State::operator+=(soca::Increment const&) (State.cc:102)
==2842==    by 0x48FDCB: oops::State<soca::Traits>& oops::operator+=<soca::Traits>(oops::State<soca::Traits>&, oops::Increment<soca::Traits> const&) (Increment.h:128)
==2842==    by 0x48B67A: oops::GenEnsPertB<soca::Traits>::execute(eckit::Configuration const&) const (GenEnsPertB.h:92)
==2842==    by 0x5596DA2: oops::Run::execute(oops::Application const&) (Run.cc:67)
==2842==    by 0x47D37F: main (GenEnsPertB.cc:17)
```

I did notice in the runtime output:
```
soca_fields:read_file: Not reading var cicen
soca_fields:read_file: Not reading var hicen
```

which would explain why it's complaining about uninitialized variables. I suspect this may be a cause of different answers on different machines. @guillaumevernieres any idea what I should do here? Turn off ice vars? figure out why it's not loading in the ice vars?

travissluka:
It turns out, after many months of us scratching our heads, that the large tolerances required in the ctests on Discover/theia were because of a quirk in the way the eckit kdtree was behaving. The kdtree was used for remapping the rossby radius and godas_bkgerr fields. If two points in the nearest neighbor search were the same distance away, there was not a guarantee which point would be used, depending on the machine.

I have implemented an n-nearest neighbor inverse distance weighted interpolation (modified Shepard's method in `soca_remap_idw()`) that uses the 9 nearest points. If any of the furthest points are the same distance away, they are thrown away. This helps ensure reproducibility across machines. And, we have a nicer interpolation for those fields now.

Most of the large tolerances have been completely removed, a couple of (much smaller) tolerances remain to ensure all tests pass on Discover gcc/intel17. These tolerances can be further improved in a subsequent PR by changing the way the state/increment norms are displayed (they are currently rounded, and so some differences are due to only the last digit).

Let me know if I need to increase any tolerances needed by Rahul's mac or hera.

closes #197 ... finally


guillaumevernieres (approved):
Looks good! 
All tests are OK on Hera with intel 19

travissluka:
awesome! I'll merge after I get a chance to do a quick check with the 1/4 to see how the rossby radius field looks now.

guillaumevernieres:
> awesome! I'll merge after I get a chance to do a quick check with the 1/4 to see how the rossby radius field looks now.

I'm on it. Running on hera right now, will post figures in a few mn.

PR #237, opened 2019-11-01
Ensemble recentering application

guillaumevernieres:
Why?
-------
Re-centering of the LETKF members is needed for the hybrid EnVAR in implementation at EMC. 

Dependency
---------------
Implementation of the ensemble re-centering app from [oops pr#452](https://github.com/JCSDA/oops/pull/452)
See [oops issue #451](https://github.com/JCSDA/oops/issues/451) for "more" details.

What this app does:
------------------------
- Scratches your back
- Re-centers an ensemble around your favorite state

See example of usage in the associated test. 

Check with gcc and intel and different HPC:
- [x] "TOL" is needed ? 
- [x] After #238 is merged, check if NOTRAPFPE is needed

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -11,6 +11,7 @@ list( APPEND soca_test_input
   testinput/dirac_socahyb_cov.yml
   testinput/enshofx.yml
   testinput/enspert.yml
+  testinput/ensrecenter.yml
   testinput/ensvariance.yml
   testinput/forecast_identity.yml
   testinput/forecast_identity_cice.yml
@@ -50,6 +51,7 @@ list( APPEND soca_test_ref
   testref/dirac_socahyb_cov.test
   testref/enshofx.test
   testref/enspert.test
+  testref/ensrecenter.test
   testref/ensvariance.test
   testref/forecast_identity_cice.test
   testref/forecast_identity.test
@@ -379,6 +381,12 @@ soca_add_test( NAME ensvariance
 soca_add_test( NAME parameters_bump_loc
                EXE  soca_parameters.x
                NOCOMPARE
+               TEST_DEPENDS test_soca_enspert )
+
+soca_add_test( NAME ensrecenter
+               EXE  soca_ensrecenter.x
+               TOL  3.0e-4 0
+               NOTRAPFPE
                TEST_DEPENDS test_soca_enspert )
 
 # TODO the next couple of tests are a little slow.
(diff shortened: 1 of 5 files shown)


test/CMakeLists.txt
```diff
@@ -379,6 +381,12 @@ soca_add_test( NAME parameters_bump_loc
                NOCOMPARE
                TEST_DEPENDS test_soca_enspert )
 
+soca_add_test( NAME ensrecenter
+               EXE  soca_ensrecenter.x
+               TOL  3.0e-4 0
+               NOTRAPFPE
```
travissluka: I'll try on discover, but the extra large tolerance and disabling fpe trapping shouldn't be needed

travissluka: okay, I was wrong, turns out floating point exceptions do occur, darn ice. Should be addressed separately, eventually. 

Large tolerance should still be removed if possible.

guillaumevernieres: @travissluka ... I do have a fix somewhere for sea-ice. I'll have a look tomorrow.

travissluka: Is the sea ice issue just that the halo for ice is not initialized or updated?

travissluka:
Ocean variables look good, I checked to make sure that the recentered ensemble mean matches the output from the identity forecast. The mean doesn't match for `aice` though, would that be expected?
Screenshot from 2019-11-05 08-29-54

guillaumevernieres:
> Ocean variables look good, I checked to make sure that the recentered ensemble mean matches the output from the identity forecast. The mean doesn't match for `aice` though, would that be expected?
> Screenshot from 2019-11-05 08-29-54

@travissluka ... man, I thought you didn't care about sea-ice!!!!
Yes, expected, we check for out of bound values for cicen inside of add increment.

guillaumevernieres:
@travissluka , Just tested on Hera. You are correct, the default small tolerances works.

travissluka (approved):

ytremolet:
Just merged the oops PR for recentering.

travissluka:
@guillaumevernieres I think this is ready to merge then. It seems we can't remove the NOTRAPFPE yet until other FPEs are addressed

guillaumevernieres:
> @guillaumevernieres I think this is ready to merge then. It seems we can't remove the NOTRAPFPE yet until other FPEs are addressed

@travissluka merging #238 first might be a good idea

guillaumevernieres:
> @guillaumevernieres I think this is ready to merge then. It seems we can't remove the NOTRAPFPE yet until other FPEs are addressed

Ha ... I obviously can't/don't read!

PR #255, opened 2019-11-09
Atmospheric Geometry (obs space only)

### LINKED ISSUE #246: Create an atmospheric Geometry in obs space
**Create an optional Atmospheric Geometry in Obs Space**
- Get "grid" size, lat, lon from obs file
- Pass atm grid spec to the soca::Geometry
- Adjust Geometry methods as necessary

**Tests**
- [ ] Backward compatible: Create ocean ice only geometry (no atmosphere)
- [ ] Exercise the new feature and create the atm geometry in addition to the ocean and ice

guillaumevernieres:
## Description
On the road to "NSST"! (although we're starting with SSS).
This feature branch adds a simple optional atmospheric geometry to soca::Geometry, based on ioda observation files. 
It uses ioda::ObsSpace for the distribution, unlike what might have been stated earlier, there will be no attempt at trying to match atm grid points to the corresponding PE domains.

Is a change of answers expected from this PR? **NO**

### Issue(s) addressed
- fixes #246 

### Other unrelated work done
- Removed redundant geometry interface in Fortran.h
- Removed "shoreline" and its siblings from the geometry

## Testing
How were these changes tested? **ctest + tested on larger atm obs file with more pe's**
What compilers / HPCs was it tested with? **gcc only**
Are the changes covered by ctests? (If not, tests should be added first.) **YES geometryatm test**

## Dependencies
None

### DIFF
--- a/src/soca/Fortran.h
+++ b/src/soca/Fortran.h
@@ -55,16 +55,6 @@ namespace soca {
 
   extern "C" {
     // -----------------------------------------------------------------------------
-    //  Geometry
-    // -----------------------------------------------------------------------------
-    void soca_geo_setup_f90(F90geom &, const eckit::Configuration * const *);
-    void soca_geo_clone_f90(const F90geom &, F90geom &);
-    void soca_geo_info_f90(const F90geom &);
-    void soca_geo_gridgen_f90(const F90geom &,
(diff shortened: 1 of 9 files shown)


guillaumevernieres:
I need to rethink that, closing for now.

Issue #254, opened 2019-11-09
Create an optional Atmospheric Increment/State in Obs Space


guillaumevernieres:
## Description
Create the obs space state and increment required to be able to exercise the CRTM UFO from soca.  

## Solution
State: Read from files dumped by an atmospheric model (testing with fv3-jedi).
Increment: Scratch your head ... I don't think there's anything that needs to be done, beyond implementing an other identity operator in getvalue TL/AD ... 

## Related to
Requires #246 to be sorted out.

Issue #248, opened 2019-11-07
Test CRTM UFO with GMI and SMAP (3D)


guillaumevernieres:
... And discover new issues.

- Add to the unit testing in 3dhofx, 3dvar (and ensemble). Use existing geovals/obs files from ufo. 
- Test with more obs

Issue #247, opened 2019-11-07
Pass necessary GeoVaLs to CRTM ufo in getvalue (NL/TL/AD)


guillaumevernieres:
**NL getvalue**
Read the atmospheric geovals from the output of the "fv3jedi_hofx" application when using the crtm ufo.

**TL/AD**
Only consider sst and sss for now.

guillaumevernieres:
@HamidehGMAO : Sorry, that was cryptic, no clue what I meant. I'm closing the issue since this work is done.

PR #259, opened 2019-11-20
Use of CRTM UFO

### LINKED ISSUE #246: Create an atmospheric Geometry in obs space

guillaumevernieres:
## Description
**Oil spill is contained!** None of the atmospheric variables are mentioned anywhere in the code (except yaml files), so far ...
Revisiting aborted [PR#255](https://github.com/JCSDA/soca/pull/255.) ... and more. This feature branch adds the possibility of using the CRTM ufo in hofx3d (not tested yet with hofx):
- Atmospheric geovals are pre-computed (fv3-jedi) and read inside of getvalue
- Ocean surface (just sst for now) is provided by soca through getvalue.

Is a change of answers expected from this PR? **NO but overflow issue in adt hofx. No clue why. @travissluka and @aerorahul said they'll fix it.**

### Issue(s) addressed
- closes #246 
- Partially  addresses #248, #247, #254 

### What's not too flash
The atmospheric "geometry" and "state" are part of the soca geometry. Not pretty but I can't think of a better way that would work.

## Testing
How were these changes tested? **hofx3dcrtm and geometryatm ctest**
What compilers / HPCs was it tested with? **gcc only**
Are the changes covered by ctests? (If not, tests should be added first.) **YES geometryatm test**

## Dependencies
Need the CRTM

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,10 +66,6 @@ include_directories( ${IODA_INCLUDE_DIRS} )
 ecbuild_use_package( PROJECT ufo VERSION 0.1.0 REQUIRED )
 include_directories( ${UFO_INCLUDE_DIRS} )
 
-# ufo
-ecbuild_use_package( PROJECT ufo VERSION 0.1.0 REQUIRED )
-include_directories( ${UFO_INCLUDE_DIRS} )
-
 # fms
 ecbuild_use_package( PROJECT fms REQUIRED )
 include_directories( ${FMS_INCLUDE_DIRS} )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,6 +66,10 @@ include_directories( ${IODA_INCLUDE_DIRS} )
 ecbuild_use_package( PROJECT ufo VERSION 0.1.0 REQUIRED )
 include_directories( ${UFO_INCLUDE_DIRS} )
 
+# ufo
+ecbuild_use_package( PROJECT ufo VERSION 0.1.0 REQUIRED )
+include_directories( ${UFO_INCLUDE_DIRS} )
+
 # fms
 ecbuild_use_package( PROJECT fms REQUIRED )
 include_directories( ${FMS_INCLUDE_DIRS} )
@@ -78,6 +82,10 @@ include_directories( ${MOM6_INCLUDE_DIRS} )
 ecbuild_find_mpi( COMPONENTS CXX Fortran REQUIRED )
 ecbuild_include_mpi()
 link_libraries(${MPI_CXX_LIBRARIES} ${MPI_Fortran_LIBRARIES})
+
+# crtm
+ecbuild_use_package( PROJECT crtm VERSION 2.2.3 )
+include_directories( ${CRTM_INCLUDE_DIRS} )
 
 ################################################################################
 # Definitions
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,7 +81,6 @@ link_libraries(${MPI_CXX_LIBRARIES} ${MPI_Fortran_LIBRARIES})
 
 # crtm
 ecbuild_use_package( PROJECT crtm VERSION 2.2.3 )
-include_directories( ${CRTM_INCLUDE_DIRS} )
 
 ################################################################################
 # Definitions
--- a/src/soca/Fields/soca_interpfields_mod.F90
+++ b/src/soca/Fields/soca_interpfields_mod.F90
@@ -151,13 +151,15 @@ subroutine interp(fld, locs, vars, geoval, horiz_interp, traj)
      call nlev_from_ufovar(fld, vars, ivar, nval)
 
      ! Allocate GeoVaLs (fields at locations)
-     geoval%geovals(ivar)%nval = nval
-     if (.not.(allocated(geoval%geovals(ivar)%vals))) then
-        ! Number of obs in pe
-        nlocs = geoval%geovals(ivar)%nlocs
+     if (nval/=0) then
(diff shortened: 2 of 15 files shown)


travissluka:
tests are failing because CRTM is not added to travisCI (so, ha, both Travises dont want CRTM).

Let me do a quick PR first that is needed for letting travisCI config set the bundle-wide ecbuild options .... while we continue to debate the merits of requiring CRTM.

I vote for adding in if statements to the cmakelist so that the crtm tests are skipped if crtm is not present (see the way ufo ctests are setup). We'll have crtm on by default on travisci... and off by default on my and @aerorahul 's laptops. everyone wins

travissluka:
@guillaumevernieres yeah, that hofx is expected to fail on intel. I don't know what i was thinking enabling SIGFPE trap on that given that the model runs for hofx.

travissluka (changes requested):
after JCSDA/soca/pull/260 gets merged, modify the CMakeLists.txt to wrap the ctests that need CRTM with `if( ${CRTM_FOUND} )`

guillaumevernieres:
@travissluka I must have screwed up somewhere, the crtm test didn't run. I'll look at it later tonight or tomorrow.

travissluka:
I think the problem is that soca isn't seeing that crtm is available because it isn't listed as a dependency in the main `CMakeLists.txt`. This doesn't matter if building as a normal bundle since ufo pulls in crtm, but the TravisCI tests build each repo separately.

try adding the line
`ecbuild_use_package( PROJECT crtm VERSION 2.2.3)`

Also, there's probably no harm in leaving the `geometryatm` test enabled even if crtm is not present

guillaumevernieres:
@travissluka OK, trying again.

CMakeLists.txt
```diff
@@ -66,6 +66,10 @@ include_directories( ${IODA_INCLUDE_DIRS} )
 ecbuild_use_package( PROJECT ufo VERSION 0.1.0 REQUIRED )
 include_directories( ${UFO_INCLUDE_DIRS} )
 
+# ufo
+ecbuild_use_package( PROJECT ufo VERSION 0.1.0 REQUIRED )
+include_directories( ${UFO_INCLUDE_DIRS} )
+
```
travissluka: oooh, double UFO !

travissluka:
@guillaumevernieres can't find the binary file now... because we aren't building everything as a single bundle. I had the same issue with files from oops, I'll lookup the syntax needed and get back to you...

guillaumevernieres:
Still no cigar, but closer. Looks like the .bin for the crtm were not found ...  Ha, just seeing your post above.  Nice, thanks for your help.

CMakeLists.txt
guillaumevernieres: dang, I knew the copy paste went badly ... Will wait until you send me a solution for the lfs files before pushing the fix.

guillaumevernieres:
Thanks @travissluka

CMakeLists.txt
```diff
@@ -79,6 +79,10 @@ ecbuild_find_mpi( COMPONENTS CXX Fortran REQUIRED )
 ecbuild_include_mpi()
 link_libraries(${MPI_CXX_LIBRARIES} ${MPI_Fortran_LIBRARIES})
 
+# crtm
+ecbuild_use_package( PROJECT crtm VERSION 2.2.3 )
+include_directories( ${CRTM_INCLUDE_DIRS} )
```
travissluka:
```suggestion
```

src/soca/State/State.cc
```diff
@@ -84,7 +84,17 @@ namespace soca {
   void State::getValues(const ufo::Locations & locs,
                         const oops::Variables & vars,
                         ufo::GeoVaLs & cols) const {
+    if (fields_->geometry()->getAtmInit())
+      {
+        // Get atm geovals
+        // The variables in vars that are also defined in soca will be
+        // over-written in the interpolation call bellow
+        getValuesFromFile(locs, vars, cols);
+        Log::trace() << cols << std::endl;
```
travissluka: remove, or make the output informative

src/soca/State/State.cc
```diff
@@ -84,7 +84,17 @@ namespace soca {
...
+        }
+    // Get ocean geovals
     fields_->getValues(locs, vars, cols);
+    Log::trace() << cols << std::endl;
```
travissluka: same as above

travissluka (changes requested):

src/soca/Fields/soca_interpfields_mod.F90
```diff
@@ -188,6 +190,11 @@ subroutine interp(fld, locs, vars, geoval, horiz_interp, traj)
      case ("sea_surface_temperature")
         fld3d(isc:iec,jsc:jec,1) = fld%tocn(isc:iec,jsc:jec,1)
 
+     ! TODO: Move unit change elsewhere, check if surface_temperature_where_sea
```
travissluka, line 193: might want to create a github issue to remind us.... we had discussed before that all of our ocean temp variables should be in K to comply with CF conventions. But of course that requires changing ufo at the same time... a good project for when someone is bored and doesn't want to do real work.

travissluka (approved):
Less of a Frankenstein than I was expecting 👍 

I tested on intel19 as well and all looks good.

PR #262, opened 2019-11-21
Fix all Intel19 debug runtime/compiler warnings

travissluka:
## Description

This fixes all soca runtime and compiler warnings for Intel19 when building in debug, which should make debugging in the future a little easier.

- All unused fortran variables were removed. We had accumulated quite a few.
- "`array temporary created`" runtime warnings are fixed in two locations,  `getvalues_ad()` and `soca_fields_mod::read_file()`.  In addition to suppressing the runtime warnings, this should help slightly with the runtime performance, especially with `getvalues_ad()`
- while I was in a cleaning mode I deleted some functions that are not used and are likely not going to be used, should get a nice small bump in codecov.
  - `random_vectors_mod` - there are routines in oops for random numbers
  - geometry print routines - these are not used, and are also not implemented correctly, will reimplement a proper print for geometry later

There is no change in answers.

I'm still seeing that answers in intel19 debug mode are not matching intel19 release and gcc7 debug/release. I think there is something wrong with the ice mask used in `soca_static_socaerror_init` ... another issue for another PR...

## Testing

- Answers are unchanged for GCC7 debug/release and Intel19 debug/release on my desktop.
- `test_soca_increment` and `test_soca_state` no longer produce runtime warnings with Intel19 debug


guillaumevernieres:
> I'm still seeing that answers in intel19 debug mode are not matching intel19 release and gcc7 debug/release. I think there is something wrong with the ice mask used in **`soca_static_socaerror_init`** ... another issue for another PR...
> ## Testing

That reminds me I was suppose to simplify the ice mask/bump: #263

guillaumevernieres (approved):
Oooo, 93.7% :+1:

PR #264, opened 2019-11-22
Seaice mask removal

### LINKED ISSUE #263: Remove the sea-ice mask in B
**Description**
The background based mask in the sea-ice B creates issues  when using an ensemble for the background error.

 **Solution**
Use the same mask as the ocean for sea-ice.

guillaumevernieres:
## Description
See issue #263: The sea-ice mask in B is removed, which changes all the tests making use of bump.
 
What bug does it fix, or what feature does it add? **Not a bug But also not a feature**
Is a change of answers expected from this PR? **Yes**

### Issue(s) addressed
- closes #263 
- May address an issue observed in #262 

## Testing
How were these changes tested? **ctest only**
What compilers / HPCs was it tested with?**gcc**

## Dependencies
None

### DIFF
--- a/src/soca/Covariance/soca_covariance_mod.F90
+++ b/src/soca/Covariance/soca_covariance_mod.F90
@@ -32,7 +32,6 @@ end type soca_pert
 type :: soca_cov
    type(bump_type), allocatable :: ocean_conv(:)  !< Ocean convolution op from bump
    type(bump_type), allocatable :: seaice_conv(:) !< Seaice convolution op from bump
-   integer,         allocatable :: seaice_mask(:,:)
    type(soca_field),    pointer :: bkg            !< Background field (or first guess)
    logical                      :: initialized = .false.
    type(soca_pert)              :: pert_scale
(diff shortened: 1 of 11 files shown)


aerorahul (commented):
@guillaumevernieres 
Can you explain why answers change?  Is it because a different mask is being applied?

travissluka:
@aerorahul yes. Due to the different mask, enspert will create different answers, which of course effects everything downstream

guillaumevernieres:
... And the sea-ice mask was also fed to bump. So the correlation operator is different.

travissluka (changes requested):
remove `i`,`j` at `soca_covariance_mod.F90` (line 66) as they are no longer used now.

I'm getting some nice divide by zeros in staticbinit with intel19 debug. I'm looking into it, might be a saber problem.

travissluka:
hmm, weird, intel19 debug throws a div by zero in the nicas normalization for the second horiz_convol that is created (in this case the ice, but if I switch the calls so that ice is created first, then the div by zero occurs in the ocean). Weird

guillaumevernieres:
image
Blame the rest of the world and move on?

travissluka:
@guillaumevernieres I found where the bug is occurring in saber, but when I try to debug with print statements the bug goes away :thinking: . Probably a compiler bug, move along. 

Fix the i,j variable issue and we can merge

travissluka (approved):

PR #267, opened 2019-12-04
randomize only the  fields listed in the yaml

travissluka:
## Description

All ocean/ice/sfc fields were being randomized when `Fields::random()` was called. This was causing problems when calculating norms with the upcoming addition of the `geometryIterator` test if not all fields were to be used. Now, only the fields used in the yaml file will be randomized, the rest will remain 0.0.

## Testing
* no change in the test answers (`enshofx` is currently failing due to unrelated reasons)
* @frolovsa the geometryIterator test should pass now when this is merged into your branch


travissluka:
note, the `test_soca_enshofx` fails on travisCI as expected. The code coverage is reduced because that test fails.

travissluka:
... and everything passes again. (Ignore that 0.1% decrease in code coverage!)

frolovsa (approved):
looks good to me

PR #281, opened 2020-02-15
Masked interpolation

travissluka:
## Description

horizontal interpolation was not taking the mask into consideration before, meaning a lot of zeros were getting interpolated in when in a gridbox near the coast.

This PR creates two sets of horizontal interpolation weights, one for masked grids, and one for unmasked. All fields except for `sea_area_fraction` are now interpolated using the masked grid.

The `hofx` and `3dvar` output looks happier now, minimum sss for hofx3d is now a more realistic 33psu (was 15psu), same thing for insitu S and SST values. 

I will do a quick test with RTS (real-time soca) before merging this to confirm all is well

### DIFF
--- a/src/soca/Fields/soca_interpfields_mod.F90
+++ b/src/soca/Fields/soca_interpfields_mod.F90
@@ -33,10 +33,11 @@ contains
 
 ! ------------------------------------------------------------------------------
 !> Compute interpolation weights
-subroutine initialize_interph(fld, locs, horiz_interp)
+subroutine initialize_interph(fld, locs, horiz_interp, horiz_interp_masked)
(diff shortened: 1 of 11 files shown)


guillaumevernieres:
@travissluka ... A drop in coverage?

travissluka:
@guillaumevernieres that's because none of the fields use the unmasked grid for backward interpolation, though I left the code for that in there. I thought about using the unmasked interpolation for the atmospheric variables, but wasn't sure if that would lead to inconsistencies with cool skin given the ocean variables used masked interpolation.... deal with that later once we start doing science.

guillaumevernieres (approved):
:+1:

PR #288, opened 2020-03-03
Make soca_fields more generic

travissluka:
## Description

Previously, all of the soca fields (`tocn`, `socn`, ...) were hardcoded resulting in a lot of duplication of code and made it a pain whenever a new field needed to be added. This tries to generalizes as much as possible the `soca_fields` code so that the lists of hardcoded field names can be removed in many places.

## details
- **900** fewer lines of code, wooo 🎉 
- `soca_field` was renamed to `soca_fields`. This type now holds an array of `soca_field` objects and the subroutines that operate on the fields
- `soca_field` added, this contains the data and metadata (such as name, domain, restart file name) associated with each field.  Currently the metadata for each field is defined in `soca_fields_init_vars()`, but this will be moved eventually to a yaml config file
- the ice and surface specific field types were removed
- to get field values now, use `has()` to see if a field exists, and `get()` to retrieve a pointer to the field. for example:

```
type(soca_fields) :: fields
type(soca_field), pointer :: cicen

...

if (fields%has("cicen")) then
  call fields%get("cicen", cicen)
  print *, sum(cicen%val(i,j,:))
end if
```

## future work
- I generalized the code in places that **wouldn't** result in significant changes to any yaml files or test answers. There are a couple more places where the code can be generalized more if we make changes to yaml config structure, this will be done in a subsequent PR.
- This PR keeps `Fields.cc`, future PR will move some functionality into `State` and `Increment`
- there are derived fields (`mld` and `layer_depth`) that are still hardcoded, need to be similarly generalized
- `has()` was not used much. It should be used more to make the code more robust. For example, the balance transform should still work if we omit ice concentration from our model state. (No point keeping ice around in the future Gulf of Mexico regional domain)

## Testing
- Tested with gcc and intel19.
- Answers to tests are all the same, except for 
  -  `3dvargodas` (an answer change from a previous PR somehow got missed)
  -  `enspert`, eventhough `hsnon` is not perturbed, it is needed by subsequent tests, and so it needed to be added to the `variables` section of the yaml file, which with this new code now makes it print test 'hsnon' output from `gpnorm`. Otherwise the `enspert` answers are identical


guillaumevernieres:
Nice coverage too! 5.79% to go ...

travissluka:
FYI, i'm seeing intermittent failures in `test_soca_dirac_socahyb_cov`, i'm trying to track down why

guillaumevernieres:
> FYI, i'm seeing intermittent failures in `test_soca_dirac_socahyb_cov`, i'm trying to track down why

I see a failure testing on singularity, but the issue seems to be in the compare script, "log=ref" on my side.
Thanks for the clean-up/refactoring, it definitely looks a LOT better now. One minor issue: Your new fancy editor is dumping a bunch of white spaces (witch makes my editor scream at me).

travissluka:
Yeah I've been meaning to fix that white space issue, I'll clean it up
before merge

guillaumevernieres (commented):
:+1: 
Looks good. Happy to merge that tomorrow unless you find a real issue with the dirac test but it looks fine on my side, besides the issue with compare.py. I'll test it on hera tomorrow before I approve ... just to make sure.

travissluka:
I think the `test_soca_dirac_socahyb_cov` issue I'm occasionally seeing is a  version of https://github.com/JCSDA/soca/issues/240. It seems to be caused by nonsense being placed in the halos of `cicen`, though I can't figure out why because all the fields are having their halos updated now. I'm going to write it off as "to be dealt with in a separate PR". Perhaps with fields cleaned up a valgrind sessions would be more productive now.

I am curious about your `comapre.py` issue though.

guillaumevernieres:
> I am curious about your `comapre.py` issue though.
Meh, old image, not worth investigating. All tests are OK on Hera/intel18.

guillaumevernieres (approved):
:+1: All good. Merging tonight.

PR #290, opened 2020-03-04
soca works without cicen

travissluka:
## Description

I know that ice is awesome, but soca-realtime is currently not using ice in its 3dvar. This PR restores the capability to run soca without ice.
- switched a line in `soca_fields_mod` so that it doesn't try to read an ice restart file if no filename is given
- added `%has('cicen')` statements to the balance operator so that it skips sections of the code if no `cicen` is given

## Testing

- A new test (`test_soca_varchange_balance_TSSSH`) was added that tests only the T/S/SSH parts of the balance. (similarly, when U/V balance is eventually added, a `_TSSSHUV` test should be added)
- no other tests are impacted, passes on gcc and intel19, confirmed to be working with soca-realtime


guillaumevernieres (approved):
:+1:

Issue #282, opened 2020-02-18
Geostrophic balance


guillaumevernieres:
Add geostrophic constraint in covariance model.

PR #291, opened 2020-03-04
u and v added along with corresponding geometry

### LINKED ISSUE #283: C-grid geometry
## Description
Add C-grid info to the current Geometry relevant to the staggered velocities.

### LINKED ISSUE #284: Add staggered u,v to State and increment
## Description
Add velocities to State/Incr and account for rotated u,v and grid staggering.

guillaumevernieres:
## Description
Addition of u,v in field and model and associated geometry (lon, lat and mask).
Slight change of answer in the forecast test when u and v are used and passed between soca and mom6 (type change is not reversible).

### Issue(s) addressed
This is 1 of many pr's that are needed to address epic #282.
closes #283
closes #284  

## Testing
Tested with singularity/gcc and hera/intel18
Changes are covered in the forecast_mom6 test

### DIFF
--- a/src/soca/Fields/soca_fields_mod.F90
+++ b/src/soca/Fields/soca_fields_mod.F90
@@ -977,7 +977,7 @@ subroutine soca_fields_read(fld, f_conf, vdate)
             ! TODO remove hardcoded variable names here
             ! TODO Add u and v. Remapping u and v will require interpolating h 
             case ('tocn','socn')
-              if (field%mask(i,j).eq.1) then
+              if (associated(field%mask) .and. field%mask(i,j).eq.1) then
                 varocn_ij = field%val(i,j,:)
                 call remapping_core_h(remapCS, nz, h_common_ij, varocn_ij,&
                       &nz, hocn_ij, varocn2_ij)
--- a/src/soca/Fields/soca_fields_mod.F90
+++ b/src/soca/Fields/soca_fields_mod.F90
@@ -130,6 +130,7 @@ subroutine soca_field_copy(self, rhs)
   self%cf_name = rhs%cf_name
   self%io_name = rhs%io_name
   self%io_file = rhs%io_file
+  self%mask => rhs%mask
 end subroutine soca_field_copy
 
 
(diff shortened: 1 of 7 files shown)


src/soca/Fields/soca_fields_mod.F90
```diff
@@ -45,6 +45,7 @@ module soca_fields_mod
   integer                           :: nz         !< the number of levels
   logical                           :: masked     !< if true, land mask should be applied
```
travissluka: delete `masked`, we don't need it now that we have `mask`... redundant information. 
Just check for `associated(mask)` anywhere `masked` is used.
Also, need to add `self%mask => rhs%mask` inside `soca_field_copy`

src/soca/Fields/soca_fields_mod.F90
```diff
@@ -45,6 +45,7 @@ module soca_fields_mod
...
   real(kind=kind_real), allocatable :: val(:,:,:) !< the actual data
+  real(kind=kind_real),     pointer :: mask(:,:)  !< field mask
```
travissluka:
```suggestion
  real(kind=kind_real),     pointer :: mask(:,:) => null() !< field mask
```

src/soca/Fields/soca_fields_mod.F90
```diff
@@ -954,12 +975,13 @@ subroutine soca_fields_read(fld, f_conf, vdate)
...
             select case(field%name)
             ! TODO remove hardcoded variable names here
+            ! TODO Add u and v. Remapping u and v will require interpolating h 
             case ('tocn','socn')
-              if (fld%geom%mask2d(i,j).eq.1) then
-                tsocn_ij = field%val(i,j,:)
-                call remapping_core_h(remapCS, nz, h_common_ij, tsocn_ij,&
-                      &nz, hocn_ij, tsocn2_ij)
-                field%val(i,j,:) = tsocn2_ij
+              if (field%mask(i,j).eq.1) then
```
travissluka:
```suggestion
              if (associated(field%mask) .and. field%mask(i,j).eq.1) then
```

travissluka (changes requested):

src/soca/Fields/soca_fields_mod.F90
guillaumevernieres: Good catch, thanks. Re-tested on hera with suggested changes.

travissluka (approved):

PR #294, opened 2020-03-05
combine run/compare steps in ctests

travissluka:
## Description

@ytremolet  wisely pointed out that we shouldn't have separate `run` and `compare` steps of the ctests. Combining these two steps into one lets us use `cmake --rerun-failed` now. 

A `test_wrapper.sh` script was added that is called by ctests and is responsible for running the run and compare steps together.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -283,6 +283,9 @@ function(soca_add_test)
     LIST(GET ARG_TOL 0 TOL_F)
     LIST(GET ARG_TOL 1 TOL_I)
 
+    # find the MPI command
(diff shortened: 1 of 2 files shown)


jjguerrette:
Very nice solution @travissluka!  Could test_wrapper.sh or something similar live in oops/tools alongside the compare scripts?

travissluka:
@jjguerrette certainly we can move it there eventually.... I just wasn't in the mood today to refactor the 2k lines of oops `test/CMakeLists.txt` files

jjguerrette:
An argument well made @travissluka.  Perhaps I'm being selfish, but we can use the script in mpas-jedi whether it gets used in OOPS ctests or not...

guillaumevernieres (approved):

PR #295, opened 2020-03-06
Spring cleaning of soca headers

travissluka:
## Description

Nothing fundamentally changed, just a tidying up...

- update copyright on all our files (so no more complaints for the rest of the year from me about "hey you forgot to update the year on the file you changed in your PR" )!
- wherever possible in the `.h` files, changed from using `#include` to using a forward declaration for outside classes, which is generally good practice, and should hopefully cut down on annoying chain reaction recompilations when something small changes in one of the `.h` files
   - some places, such as `Increment.h` had to have implementations moved from the `.h` to the `.cc` to make this possible
- Instead of throwing all the files to be compiled in the main `src/CMakeLists.txt`, each directory in the source now  has its own `CMakeLists.txt`. More in line with the cmake philosophy.
- Similarly, instead of most of the fortran interface functions being thrown into `Fortran.h`, each class keeps its interfaces in its subdirectory. (I'm still plotting ways to remove what is left in the `Fortran.h`)

### Issue(s) addressed
None, I was just bored.
A general cleanup before I start ripping apart the `Fields.cc` class.


guillaumevernieres (approved):
:1st_place_medal:

Issue #292, opened 2020-03-04
More variable transforms


guillaumevernieres:
- Implement interpolation from u or v grid to tracer grid. Note: No need to use a kdtree since neighbors are obtained from the logical grid (look at the possibility of using fms interp)   
- u,v rotation: Linear transformation of (u,v) between logical north and geographic north

PR #296, opened 2020-03-11
Feature/linear geostrophy

guillaumevernieres:
## Description
Addition of a linear geostrophy operator. Having "uocn" and "vocn" in the list of variables will trigger the calculation of the geostrophic current in "add_incr". Figure below is showing geostrophic surface current increment and ssh increment, looks right (1 deg resolution) ...
 
geostrophy-example

Missing feature that will be addressed in other PR's: 
- No beta plane proxi close to/at the equator
- Implement inside the "balance" variable change
- rotation to logical north
- Interpolation between u-grid, v-grid and tracer grid

### Issue(s) addressed
Very partially addresses #292 and #282   

## Testing
ctest, and 1 month cycling on Hera (what 0.5deg between friends!) adding the velocity increment to the state. Only a sanity check since the currents are not yet rotated nor interpolated on their respective grids

What compilers / HPCs was it tested with? gcc/intel18

Are the changes covered by ctests? Yes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# (C) Copyright 2017-2019 UCAR.
+# (C) Copyright 2017-2020 UCAR.
 #
 # This software is licensed under the terms of the Apache Licence Version 2.0
 # which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 JEDI encapsulation of MOM6  
 
-(C) Copyright 2017-2019 UCAR.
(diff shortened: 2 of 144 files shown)


travissluka:
cool, I did some single obs tests to convince myself things look physically correct 👍 
I'll take a closer look at the code this evening.

travissluka (approved):
My eyes glazed over when looking over the math.... we'll just assume it's all correct.

U increment from single ob test at 0 lat was a little ugly, I wonder how beneficial it would be to get the beta plane approx added next, I would imagine imbalances right on the EQ are not helping your over-excited Kelvin wave situation.

PR #301, opened 2020-03-17
Feature/rotate

guillaumevernieres:
## Description
- Addition of a variable change (Ana2Model) to rotate horizontal vectors between Model and State (for now only implemented for (u,v) ). The Model coordinate for velocities is locally defined by the grid. Using the "Ana2Model" variable change will rotate to a zonal/meridional coordinate system.  
- Removed the passing of fields from state to model at every time steps (not needed), this changes the forecast output slightly. 

### Issue(s) addressed
Partially addresses issue #292, #282.

## Testing

How were these changes tested? **Addition of a test for non-linear variable changes** 
What compilers / HPCs was it tested with? **gcc**


guillaumevernieres:
Oops ... No excuses, it even fails on my laptop.

guillaumevernieres:
First try!

src/soca/Fields/soca_fields_mod.F90
```diff
@@ -1471,6 +1472,54 @@ subroutine soca_fields_write_rst(fld, f_conf, vdate)
...
+subroutine soca_fields_rotate(self, coordinate)
+  class(soca_fields), intent(inout) :: self
+  character(len=*)                  :: coordinate ! "north" or "grid"
+  integer :: z, i
+
+  type(soca_field), pointer :: uocn, vocn
+  real(kind=kind_real), allocatable :: un(:,:,:), vn(:,:,:)
+
+  ! get (uocn, vocn) and make a copy
+  if (self%has("uocn").and.self%has("vocn")) then
```
travissluka: I don't like the hardcoding of "uocn" and "vocn". The pairs of variable names should be configurable in the yaml somwehere, or at the very least at a higher level in the code and passed into this function. 
Granted, we only care about ocean u/v for now, but what about in the future when we have  ice u/v, wave u/v, whale/sea-lion trajectory u/v.... etc

src/soca/Fields/soca_fields_mod.F90
```diff
@@ -1471,6 +1472,54 @@ subroutine soca_fields_write_rst(fld, f_conf, vdate)
 
 end subroutine soca_fields_write_rst
 
+! ------------------------------------------------------------------------------
+!> Rotate horizontal vector
+subroutine soca_fields_rotate(self, coordinate)
+  class(soca_fields), intent(inout) :: self
+  character(len=*)                  :: coordinate ! "north" or "grid"
+  integer :: z, i
```
travissluka: `i` is unused, delete

travissluka (changes requested):

src/soca/Fields/soca_fields_mod.F90
guillaumevernieres: @travissluka agreed, but we would need to re-write the "read" routine. For example, we try to get "hocn" at the start of read, without checking if it is part of the list of variables. Not a huge tasks, but I'd rather confine it to its separate issue.

guillaumevernieres:
@travissluka , as requested, I added the possibility of rotating a pair of fields. A scientifically sound example is in 3dvar_godas and a test with 3 pairs is in the ana2model ctest were we rotate things like (T,S) ... because we can.

travissluka:
@guillaumevernieres did you happen to do any plotting to visually confirm it's working as intended?

guillaumevernieres:
@travissluka I dumped the output on the screen as I was putting that thing together, but good point, that's not enough. Let me check/plot.

travissluka:
ack, we're running with MOM6 settings that keep a rotation bug at the tripolar fold. Look at sin(angle) at the top right
image

travissluka:
@guillaumevernieres I'll push a PR first that fixes the above sin/cos from MOM6, since that'll change existing answers... should just take me a minute

guillaumevernieres:
> @guillaumevernieres I'll push a PR first that fixes the above sin/cos from MOM6, since that'll change existing answers... should just take me a minute

Will do. Good catch, again! I was looking at the 1 deg grid (see below), which as the same issue but quite harder to see.
sinbug

travissluka:
@guillaumevernieres  actually... i'm not doing a separate PR, that bug seems to not affect MOM6 answers (if I remember correctly it does affect SIS2 answers though, so not a concern for us). But to get you the right sin, add these lines to the end of  `test/Data/72x35x25/MOM_input` and rerun the gridgen app

```
USE_TRIPOLAR_GEOLONB_BUG = False
GRID_ROTATION_ANGLE_BUGS = False
```

guillaumevernieres:
woot ...
goodsin

travissluka (approved):
If you want to go ahead and merge, so that we can keep up with the Fields changes i'm doing.
We'll make sure it's scientifically correct later.

guillaumevernieres:
> If you want to go ahead and merge, so that we can keep up with the Fields changes i'm doing.
> We'll make sure it's scientifically correct later.

I'm pretty sure it's fine, but I'll try to think of a test in a different pr.

PR #302, opened 2020-03-17
fix an "array temporary" intel warning

travissluka:
## Description

Fixes an "array temporary" runtime warning that pops up thousands of time when compiled with intel19/debug. Likely was causing some small slowdown.

similar to JCSDA/oops/pull/551, JCSDA/oops/issues/550

## Testing

GCC is still the same, intel19/debug runtime warnings go away


guillaumevernieres (approved):
:+1:

PR #303, opened 2020-03-18
remove soca Fortran registries

travissluka:
## Description
Continuing the SOCA spring cleaning...

I don't like the linked list registry. It bothers me. Most recently because I wanted to use a linked list in a fortran module for another purpose, but you can't have more than one linked list type defined in a single file. This PR plays around with removing all SOCA registries and replacing them with directly passed pointers.

We might go through with this on the soca side regardless for the sake of cleanliness, but @ytremolet I wanted to run it by you anyway. @shlyaeva says the reason the linked lists were used in the first place was because of cray compiler issues, do we know if this is still true?

- pointers are passed directly and translated using `c_f_pointer()` and `c_loc()`
- instead of an integer key passed to fortran, a dummy `struct Ftn{};` is nested in each relevant C++ class, and a `Ftn * ftn_` member variable is kept. Passing pointers of type `<c_class>::Ftn *` ensures C++ type checking works. (`typedef int` are all the same as far as the compiler is concerned, and there were instances where the soca fortran/C bindings were defined incorrectly  though luckily the right things were being passed in)
- gets rid of the global `Fortan.h` that I also hate for some reason

## Testing

GCC and intel19 are both happy


guillaumevernieres:
I like it.

ytremolet:
Nobody wants these linked lists! The solution you propose here is what was originally in the code.

As described above, it did not work with the Cray compilers. It would have to be tested with more recent versions. I think it is important to keep these compilers because many people who have tried seriously find that codes runs faster than with intel. 

There was another compiler (I forgot which) for which the code would work but it created a memory leak in the destructor because the fortran pointer was never deallocated. It was not a serious issue in terms of memory because it was only the pointer, but it prevented the use of tools like valgrind (everything was swamped in errors related to those fortran pointers). So also something to check before making those changes.

travissluka:
abandoning

Setting my dreams higher of someday switching fields to use atlas and ripping out 95% of the Fortran code.

PR #304, opened 2020-03-19
fix issue with ufo 

travissluka:
## Description

A work around to https://github.com/JCSDA/ufo/pull/835 breaking TravisCI builds for SOCA

`LOCAL_PATH_TESTFILES_IODA` is set to a nonsense value so that the sections of the UFO `CMakeLists.txt` that are having problems are skipped.

## Testing

TravisCI tests now work.


guillaumevernieres (approved):
Amazing!

PR #305, opened 2020-03-20
Remove Fields C++ class

travissluka:
## Description

This removes the `Fields.cc` and `Fields.h` files, on my continued march to separate state/increment. The `Increment` and `State` C++ classes now directly call the relevant Fortran.

I haven't decided yet if I'll leave the fortran Fields the way it is, or create separate Fortran state/increment derived classes, to be determined later

359 fewer lines of code now :tada: 

## Testing

Everything is the same, no answers changed


src/soca/Model/Model.cc
```diff
@@ -53,27 +52,23 @@ namespace soca {
   }
   // -----------------------------------------------------------------------------
   void Model::initialize(State & xx) const {
-    ASSERT(xx.fields().isForModel(true));
-    soca_initialize_integration_f90(keyConfig_, xx.fields().toFortran());
-    Log::debug() << "Model::initialize" << xx.fields() << std::endl;
+    soca_initialize_integration_f90(keyConfig_, xx.toFortran());
```
guillaumevernieres, line 55: Yeah, that error trap wasn't doing much of anything.

guillaumevernieres (approved):
Thanks for that work. Not much we can do about the coverage for this pr, so all good.

PR #306, opened 2020-03-20
Remove Fields interface

travissluka:
## Description

The C++/Fortran interface for `Fields` is removed. From the C++ side `State` and `Increment` are now completely separated and all references to `Fields` are gone. 

- removed `FieldsFortran.h`, `soca_fields.interface.F90`
- added separate C++/Fortran interfaces for state/increment. `IncrementFortran.h` and `StateFortran.h` only contain the respective relevant functions that were in `FieldsFortran.h`
- On the Fortran side, `soca_state` and `soca_increment` were added, and for now are just a meaningless extension of `soca_fields` (this will change in next PR). There are some places where `soca_fields` is still used internally (such as bkg standard deviation, which is neither a state nor increment)
- `soca_field_registry` is gone, replaced with `soca_increment_registry` and `soca_state_registry`
- I had to put the state/increment registries in separate files due to a cyclic dependency

The next step is a separation of non-common code from `soca_fields`. `soca_fields` will retain all common subroutines and data structures, but everything else will be moved to `soca_state` and `soca_increment`

## Testing

tested on GCC and intel19/release


guillaumevernieres (approved):
:+1:

PR #307, opened 2020-03-23
separate soca_fields into soca_state and soca_increment

travissluka:
## Description

A follow up to https://github.com/JCSDA/soca/pull/306
Separating `Fields` on the fortran side. Hopefully the last giant PR in this series (more small ones to come!)

- subroutines that were NOT common to both `soca_state` and `soca_increment` were moved out of `soca_fields` and into the correct sub type
- The hardcoded `mld` and `layer_depth` arrays were removed. Instead,  `soca_state%create()` augments the list of variables before passing it to `soca_fields%create()`. This means state and increment now have different fields
- To support state and increment having different fields: 
  - added the `soca_fields%check_subset()` checker. It makes sure the given set of fields are a proper subset of the other set of fields. In contrast to `soca_fields%check_congruent()` which makes sure they are the same.
  - some of the `multiply()` subroutines of the transforms have been modified to handle input/output/background having different fields. (i.e. explicitly calling `%get()` on each set of fields because now the indexes dont match up)

## Next steps
 - change the yaml files so that the transforms dont get passed the entire state
 - remove some more non-generic spots in the code that will likely create small differences in the answers (e.g. printing the state min/max/mean)
 - fields I/O is a mess, need to split that up.

## Testing

No change in answers


travissluka:
ugh, code coverage decrease of 0.01%, i'm horrible

guillaumevernieres:
> ugh, code coverage decrease of 0.01%, i'm horrible

big time :-1: 
I'm worried you won't be able to sleep over this, perhaps you could remove the few unused routines in utils?

guillaumevernieres:
@travissluka , it looks like the "iread==2" isn't tested in [soca_fields_mod](https://codecov.io/gh/JCSDA/soca/pull/307/src/src/soca/Fields/soca_fields_mod.F90#L804), and probably never was. It should be, but not the work of this pr.
I'm pretty sure [h2z](https://codecov.io/gh/JCSDA/soca/pull/307/src/src/soca/Utils/soca_utils.F90#L102) could be yanked out.

guillaumevernieres:
@travissluka , I was also going to suggest to trigger the writing of the local grid in the "gridgen" test to cover write2pe (only used for debugging), but it looks like we now have a bug in write2pe ... We probably need a specific issue for improved coverage. I'm fine with how things are, thanks for your work.

guillaumevernieres (approved):
:+1:

PR #316, opened 2020-03-30
Feature/colocate

### LINKED ISSUE #315: Add colocation method in Field
## Description
The figure below depicts what grid we decided to use for the Model, State and Increment.
SOCA-Horizontal-Grids

Adding the increment to the state requires colocating the increment to the state's grid.

## Solution
Add a colocation method to the field object.

## Related to
Partially addresses #282

guillaumevernieres:
## Description
Colocate increment to state in `add_incr`. Interpolation uses fms idw ... hu-hoooo .... For now.
**bug fix:** Geostrophic increment for u-v was added without colocation.
Change of answer expected in the 3dvar_godas test, which makes use of geostrophy.

### Issue(s) addressed
closes #315

## Testing
How were these changes tested? **Colocation called inside of 3dvar_godas**
What compilers / HPCs was it tested with? **gcc**

## Dependencies
- ~[oops PR#585](https://github.com/JCSDA/oops/pull/585)~
Deferred to an other PR, we'll just have to remove the NOCOMPARE in the test and add the reference file.

### DIFF
--- a/src/soca/Fields/soca_fields_mod.F90
+++ b/src/soca/Fields/soca_fields_mod.F90
@@ -26,6 +26,10 @@ use soca_geom_mod, only : soca_geom
 use soca_fieldsutils_mod, only: soca_genfilename, fldinfo
 use soca_utils, only: soca_mld
 
+use horiz_interp_mod, only : horiz_interp_type
+use horiz_interp_spherical_mod, only : horiz_interp_spherical_new, horiz_interp_spherical
+use tools_const, only: deg2rad
+
 implicit none
 
 private
@@ -40,6 +44,9 @@ type :: soca_field
   integer                           :: nz         !< the number of levels
   real(kind=kind_real), allocatable :: val(:,:,:) !< the actual data
   real(kind=kind_real),     pointer :: mask(:,:) => null() !< field mask
+  real(kind=kind_real),     pointer :: lon(:,:) => null()  !< field lon
+  real(kind=kind_real),     pointer :: lat(:,:) => null()  !< field lat
+  character(len=1)                  :: c_grid_loc !< "h", "u" or "v"
   character(len=:),     allocatable :: cf_name    !< the (optional) name needed by UFO
   character(len=:),     allocatable :: io_name    !< the (optional) name use in the restart IO
   character(len=:),     allocatable :: io_file    !< the (optional) restart file domain
@@ -50,6 +57,8 @@ contains
 
   procedure :: check_congruent => soca_field_check_congruent
   procedure :: update_halo     => soca_field_update_halo
+  procedure :: stencil_interp  => soca_field_stencil_interp
+
 end type soca_field
 
 ! ------------------------------------------------------------------------------
@@ -90,6 +99,7 @@ contains
 
   ! misc
   procedure :: update_halos => soca_fields_update_halos
+  procedure :: colocate  => soca_fields_colocate
 
 end type soca_fields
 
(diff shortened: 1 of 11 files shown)


guillaumevernieres:
I'll blame the multiple test failure on the coffee rationing. I know the reviewers are eager to start plunging into the added code (it's quite amazing if I may say), so I'll be pushing the fix in a few minutes.

src/soca/Fields/soca_fields_mod.F90
```diff
@@ -1060,4 +1099,73 @@ subroutine soca_fields_write_rst(fld, f_conf, vdate)
...
+
+  ! Apply interpolation to all fields, when necessary
+  do i=1,size(self%fields)
+
+    ! Check if already colocated
+    if (self%fields(i)%c_grid_loc == cgridlocout) cycle
+
+    ! Initialize fms spherical idw interpolation
+     g => self%geom
+     call horiz_interp_spherical_new(interp2d, &
```
travissluka, line 1140: memory leak, you need a matching call to `horiz_interp_spherical_del` at least once at the end
(If you look inside `horiz_interp_type` all the data members are pointers and so wont automatically be deallocated upon `interp2d` deallocation)

travissluka (changes requested):

src/soca/Fields/soca_fields_mod.F90
guillaumevernieres: Arf, yes, good catch, forgot the cleaning. I'll spend "some" time on the coverage too.

travissluka: :+1: is there another way we we could test it? I'm feel uneasy using `3dvar_godas` as the dumping ground for testing new features

guillaumevernieres: > +1 is there another way we we could test it? I'm feel uneasy using `3dvar_godas` as the dumping ground for testing new features

Ideally it should be tested in the oops State test through the testing of `add_incr`, but `add_incr` is not tested. Maybe I'll add one tomorrow while I procrastinate over my "smart" goals. This doesn't resolve the issue of poor coverage (the test only exercising the colocation to the h-grid). I'll think about it tomorrow, but maybe I need to remove the u-grid and v-grid options.

guillaumevernieres:
The add incr test will fail until [oops PR#585](https://github.com/JCSDA/oops/pull/585) is merged, or someone that's not @guillaumevernieres tells travis-ci to build with the bugfix branch in oops ...

guillaumevernieres:
- Commented out the u and v colocation option until we figure out a way to test these options
- Removed the dependency to [oops pr585](https://github.com/JCSDA/oops/pull/585) by removing the compare test in the `addincrement` test (and realizing as I am typing that I forgot to remove the testref file ...)
- @travissluka I'm OK using 3dvar_godas to test new features, why the uneasiness?

travissluka:
@guillaumevernieres It's called a "unit test" for a reason. Throwing everything in 3dvar_godas no longer makes it a "unit". If the answers change, we have a harder time tracking down why given that there are multiple features that aren't tested elsewhere. But whatever, deal with that later

travissluka:
... you never committed your addincrement source files.

guillaumevernieres:
@travissluka 
- Unit test for rotate is in varchange and convert state
- Unit test for colocate is in addincrement
3dvar_godas makes use of these new features, the same way is uses a bunch of other stuff. But you are correct, it was not tested before my unfortunate latest push that was suppose to have the `addincrement` test ... My bad, will push in a few minutes.

travissluka:
:+1: yeah my comment about testing was before `addincrement`

travissluka (approved):

PR #322, opened 2020-04-06
Improved HDIAG

benjaminmenetrier:
This PR is an update of references for tests using BUMP, linked to SABER PR https://github.com/JCSDA/saber/pull/66.

One source file using the Gaspari and Cohn function from SABER is also updated (maybe we should use [the one implemented in OOPS](https://github.com/JCSDA/oops/blob/develop/src/oops/generic/gc99.cc)?).


travissluka (approved):

guillaumevernieres (approved):
Merci @benjaminmenetrier

PR #326, opened 2020-04-14
GetValues and LinearGetValues classes for soca

### LINKED ISSUE #312: Implement the new GetValue class
## Description
Adapt getvalue to [oops PR #514](https://github.com/JCSDA/oops/pull/514).

guillaumevernieres:
## Description

Implements changes required by  [oops #514](https://github.com/JCSDA/oops/pull/514). 
Copy/paste/delete (lots of delete!) of @danholdaway 's [fv3-jedi #314](https://github.com/JCSDA/fv3-jedi/pull/314).

Is a change of answers expected from this PR?
NO. Zero diff with the current getvalue implementation in develop. 

### Issue(s) addressed
- closes #312

### Possible issue remaining
Addition of lint directive to pass the code-norm test. cpplint.py complains about non const references [here](https://github.com/JCSDA/soca/blob/aa5232193d1836e5a44d84d85295be4d5bbde85c/src/soca/GetValues/LinearGetValues.h#L60), [here](https://github.com/JCSDA/soca/blob/aa5232193d1836e5a44d84d85295be4d5bbde85c/src/soca/GetValues/LinearGetValues.h#L66) and [here](https://github.com/JCSDA/soca/blob/aa5232193d1836e5a44d84d85295be4d5bbde85c/src/soca/GetValues/LinearGetValues.h#L67). Inherited, from oops, so I don't have the choice of converting to const or pointers. 
Results are zero diff with develop, so ignoring sounds good to me. Could it be hiding an underlying problem?

## Testing

Addition of 2 ctest, one for each class.

Are the changes covered by ctests? (If not, tests should be added first.)

## Dependencies

- waiting on [oops #514](https://github.com/JCSDA/oops/pull/514)
- waiting on [ufo #822](https://github.com/JCSDA/ufo/pull/822)


src/soca/GetValues/GetValues.cc
```diff
@@ -0,0 +1,89 @@
...
+void GetValues::fillGeoVaLs(const State & state,
+                            const util::DateTime & t1,
+                            const util::DateTime & t2,
+                            ufo::GeoVaLs & geovals) const {
+  const util::DateTime * t1p = &t1;
+  const util::DateTime * t2p = &t2;
+  // Get atm geovals
+  if (geom_->getAtmInit())
+  {
+    // Get atm geovals
```
shlyaeva, line 47: An uninvited suggestion: I think the read may be moved to constructor, so it's only called once per H(x) computation; and then in fillGeoVaLs you could just fill in relevant variables for relevant locations from the atmospheric geovals that were read in the constructor.

guillaumevernieres: Sounds good. Thanks for looking at it @shlyaeva , I'll address the issue in a different pr, maybe at the same time we add 4d to the atm geovals.

src/soca/GetValues/soca_getvalues_mod.F90
```diff
@@ -195,7 +156,7 @@ subroutine interp(fld, locs, vars, geoval, horiz_interp, horiz_interp_masked)
       call fld%get("tocn", fldptr)
       fld3d(isc:iec,jsc:jec,1) = fldptr%val(isc:iec,jsc:jec,1)
 
-    ! TODO: Move unit change elsewhere, check if is COARDS.
+      ! TODO: Move unit change elsewhere, check if is COARDS.
```
danholdaway, line 159: This could be done in the variable change. We do something similar for Ozone in a variable change.

danholdaway (approved):

src/soca/GetValues/soca_getvalues_mod.F90
guillaumevernieres: Yup, it's in the `why do today what can be done tomorrow epic.`

src/soca/GetValues/GetValues.cc
```diff
@@ -0,0 +1,89 @@
+/*
+ * (C) Copyright 2019-2020 UCAR
+ *
+ * This software is licensed under the terms of the Apache Licence Version 2.0
+ * which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
```
travissluka, line 7:
```suggestion
#include "eckit/config/LocalConfiguration.h"

```

src/soca/GetValues/GetValues.h
```diff
@@ -0,0 +1,80 @@
...
+#include "oops/util/ObjectCounter.h"
+#include "oops/util/Printable.h"
+#include "oops/util/Timer.h"
+
+#include "ufo/Locations.h"
+
+// Forward declarations
+namespace eckit {
+  class Configuration;
+}
```
travissluka:
```suggestion
#include "soca/Fortran.h"

#include "oops/util/ObjectCounter.h"
#include "oops/util/Printable.h"

#include "ufo/Locations.h"

// Forward declarations
```

src/soca/GetValues/LinearGetValues.cc
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+#include "oops/util/Logger.h"
+
+#include "soca/Geometry/Geometry.h"
+#include "soca/GetValues/LinearGetValues.h"
+#include "soca/GetValues/GetValuesFortran.h"
+
+#include "ufo/GeoVaLs.h"
+#include "ufo/Locations.h"
```
travissluka:
```suggestion
#include "oops/util/DateTime.h"
#include "oops/util/Logger.h"

#include "soca/Geometry/Geometry.h"
#include "soca/GetValues/GetValuesFortran.h"
#include "soca/GetValues/LinearGetValues.h"
#include "soca/Increment/Increment.h"
#include "soca/State/State.h"

#include "ufo/GeoVaLs.h"
#include "ufo/Locations.h"
```

src/soca/GetValues/LinearGetValues.h
```diff
@@ -0,0 +1,82 @@
...
+#include "ufo/Locations.h"
+
+// Forward declarations
+namespace eckit {
+  class Configuration;
+}
+namespace ufo {
+  class GeoVaLs;
+  class Locations;
+}
```
travissluka:
```suggestion
#include <ostream>
#include <string>
#include <memory>

#include "soca/Fortran.h"

#include "oops/util/ObjectCounter.h"
#include "oops/util/Printable.h"

#include "ufo/Locations.h"

// Forward declarations
namespace ufo {
  class GeoVaLs;
  class Locations;
}
namespace soca {
  class Increment;
  class State;
}
```

travissluka (changes requested):
see my nit picking comments . Try to keep only the bare minimum of `#include`s needed in header files, they should be in the `.cc` instead when possible (basically only things that are related to class member variables stay in the `.h`, not things in the class method prototypes). I removed `#include`s that weren't needed, and moved a couple things to forward declarations instead.

travissluka:
also, why the `feature/getvalues-gv` branch and not `feature/getvalues`?? This could be dangerous if something changes in the oops/ufo `getvalues` branch and we think everything is still working when in reality it isn't

also, also... thanks :+1:

guillaumevernieres:
> see my nit picking comments . Try to keep only the bare minimum of `#include`s needed in header files, they should be in the `.cc` instead when possible (basically only things that are related to class member variables stay in the `.h`, not things in the class method prototypes). I removed `#include`s that weren't needed, and moved a couple things to forward declarations instead.

@travissluka : I did try to move as many as I could to the .cc before submitting this pr. Did you build/test your suggestions? I'll test locally.

travissluka:
Yes, I did. Would you like me to just push a commit with those?

guillaumevernieres:
> also, why the `feature/getvalues-gv` branch and not `feature/getvalues`?? This could be dangerous if something changes in the oops/ufo `getvalues` branch and we think everything is still working when in reality it isn't
> 
> also, also... thanks +1

Poor judgement from my side. I forgot exactly what was the issue, I'll keep track of the changes in ufo and oops (yeah right!).

guillaumevernieres:
> Yes, I did. Would you like me to just push a commit with those?

:+1:  Yes, please do!

travissluka (approved):

Issue #340, opened 2020-05-01
Vertical change of resolution


guillaumevernieres:
## Description
This is a follow up, or in complement to #339 and will allow to interpolate between geometries with different vertical resolutions.

## Solution
Use MOM6's internal vertical interpolations. 

## Dependencies
- #339 needs to be addressed first

PR #341, opened 2020-05-04
Allow multiple communicators

guillaumevernieres:
## Description
Finishing the work started in #208 : Pass the mpi communicator to the lower fortran levels.

**What bug does it fix, or what feature does it add?**
This should allow the use of multiple communicators, required to implement some of the oops applications. This PR should allow the implementation of the change of resolution (#206, #339 ), and EDA, possibly, unless I forgot something ...

Is a change of answers expected from this PR? **No**

### Issue(s) addressed
- closes #210 

## Testing
How were these changes tested? **Using the current ctests**
What compilers / HPCs was it tested with? **gcc 7.4, Intel 18**
Are the changes covered by ctests? **yes**

## Dependencies
None

Issue #339, opened 2020-05-01
Horizontal change of resolution


guillaumevernieres:
## Description
Currently, "change_resolution" in State and Increments are not implemented. As a first step, implement methods to interpolate a State or an Increment between 2 geometries which have the same vertical coordinate. 

## Solution
- Simple/naive 2D interpolation (by @guillaumevernieres or @travissluka ) that will be deemed sufficient for our DA needs
- More elaborate case that will enable the interpolation of the entire model state (This work will be done by @clouden90 and probably requires its own issue)

I both cases, we will investigate the use of the available fms interpolation package and the unstructured one in oops.

## Dependencies
- This work cannot start before #210 is resolved

guillaumevernieres:
@clouden90 , assign yourself or @flampouris , please assign @clouden90 . We'll talk today (05/07/2020) about @clouden90 's level of involvement for this step.

clouden90:
> @clouden90 , assign yourself or @flampouris , please assign @clouden90 . We'll talk today (05/07/2020) about @clouden90 's level of involvement for this step.

@guillaumevernieres , sounds perfect. Let's talk about the plans, step by step for #339 today.

Issue #11, opened 2018-03-27
Multiple observation types


guillaumevernieres:
Currently, we can only assimilate one obs type in the DA window. Look at qg in oops or in the old mom5cic5 interface.

guillaumevernieres:
Fixed and pushed into develop. "We" can now assimilate multiple obs types within the same DA window.

PR #344, opened 2020-05-05
Feature/convert stat

### LINKED ISSUE #309: Interpolate MOM6 restarts between domains
Given 2 geometries, interpolate all the necessary MOM6 fields included in a restart file.

clouden90:
## Description
SOCA State Converter

What bug does it fix, or what feature does it add?
This application performs state conversion both horizontally and vertically. A few changes has been made based on the previous comments
.
The major change is now the intermediate grid step is removed. The [mom_diag_remap module](https://noaa-gfdl.github.io/MOM6/APIs/namespacemom__diag__remap.html#a3cfd706c08bbd33ad21b77c6b045a9c5) is used and the re-gridding is done in soca_gridgen step.

Now the application will do conversion based on the following steps:
(0) Pre-processing: run soca_gridgen to convert grid coordinate to z* grid for both source & target grid;
(1) Call soca_convertstate_setup to read pre-computed z* grid and original grid from soca_gridspec  nc file for both source & target grid;  
(2) Call soca_convertstate_change_resol to convert selected state variables (except target grid cell thickness, which has been defined in step (1)) by following steps:
a.	[Vertical remapping](https://mom6.readthedocs.io/en/dev-gfdl/api/generated/modules/mom_remapping.html#f/mom_remapping/remapping_core_h) from source grid to source_z* grid for selected variable;
b.	[Horizontal remapping](https://noaa-gfdl.github.io/MOM6/APIs/namespacemom__tracer__initialization__from__z.html#a5f3dc84eae5f78304c8e366bf80e3afd) from source_z* grid to target_z* grid for selected variable;
c.	Vertical remapping from target_z* grid to target grid for selected variable.

Several Issues:
1.	The horiz_interp_and_extrap_tracer() is the subroutine used for horizontal interpolation in MOM6 . The main problem we cannot directly use horiz_interp_and_extrap_tracer() is that it only supports tracer variables (e.g. T, S), but we need conversion of vector variables in our application.  Also I found that fill_miss_2d may not work properly in the case that grid cell is surrounded by land mask. 

2.	The example of how to use the state converter can be found in test case “test_soca_convertstate” and associated yaml file “convertstate.yml”. This case shows conversion from global domain (5deg) to coarser global grid (generated from test_soca_gridgen_small). I have done several tests to validate the accuracy/performance of the SOCA state converter and the results are well-documented in a separated report.

## Issue(s) addressed
closes #309 

## Testing
How were these changes tested? Updated test_soca_convertstate (Global -> Coarse global grid)
What compilers / HPCs was it tested with? gcc 7.4, Intel 18, Intel 19
Are the changes covered by ctests? Yes

## Dependencies
#341 , #339 and possibly #340, or #340 could be addressed here.
meshgrid and soca_fill_miss_2d will be removed after PR: [Change meshgrid and fill_miss_2d to public #11](https://github.com/JCSDA/MOM6/pull/13)
#348 (future work)


guillaumevernieres:
@clouden90 , we need a few pr's to go in before we can think about merging that work. @travissluka and I will address #339 (#210 is in review) since this is going to break a few of our change of variables ... I think. Your work should be a little easier to fit in after #210 and #339 to then address and hopefully close #340 and #309, possibly enhancing whatever we'll come up with in #339.
Sounds like a plan?

clouden90:
> @clouden90 , we need a few pr's to go in before we can think about merging that work. @travissluka and I will address #339 (#210 is in review) since this is going to break a few of our change of variables ... I think. Your work should be a little easier to fit in after #210 and #339 to then address and hopefully close #340 and #309, possibly enhancing whatever we'll come up with in #339.
> Sounds like a plan?

@guillaumevernieres , sure, it sounds a perfect plan to me. I do see #210 before reopening this draft pull request. This draft PR is mainly for tracking purpose as well as exchanging ideas. I will also take a look #210 and other dependencies to see how I am going to fit in my work.

clouden90:
@travissluka , I accidentally removed you from the reviewer's list. Just added you back to keep you in the loop. No need actions now.

src/soca/State/soca_state_mod.F90
```diff
@@ -8,6 +8,7 @@ module soca_state_mod
 use soca_geom_mod
 use soca_fields_mod
 use soca_increment_mod
+use soca_convert_stats_mod
```
guillaumevernieres: Typo? convert_state instead of convert_stats

src/soca/Utils/soca_convert_stats_mod.F90
```diff
@@ -0,0 +1,476 @@
+! (C) Copyright 2017-2020 UCAR.
```
guillaumevernieres: 2020-2020 instead of 2017-2020

src/soca/Utils/soca_convert_stats_mod.F90
```diff
@@ -0,0 +1,476 @@
...
+!  
+!  call rhs%get("hocn", hocn1) ; call self%get("hocn", hocn2)
+!  call convert_stat%setup(rhs%geom, self%geom, hocn1, hocn2)
+!  do n = 1, size(rhs%fields)
+!    field1 => rhs%fields(n) ; call self%get(trim(field1%name),field2)
+!    if (field1%io_file=="ocn") call convert_stat%change_resol(field1, field2, rhs%geom, self%geom)
+!  end do !n
+!  call convert_stat%clean()
+!end subroutine soca_state_convert
+! ------------------------------------------------------------------------------
```
guillaumevernieres: I don't think we need the example.

src/soca/Utils/soca_convert_stats_mod.F90
```diff
@@ -0,0 +1,476 @@
...
+
+module soca_convert_stats_mod
+
```
guillaumevernieres: state instead of stats? Or maybe you had something else in mind? I'm open to suggestion, but your application is converting a state.

src/soca/Utils/soca_convert_stats_mod.F90
```diff
@@ -0,0 +1,476 @@
...
+  do k = 2,kd ; self%z_bot(k)= self%z_bot(k-1)+self%h_common(k) ; end do  
+ 
+  !
+  allocate(self%hocn_src(src%isd:src%ied,src%jsd:src%jed,1:src%nzo))
+  allocate(self%hocn_des(des%isd:des%ied,des%jsd:des%jed,1:des%nzo))
+
+  ! set hocn for src & des grid
+  self%hocn_src = hocn%val
+  hocn2%val = des%h
+  self%hocn_des = hocn2%val
```
guillaumevernieres: Why do you need hocn2?

src/soca/Utils/soca_convert_stats_mod.F90
```diff
@@ -0,0 +1,476 @@
...
+  !local
+  integer :: i, j, k, kd
+  real(kind=kind_real) :: max_depth
+  !
+  real(kind=kind_real), dimension(40) :: woa_dz   = (/ 5.,  10.,  10.,  15.,  22.5, 25., 25.,  25.,  &
+                                                      37.5, 50.,  50.,  75., 100., 100., 100., 100., &
+                                                     100., 100., 100., 100., 100., 100., 100., 175., &
+                                                     250., 375., 500., 500., 500., 500., 500., 500., &
+                                                     500., 500., 500., 500., 500., 500., 500., 500. /)
+  !
```
guillaumevernieres: I do cringe here, but I see you point ... But if we do it this way, I do think the intermediate grid resolution should be flexible. Did you contact the GFDL developers by any chance? I feel that this problem has been resolved already.

src/soca/Utils/soca_convert_stats_mod.F90
```diff
@@ -0,0 +1,476 @@
...
+                                                      37.5, 50.,  50.,  75., 100., 100., 100., 100., &
+                                                     100., 100., 100., 100., 100., 100., 100., 175., &
+                                                     250., 375., 500., 500., 500., 500., 500., 500., &
+                                                     500., 500., 500., 500., 500., 500., 500., 500. /)
+  !  
+  kd = size(woa_dz) ; self%kd = kd
+  allocate(self%z_bot(1:kd)) ; allocate(self%h_common(1:kd)) 
+  self%h_common(:) = woa_dz(:) ; self%z_bot(1) = self%h_common(1) 
+  do k = 2,kd ; self%z_bot(k)= self%z_bot(k-1)+self%h_common(k) ; end do  
+
```
guillaumevernieres: I know we don't have a fortran code norm, but I'm not a big fan of the ; separated 1 liner.

src/soca/Utils/soca_convert_stats_mod.F90
```diff
@@ -0,0 +1,476 @@
...
+      end do ! k
+    end do ; end do !i j
+  else
+    mask_ = 1.d0 ; mask_des = 1.d0
+  end if 
+ 
+  ! set lon lat for src & des
+  if (field_src%name == "hocn" .and. field_des%name == "hocn" ) then
+    return
+  end if
```
guillaumevernieres: Sorry, I lost the logic. Why do we return at that point?

src/soca/Utils/soca_convert_stats_mod.F90
```diff
@@ -0,0 +1,476 @@
...
+   if (field_des%io_file=="ocn") field_des%val(:,:,1) = tmp2(:,:,1)*field_des%mask(:,:) ! 2D
+   if (field_des%io_file=="sfc") field_des%val(:,:,1) = tmp2(:,:,1) ! 2D no mask
+   if (field_des%io_file=="ice") field_des%val(:,:,1:field_des%nz) = tmp2(:,:,1:field_des%nz)
+  end if ! nz > 1
+  call mpp_update_domains(field_des%val, geom_des%Domain%mpp_domain)
+
+end subroutine soca_convertstat_change_resol         
+
+! ------------------------------------------------------------------------------
+!> Create a 2d-mesh of grid coordinates from 1-d arrays.
```
guillaumevernieres: No need to copy paste: [please use the mom6 meshgrid](https://github.com/JCSDA/MOM6/blob/84c2432670659e3ca0aadb401a0834595ddac605/src/framework/MOM_horizontal_regridding.F90#L878)

src/soca/Utils/soca_convert_stats_mod.F90
```diff
@@ -0,0 +1,476 @@
...
+    x_T(i,j) = x(i)
+  enddo ; enddo
+
+  do j=1,nj ; do i=1,ni
+    y_T(i,j) = y(j)
+  enddo ; enddo
+
+end subroutine meshgrid
+
+! ------------------------------------------------------------------------------
```
guillaumevernieres: Same here: This is again too close to [mom6 fill_miss_2d](https://github.com/JCSDA/MOM6/blob/84c2432670659e3ca0aadb401a0834595ddac605/src/framework/MOM_horizontal_regridding.F90#L104)
Making use of MOM_memory.h seems appropriate in this case, although we do need to look closely at what's in that file.

src/soca/Utils/soca_convert_stats_mod.F90
```diff
@@ -0,0 +1,476 @@
...
+    if (k==1) prev(:,:) = tr_out(:,:)
+    call fill_miss_2d(tr_out, good, fill, prev=prev, G=self, &
+         smooth=.true.)
+
+    field2(:,:,k) = tr_out(:,:)*mask2(:,:,k)
+    prev(:,:) = field2(:,:,k)
+
+  end do
+
+end subroutine soca_hinterp
```
guillaumevernieres: Note to self: Lost focus here, need to look in more detail at hinterp.

test/testref/convertstate.test
```diff
@@ -1,16 +1,32 @@
...
+Test     :   hicen   min=    0.001532   max=    2.644540   mean=    0.520442
+Test     :   hsnon   min=    0.001055   max=    0.557002   mean=    0.170353
+Test     :    uocn   min=   -0.858166   max=    0.700095   mean=    0.000127
+Test     :    vocn   min=   -0.454376   max=    0.519210   mean=    0.001639
+Test     :     ssh   min=   -1.894600   max=    0.895962   mean=   -0.276549
+Test     :      sw   min= -224.275017   max=   -0.057449   mean= -110.272981
+Test     :     lhf   min=  -25.376282   max=  240.765434   mean=   64.166450
+Test     :     shf   min=  -53.379915   max=  178.082443   mean=    9.401263
+Test     :      lw   min=    0.008204   max=   84.720299   mean=   48.451667
+Test     :      us   min=    0.004462   max=    0.019176   mean=    0.009315
```
guillaumevernieres: We have the truth output state since it was sub-sampled from the original grid. I wonder if we should use something like nccmp with some tolerances that we'd be happy with. 
@clouden90 , did you compare the output state to the sub-sampled one? Some of the min/max are quite different between the input and output.

Also, we need an other test that converts from coarse to fine resolution, which is actually more appropriate for what you want to be doing ultimately.

guillaumevernieres (changes requested):
@clouden90 , I went through most of the code and left a few comments. Please have a look.

Things that needs to be done before you spend more time on that work:

1. Contact GFDL, or dig into MOM6, it seems to me that this feature already exists.
2. We need two `convert_state` tests: fine -> coarse and coarse -> fine
3. Since we have the exact solution of the 2 tests, we also need to compare the interpolated state to the truth.

That's a lot of work so thanks!

clouden90:
> @clouden90 , I went through most of the code and left a few comments. Please have a look.
> 
> Things that needs to be done before you spend more time on that work:
> 
> 1. Contact GFDL, or dig into MOM6, it seems to me that this feature already exists.
> 2. We need two `convert_state` tests: fine -> coarse and coarse -> fine
> 3. Since we have the exact solution of the 2 tests, we also need to compare the interpolated state to the truth.
> 
> That's a lot of work so thanks!

@guillaumevernieres , Thanks for the comments. I will have a look. Regarding the horizontal remapping feature, there are two subroutines existed in mom_horizontal_regridding module:
1. [horiz_interp_and_extrap_tracer_record](https://mom6.readthedocs.io/en/dev-gfdl/api/generated/modules/mom_horizontal_regridding.html#f/mom_horizontal_regridding/horiz_interp_and_extrap_tracer_record)
2. [horiz_interp_and_extrap_tracer_fms_id](https://mom6.readthedocs.io/en/dev-gfdl/api/generated/modules/mom_horizontal_regridding.html#f/mom_horizontal_regridding/horiz_interp_and_extrap_tracer_fms_id)

I spent some time on these two subroutines and found that either of them will require certain level of modification to fit our need. Maybe I was wrong. I can re-visit them again to see the possibility.

I think it is a good idea to have two 'convert_state tests', but I am not quite sure the comparison you propose will be an apple-to-apple comparison due to the different remapping algorithms are used. Perhaps we can discuss details offline.

guillaumevernieres:
@clouden90 : As a sanity check. You'll have a hard time convincing me that we should not compare to the truth when we have it! We don't expect bit reproducibility, but we expect a level of sanity we think we can live with. Makes sense?

clouden90:
> @clouden90 : As a sanity check. You'll have a hard time convincing me that we should not compare to the truth when we have it! We don't expect bit reproducibility, but we expect a level of sanity we think we can live with. Makes sense?

@guillaumevernieres : There is no doubt that we should have sanity check to validate the state converter we are developing. The question is how to conduct the check correctly. Please find attached figure and I will use it to explain my point. The attached figure shows SST from original 5-deg dataset (left); truth output (middle, sub-sampling from original dataset); and from state converter (right, bilinear interpolation). Let’s focus on the area with red circle. The original dataset contains two data points, and there is only one data point (middle point between the original data points) in the coarse grid. To me, the value extracted from “truth output” is simply a result of “nearest interpolation”, and the value from “state converter” is the average of original data points. In this case, I do not think it is not fair to compare the results from “state converter” to the one from “truth output” and simply make a conclusion that our “state converter” is wrong! Actually in this case the converter works as it should be and the “truth output” is not really a truth. Again I totally agree with your comments and I am fine with what you proposed (it at least can give us first-order comparison). I am just giving some ideas for a better way of performing the comparison (e.g. comparing domain mass conservation with original dataset maybe?) I hope this explanation will help you to understand my point. Again I truly appreciate your time and comments. Feel free to let me know if you have any other thoughts.
image

guillaumevernieres:
@clouden90 : Yup, find a set of metrics that you/we are happy with and use that to unit test the interpolation.

FYI: MOM6 can already 3d interpolate restarts, but according to @travissluka , only at tracer locations. Did you look into that mom6 feature?  I'm not saying this is possible, I'm just wondering if we could make use of already existing features that are part of the libraries we are using in this project.

clouden90:
> @clouden90 : Yup, find a set of metrics that you/we are happy with and use that to unit test the interpolation.
> 
> FYI: MOM6 can already 3d interpolate restarts, but according to @travissluka , only at tracer locations. Did you look into that mom6 feature? I'm not saying this is possible, I'm just wondering if we could make use of already existing features that are part of the libraries we are using in this project.

@guillaumevernieres : Thx, I will think about appropriate metrics that we can use to test the interpolation. And thanks for the info. As I mentioned before, I do found an existing feature for horizontal remapping in mom6 for state (T&S) and tracers variables but it's not quite fit to what we need because it requires input source file that contains stat/tracer info:
[horiz_interp_and_extrap_tracer()](https://noaa-gfdl.github.io/MOM6/APIs/namespacemom__tracer__initialization__from__z.html#a5f3dc84eae5f78304c8e366bf80e3afd)

The required inputs are:
subroutine, public mom_tracer_initialization_from_z::horiz_interp_and_extrap_tracer ( 
 character(len=*), intent(in) filename,
 character(len=*), intent(in) varnam,
 real, intent(in) conversion,
 integer, intent(in) recnum,
 type(ocean_grid_type), intent(inout) G,
 real, dimension(:,:,:), allocatable tr_z,
 real, dimension(:,:,:), allocatable mask_z,
 real, dimension(:), allocatable z_in,
 real, dimension(:), allocatable z_edges_in,
 real, intent(out) missing_value,
 logical, intent(in) reentrant_x,
 logical, intent(in) tripolar_n,
 logical, intent(in), optional homogenize
  )
You can find an example in soca/bundle/mom6/src/initialization/MOM_state_initialization.F90; LINE:2132

Maybe there are other features but so far this is the one I found.

src/soca/Utils/soca_convert_stats_mod.F90
clouden90: hocn2 is target grid cell thickness. We keep it here in case in the future if we have regional mom6 model setup and one would like to use the converted results to hotstart the model directly.

src/soca/Utils/soca_convert_stats_mod.F90
clouden90: Agree. See description section. Now mom_diag_remap module is used to get zstar grid.

src/soca/Utils/soca_convert_stats_mod.F90
clouden90: hocn has been set in soca_convertstate_setup so skip here.

src/soca/Utils/soca_convert_stats_mod.F90
clouden90: I also dislike repeat codes here. Unfortunately they are private subroutines in MOM6 and I have to keep them in this moment. Is there any way to get rid of them besides changing them from private to public?

src/soca/Utils/soca_convert_stats_mod.F90
```diff
@@ -0,0 +1,554 @@
+! (C) Copyright 2020-2020 UCAR.
```
guillaumevernieres: Replace `stats` with `state` in filename

src/soca/Utils/soca_convert_stats_mod.F90
```diff
@@ -0,0 +1,554 @@
...
+!
+! This software is licensed under the terms of the Apache Licence Version 2.0
+! which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+! ------------------------------------------------------------------------------
+
+module soca_convert_statse_mod
+
```
guillaumevernieres: typo: `state` not statse

src/soca/Utils/soca_convert_stats_mod.F90
```diff
@@ -0,0 +1,554 @@
...
+  call mpp_update_domains(tmp2, geom_des%Domain%mpp_domain)
+
+  ! Final step: vertical remapping to desired vertical coordinate
+  if(allocated(h1)) deallocate(h1)
+  if(allocated(h2)) deallocate(h2) 
+  allocate(h1(nz_),h2(field_des%nz))
+  if ( field_des%nz > 1 .and. field_des%io_file/="ice") then
+    do i = isc2-1, iec2+1
+      do j = jsc2-1, jec2+1
+        tmp_nz = nz_ !assume geom_src%nzo_zstar == geom%des%nzo_zstar
```
guillaumevernieres: Not amazingly important, but swap the indices in your loop. I have the same bad habit ...
Why loop outside of the compute domain if you fill the halo after the loop?

src/soca/Utils/soca_convert_stats_mod.F90
guillaumevernieres: Not sure why it's marked as resolved. See above comment, that routine is available in mom6.

src/soca/Utils/soca_convert_stats_mod.F90
guillaumevernieres: I don't think so. We made a few similar changes in our mom6 fork. I'd rather see a change in mom6 than copy paste of mom6 into soca.

test/testinput/convertstate.yml
```diff
@@ -10,12 +10,8 @@ outputresolution:
   num_ice_cat: 5
   num_ice_lev: 4
   num_sno_lev: 1
-  geom_grid_file: soca_gridspec.nc
-  mom6_input_nml: ./inputnml/input.nml
-varchange: Ana2Model
-rotate:
-  u: [uocn]
-  v: [vocn]
+  geom_grid_file: soca_gridspec.regional.nc
```
guillaumevernieres: why not use the sub-sampled grid?

src/soca/Utils/soca_convert_stats_mod.F90
clouden90: Got your point. Right now I cannot create branch in JSCDA/MOM6, can you add me in so I can make a PR and change meshgrid & fill_miss_2d to public

test/testinput/convertstate.yml
clouden90: The new test case is more close to "real-world" application (coarse global domain -> fine regional domain). Personally I think it is a better example to show users how the state converter works.

src/soca/Utils/soca_convert_stats_mod.F90
clouden90: Good catch. That's why we need the second eye! Indeed there is no need +1-1 outside the compute domain.

guillaumevernieres:
@clouden90 : travis-ci is "broken" for now. It will still fail when fixed though, your mom6 branch name doesn't match this one.

test/testinput/convertstate.yml
guillaumevernieres: I don't think it is since you are loosing access to the "true" state. Does it cover parts of your code that wouldn't be covered otherwise? If yes, then the test is needed, otherwise, remove and revert to using the other grid.

clouden90: Revert back to sub-sample-grid. Again as I replied before, I do not think the truth is really a "truth", it is just another remapping/interpolating method (nearest neighbor interpolation), and it's hard to say which one perform better by direct comparison.

guillaumevernieres: I'm open, really, you just need to justify why we need to carry an extra binary file in git-lfs. We spent quite a bit of time reducing the size of our repository, so your argument needs to be a bit more convincing than "I think it's better". You might be right, just make your case, concisely ...

clouden90: > I'm open, really, you just need to justify why we need to carry an extra binary file in git-lfs. We spent quite a bit of time reducing the size of our repository, so your argument needs to be a bit more convincing than "I think it's better". You might be right, just make your case, concisely ...

Will stick with sub-sample-grid for now because it's still a good example and demo. And it's also important to keep the whole package "light". Now I will move to test the new converter with soca-3dvar to check if there is anything else I miss ....

guillaumevernieres:
@clouden90 , sorry for the travis-ci failures over the last few days. It should work now, please merge develop into your branch and push. If you are working in containers, update your image. It was updated a few days ago.

test/Data/72x35x25/input_regional.nml
```diff
@@ -0,0 +1,28 @@
+ &MOM_input_nml
```
guillaumevernieres: I thought you got rid of the regional grid?

test/testinput/convertstate.yml
```diff
@@ -1,5 +1,5 @@
 inputVariables:
-  variables: &soca_vars [tocn, socn, hocn, cicen, uocn, vocn]
+  variables: &soca_vars [ssh, tocn, socn, hocn, cicen]
```
guillaumevernieres: What's the issue with the velocities?

src/soca/Utils/soca_convert_state_mod.F90
```diff
@@ -0,0 +1,397 @@
...
+  call fms_io_init()
+  
+  call read_data(trim(src%geom_grid_file), 'nzo_zstar', tmp(1), domain=src%Domain%mpp_domain)
+  src%nzo_zstar = tmp(1)
+
+  call read_data(trim(des%geom_grid_file), 'nzo_zstar', tmp(1), domain=des%Domain%mpp_domain)
+  des%nzo_zstar = tmp(1)
+  
+  if (des%nzo_zstar /= src%nzo_zstar) & 
+    call MOM_error(FATAL, "target nzo_zstar /= source nzo_zstar! Reset maximum depth in target grid MOM_input file and re-run soca gridgen")
```
guillaumevernieres: use fckit log, checkout the examples throughout this repository or look in fckit.

src/soca/Utils/soca_convert_state_mod.F90
```diff
@@ -0,0 +1,397 @@
...
+    if (field_src%io_file=="ocn") tmp(:,:,1) = field_src%val(:,:,1)*field_src%mask(:,:) !2D 
+    if (field_src%io_file=="sfc") tmp(:,:,1) = field_src%val(:,:,1) !2D no mask
+    if (field_src%io_file=="ice") tmp(:,:,1:nz_) = field_src%val(:,:,1:nz_)
+  end if ! field_src%nz > 1
+  call mpp_update_domains(tmp, geom_src%Domain%mpp_domain)
+
+  ! horizontal interp: convert src field to target field at zstar coord 
+  call mpp_global_field (geom_src%Domain%mpp_domain, tmp(:,:,1:nz_), gdata(:,:,1:nz_) )
+  call soca_hinterp(geom_des,tmp2(:,:,1:nz_),gdata,mask_(:,:),nz_,missing,lon_in,lat_in,field_des%lon,field_des%lat) !,lon_out,lat_out)
+
```
guillaumevernieres, line 191: remove comment

src/soca/Utils/soca_convert_state_mod.F90
```diff
@@ -0,0 +1,397 @@
...
+  call read_data(trim(src%geom_grid_file), 'h_zstar', src%h_zstar, domain=src%Domain%mpp_domain)
+
+  if (allocated(des%h_zstar)) deallocate(des%h_zstar)
+  allocate(des%h_zstar(des%isd:des%ied,des%jsd:des%jed,1:des%nzo_zstar))
+  call read_data(trim(des%geom_grid_file), 'h_zstar', des%h_zstar, domain=des%Domain%mpp_domain)
+
+  call fms_io_exit()  
+
+  !
+  allocate(self%hocn_src(src%isd:src%ied,src%jsd:src%jed,1:src%nzo))
```
guillaumevernieres, line 69: did you mean to add a comment here? comment or remove `!` ... same below.

test/testinput/convertstate.yml
clouden90: Originally got different answers for velocities with different compilers/platforms. Have fixed this problem and the ctest for converter has been updated with velocity fields.

src/soca/State/soca_state.interface.F90
```diff
@@ -267,7 +267,12 @@ subroutine soca_state_change_resol_c(c_key_fld,c_key_rhs) bind(c,name='soca_stat
     call soca_state_registry%get(c_key_rhs,rhs)
 
     ! TODO implement a proper change of resolution, just copying for now
-    call fld%copy(rhs)
+    if (size(fld%geom%lon,1)==size(rhs%geom%lon,1) .and. size(fld%geom%lat,2)==size(rhs%geom%lat,2) .and. &
```
guillaumevernieres, line 270: That's fine for now, I was supposed to add the "==" operator in geometry,and I didn't. See issue #348 that I conveniently closed without doing the work!
We can imagine cases for which the condition above is not sufficient. OK for this PR, but update the comment above to:
! TODO (Guillaume or Travis) implement == in geometry
or something to that effect.

test/testinput/convertstate.yml
```diff
@@ -1,5 +1,5 @@
 inputVariables:
-  variables: &soca_vars [tocn, socn, hocn, cicen, uocn, vocn]
+  variables: &soca_vars [ssh, tocn, socn, uocn, vocn, hocn, cicen]
 inputresolution:
```
guillaumevernieres, line 3: :+1:

guillaumevernieres (commented):
All good besides minor comments. I revived issue #348 which I was suppose to address for your work and I didn't ... But we'll need to revisit in the future since only checking the size of the domain isn't good enough to figure out if 2 geometries are the same ... Subject of an other PR, "soon".

guillaumevernieres (approved):

PR #353, opened 2020-05-12
Multiple geometries

### LINKED ISSUE #352: Handle fms hard coded "input.nml" to allow for multiple geometries
## Description
fms initializes variables from namelists from the hard coded file name "input.nml". 

## Solution
Try to keep the code backward compatible: Allow for the use of "input.nml" as default name. 
New feature to be added: Allow a different naming for "input.nml" passed from the configuration, copy or link at run-time

guillaumevernieres:
## Description
One step closer to be able to use multiple geometries. Despite the large number of files touched, the new features are confined in the Geometry class. Most of the files were touched to update the geometry configuration in the yaml files.

**New features:**
- Addition of a test to create a coarse geometry by sub-sampling the existing MOM6 netcdf files
- An fms struct to handle the hard-coded `input.nml`
- `input.nml` in the scratch directory is now a **protected name** This is going to break our scripts for GODAS (pinging @flampouris and @jkbk2004 ), the real time SOCA and our simple regular cycling scripts in soca. None of these issues are addressed in this work. 

... and just to be sure the message is received:
**THE LOCATION OF THE `input.nml` NAMELIST IS NOW REQUIRED IN THE GEOMETRY PART OF THE YAML FILES.**

Is a change of answers expected from this PR? **NO**

**Issues to debate:**
- I went against backward compatibility, in the long run we should assume that we will need multiple input.nml, thus having to remove the one in the scratch directory. It is safer to stage it from elsewhere. This is how @danholdaway implemented it for fv3-jedi, and I can't think of a better solution, short of modifying fms. 
- I added a test to sub-sample some of the mom6 files that we use, but it's slow (~4-5 sec on my computer), they could be added to the repo instead. 
- Meaning of life?

### Issue(s) addressed
closes #352 

## Testing

How were these changes tested? **Addition of a test for a coarser grid**
What compilers / HPCs was it tested with? **gcc debug/release**

## Dependencies
**NONE**

### DIFF
--- a/src/mains/GridGen.h
+++ b/src/mains/GridGen.h
@@ -35,8 +35,7 @@ namespace soca {
       const Geometry geom(geomconfig, this->getComm());
 
       //  Generate model grid
-      const eckit::LocalConfiguration gridgenconfig(fullConfig, "gridgen");
-      geom.gridgen(gridgenconfig);
+      geom.gridgen();
 
       return 0;
     }
--- a/src/soca/Geometry/CMakeLists.txt
+++ b/src/soca/Geometry/CMakeLists.txt
@@ -4,4 +4,6 @@ soca_target_sources(
   GeometryFortran.h
   soca_geom_mod.F90
   soca_geom.interface.F90
+  FmsInput.cc
+  FmsInput.h
 )
--- a/src/soca/Geometry/FmsInput.cc
+++ b/src/soca/Geometry/FmsInput.cc
@@ -1,0 +1,79 @@
+/*
+ * (C) Copyright 2020-2020 UCAR
+ *
+ * This software is licensed under the terms of the Apache Licence Version 2.0
+ * which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+#include "soca/Geometry/FmsInput.h"
+
+// -----------------------------------------------------------------------------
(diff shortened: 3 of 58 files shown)


aerorahul:
changes look good to the eye.
will do a deep dive and test later today.

I am ok with the sub-sampling script. This is done in development-mode only.  I am against increasing the size of the repo unnecessarily.

aerorahul (approved):
I don't like the symlink in the C++ code, but there is no good solution in the short term.
Other than that, all tests pass for me.

clouden90 (approved):
Just tested this and it works with the prototype convert app.

travissluka (approved):

danholdaway:
For fv3-jedi we also need to stage the input.nml in the Model class. Are you doing initialization of mom6 there because that might require a different input.nml file, for example if you were doing multiple outer loops with varying resolution.

guillaumevernieres:
> For fv3-jedi we also need to stage the input.nml in the Model class. Are you doing initialization of mom6 there because that might require a different input.nml file, for example if you were doing multiple outer loops with varying resolution.

Thanks for the info @danholdaway, yes, we are (keeping your tip in a fuzzy corner of my mind). Probably some future work for @LorentzSluka.

PR #363, opened 2020-06-03
travis-ci failing in new container with gcc 9.3

### LINKED ISSUE #362: travis-ci failing with new container
## Description
The test for empty character array doesn't work with gcc 9.3 in debug mode (fine in release). 

## Solution
Check length instead of "emptiness".

guillaumevernieres:
## Description
Multiple ctests are failing when using the newly updated container. Note that the failure only happens in debug mode and is caused by a weird behavior in testing the "emptiness" of an allocatable character array. The test was simply replaced by checking the length instead. 
   
closes #362

### DIFF
--- a/src/soca/Fields/soca_fields_mod.F90
+++ b/src/soca/Fields/soca_fields_mod.F90
@@ -1040,7 +1040,7 @@ subroutine soca_fields_write_rst(fld, f_conf, vdate)
       idr = register_restart_field( ice_restart, ice_filename, "vsnon", &
         vsnon, domain=fld%geom%Domain%mpp_domain)
 
-    else if (field%io_file /= "") then
+    else if (len(field%io_file) /= 0) then
       ! which file are we writing to
       select case(field%io_file)
       case ('ocn')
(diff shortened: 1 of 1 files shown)


src/soca/Fields/soca_fields_mod.F90
```diff
@@ -1040,7 +1040,7 @@ subroutine soca_fields_write_rst(fld, f_conf, vdate)
       idr = register_restart_field( ice_restart, ice_filename, "vsnon", &
         vsnon, domain=fld%geom%Domain%mpp_domain)
 
-    else if (field%io_file /= "") then
+    else if (len(field%io_file) /= 0) then
```
aerorahul: if you want to use `len`, use `len_trim` to get rid of empty spaces.

aerorahul: Not sure, why this is a bug, and why it has surfaced now.  I use GNU 9.3.0 on my mac and Linux box, and did not have an issue.
Either way, the change should be fine.  I would instead use `len_trim` to remove trailing blanks.

aerorahul (approved):
approving, but please change `len` to `len_trim`.

guillaumevernieres:
> approving, but please change `len` to `len_trim`.

Done, thanks for the tip.

PR #364, opened 2020-06-04
More flexible obs space in cycling scripts

guillaumevernieres:
## Description

Updates to the cycling scripts. 
- Added GEOS forecast model
- Switching the obsspace setup in the 3dvar yaml to do something similar to what we have for EMC
- Updates to how we handle `input.nml`

### Issue(s) addressed

We should be able to re-use what we developed for EMC in terms of the ObsSpace setup, which is done per instrument, as opposed to general obs types (e.g. gmi.sst instead of sst ... etc)

## Testing

Tested on discover and linux workstations. This code is not part of our automated travis-ci tests.
These are very simple and easily portable scripts, I will add the necessary data to allow us to use them on Hera and Orion.

## Dependencies

[soca-config pr#20](https://github.com/JCSDA/soca-config/pull/20)

### DIFF
--- a/scripts/workflow/apps/exp.config.mom6solo
+++ b/scripts/workflow/apps/exp.config.mom6solo
@@ -2,8 +2,8 @@ EXP_DIR=$(readlink -f .)
 
 WORKLOAD_MANAGER=slurm  # none or slurm
 MACHINE=discover        # default or discover (default is a Linux workstation, no modules loaded)
-                       # TODO(Guillaume) add Hera and Orion
-MPIRUN=mpirun          #TODO could probably get the info from somewhere else
+                        # TODO(Guillaume) add Hera and Orion
+MPIRUN=mpirun           # TODO could probably get the info from somewhere else
 
 # parameters used by SLURM
 JOB_NAME=exp13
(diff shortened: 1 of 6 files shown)


scripts/workflow/apps/exp.config.mom6solo
```diff
@@ -0,0 +1,52 @@
+EXP_DIR=$(readlink -f .)
```
aerorahul, line 1: This is specific to GEOS and Discover.
The filename should reflect that. It is not generic.

aerorahul: it also contains specific references to some one named gvernier and userspace.
Double no-no.

guillaumevernieres: @aerorahul exp.config.discover.mom6solo?

guillaumevernieres: > it also contains specific references to some one named gvernier and userspace.
> Double no-no.

great guy so I've heard.

aerorahul: > @aerorahul exp.config.discover.mom6solo?
It runs GEOS via gcm_run.j so it is not mom6solo is it?

aerorahul (approved):
much better.
Alas, I can't test on Discover. Something to do with my desertion.
I will try and cycle on my desktop.

Issue #370, opened 2020-06-10
atlas fails to build in travis-ci


guillaumevernieres:
## Description
atlas build started failing earlier today (2020-06-10)
```
/jcsda/work/repo.src/atlas/src/atlas_f/internals/atlas_write_to_fortran_unit.F90:11:36:

   11 |   use fckit_c_interop_module, only : copy_c_ptr_to_string

      |                                    1

Error: Symbol ‘copy_c_ptr_to_string’ referenced at (1) not found in module ‘fckit_c_interop_module’
```

mer-a-o:
@guillaumevernieres: you need to use release-stable branch for atlas. I think Travis-CI is using develop branch:
https://travis-ci.com/github/JCSDA/soca/builds/170692092#L283

guillaumevernieres:
> @guillaumevernieres: you need to use release-stable branch for atlas. I think Travis-CI is using develop branch:
> https://travis-ci.com/github/JCSDA/soca/builds/170692092#L283

oops ... l probably reverted to develop in the last pr. Thanks @mer-a-o . 
Update: Just looked at the code and we are pointing to release-stable, not sure why travis-ci is pointing at develop. Digging in the script will happen tomorrow!

guillaumevernieres:
> @guillaumevernieres: you need to use release-stable branch for atlas. I think Travis-CI is using develop branch:
> https://travis-ci.com/github/JCSDA/soca/builds/170692092#L283

@mer-a-o I cleaned the cache and rebuilt (my one and only strategy to fix travis-ci!) and that did not do it. I made a local build the same way travis-ci does, in singularity, and it built/tested fine. [This branch](https://github.com/JCSDA/soca/pull/371,) which has zero code change tests fine, but develop keeps failing. At a loss. Taking a break from the issue, but pointers would be greatly appreciated! 
Was the container updated 2 days ago?

mer-a-o:
@guillaumevernieres, looks like https://github.com/JCSDA/soca/pull/371 used `release-stable` branch of atlas and tests passed. But for Cron builds `develop` is used. Are the settings for branch names different for Cron jobs?

guillaumevernieres:
@mer-a-o . I don't think so, but will check. I'm going to try to turn off the branch matching also ... for now.

mer-a-o:
I think you need to remove atlas from `MATCH_REPOS` in .travis.yml.

guillaumevernieres:
[...]

PR #371, opened 2020-06-11
Compiler cache build instructions

guillaumevernieres:
## Description

In my current attempt at figuring out why travis-ci is currently failing, while the tests are fine on most hpc and containers, I built soca the same way travis-ci does. This is mainly a note to self for future reference, because I WILL forget. It might be of use to somebody else.

### Issue(s) addressed

related to #370 but unfortunately does not fix it.

## Testing

Tested within singularity.  I unfortunately cannot replicate the current travis-ci failure, soca builds and tests fine, and the ecmwf repos are pointing to the correct stable branches.


guillaumevernieres:
Now that's interesting, no failure :point_down:   ...

aerorahul (approved):

PR #376, opened 2020-06-15
pointing to dev/master-ecbuild

guillaumevernieres:
## Description
Pointing to the proper mom6 branch.


aerorahul (approved):
makes sense.

PR #381, opened 2020-06-18
Broken letkf test

guillaumevernieres:
## Description
The letkf test keeps failing (randomly?) in the compare test, for some reason the script isn't using the prescribed tolerance.  I'm not sure why ...
This bugfix isn't addressing the root cause of the problem, but it should fix the sometimes broken test.


frolovsa (approved):
Stopgap it is.

PR #388, opened 2020-06-23
Bugfix/recursive default

guillaumevernieres:
## Description
@aerorahul 
Merged develop and clone recursively by default in prep.sh.

### DIFF
--- a/.github/travisci/prep.sh
+++ b/.github/travisci/prep.sh
@@ -104,7 +104,7 @@ for repo in $LIB_REPOS $MAIN_REPO; do
     if [[ $rebuild == 1 || $skip_lfs == 0 ]]; then
         repo_src_dir=repo.src/$repo
         rm -rf $repo_src_dir
-        GIT_LFS_SKIP_SMUDGE=$skip_lfs git clone -b $repo_branch $repo_url $repo_src_dir
+        GIT_LFS_SKIP_SMUDGE=$skip_lfs git clone --recursive -b $repo_branch $repo_url $repo_src_dir
     fi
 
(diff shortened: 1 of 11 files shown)


aerorahul (approved):
Not sure what atlas geometry is being used in soca for. 
Jo seems quite a bit off.

guillaumevernieres:
@aerorahul , I was hopping it would trigger a travis-ci build/test, I guess not. The --recursive should fix the issue.

guillaumevernieres:
> Not sure what atlas geometry is being used in soca for.
> Jo seems quite a bit off.

@benjaminmenetrier 's pr that was merged a few days ago. Using atlas for B.

benjaminmenetrier:
@aerorahul: SOCA used some FCKIT functions for geometry operations (like great-circle calculations) that I had implemented in the JCSDA fork. However, ECMWF people prefer to keep the geometry stuff in ATLAS instead of FCKIT, so we now use the `atlas_geometry` object.

aerorahul:
> @aerorahul: SOCA used some FCKIT functions for geometry operations (like great-circle calculations) that I had implemented in the JCSDA fork. However, ECMWF people prefer to keep the geometry stuff in ATLAS instead of FCKIT, so we now use the `atlas_geometry` object.

@benjaminmenetrier thanks for the info. doesn’t eckit have distance methods ? Just curious

benjaminmenetrier:
You are right, I think that `atlas` calls to sphere distance or kdtrees are just pretty C++ and Fortran wrapper of `eckit` functions.

PR #397, opened 2020-07-06
fix for oops ecbuild-update-r1

travissluka:
## Description

https://github.com/JCSDA/oops/pull/639 breaks soca builds on TravisCI. This fixes it. See https://github.com/JCSDA/oops/pull/639#issuecomment-654265345 for further description

### Issue(s) addressed

- uses the `<project>_DIR=` variable instead of the previously working `<PROJECT>_PATH=`
- changes how the `compare.py` and `cpplint.py` files are found
- changes how oops toy models are disabled

## Dependencies

Must use the matching branches in oops, saber, ufo, ioda


.github/travisci/build.sh
```diff
@@ -26,9 +26,9 @@ for repo in $LIB_REPOS; do
     install_dir=${REPO_CACHE}/$repo
 
     # set the path to the install dir for subsequent repos to find
-    repo_upper=${repo/-/_}; repo_upper=${repo_upper^^}
-    typeset "${repo_upper}_PATH=$install_dir"
-    export ${repo_upper}_PATH
+    repo_underscore=${repo/-/_}
+    typeset "${repo_underscore}_DIR=$install_dir"
```
markjolah, line 30: I think this should be `${repo_underscore}_ROOT`.  That is what CMake find_package() uses and is going to be the one and only way to set these paths going forward.  

Of note, if you are installing all these packages to the same install prefix, once a package is installed there, subsequent CMakes installing to the same prefix  should find the other packages there automatically without needing a `foo_ROOT` env var.

src/CMakeLists.txt
```diff
@@ -6,11 +6,8 @@
 add_subdirectory( soca )
 add_subdirectory( mains )
 
-# find cpplint.py, if not found assume it will be built by oops in this bundle
-find_file(CPPLINT_PY cpplint.py HINTS ${OOPS_PATH}/bin)
-if(NOT CPPLINT_PY)
-  set(CPPLINT_PY ${CMAKE_BINARY_DIR}/bin/cpplint.py)
-endif()
+# find cpplint.py provided by oops
+find_file(CPPLINT_PY oops_cpplint.py HINTS ${oops_BINDIR})
```
markjolah, line 10: See I'm making your life easier here!

markjolah (approved):
Great.  Sorry I didn't warn you about the toy models flags changing.

Approve pending something along the line of `oops_ROOT=$install_prefix` works for you.  Don't use the `oops_DIR` or `oops_PATH` variables.

.github/travisci/build.sh
travissluka: The install prefix is actually different for each repo as the cache needs to be maintained separately for each repo. If an upstream repo changes, its cache is flushed as well as all repos that come after it, leaving the repos before it untouched.

When the `{repo}_PATH` stopped working, i just went with what the ecbuild log was telling me to use (`{repo}_DIR`)

guillaumevernieres (approved):

PR #409, opened 2020-07-15
Update to the  ensemble variance test

### LINKED ISSUE #408: Add a variable change to the ensemble variance  test
## Description
The current `ensvariance` test has no variable change, which will be needed if one derives the background error from an ensemble and wants to use one of the invertible soca variable change or the log transform that @liuxiao37k will be working on. 

## Solution
Add an example on how to add one.

## Alternatives
Do nothing.

## Related to
[oops issue #](https://github.com/JCSDA/oops/issues/737)

guillaumevernieres:
## Description
I just noticed that one of the yaml file I use to compute the the variance with the inverse of the balance operator was in fact not applying the variable change ... Not a huge deal since I'm probably the only one playing with that stuff now, but, since the yaml file in the tests are used as a base to build from, adding an example of change of variable is probably a good idea. 

### Issue(s) addressed
closes #408 

## Testing

How were these changes tested? **added to the ensemble variance test**

## Dependencies
None

### DIFF
--- a/test/testinput/ensvariance.yml
+++ b/test/testinput/ensvariance.yml
@@ -45,3 +45,13 @@ Ensemble:
   - <<: *_file
     ocn_filename: ocn.pert.ens.5.2018-04-15T00:00:00Z.PT6H.nc
     ice_filename: ice.pert.ens.5.2018-04-15T00:00:00Z.PT6H.nc
+  variable_changes:
+  - varchange: BalanceSOCA
+    dsdtmax: 1.0
+    dsdzmin: 3.0e-3
+    dtdzmin: 1.0e-3
+    nlayers: 10
+    inputVariables:
+      variables: *soca_vars
+    outputVariables:
+      variables: *soca_vars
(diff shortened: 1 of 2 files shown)


liuxiao37k (approved):
tested on Hera and all soca ctests passed with this update.

liuxiao37k:
@guillaumevernieres just out of curiosity, do variable_change applications follow a sequential order? e.g. in the 3dvar_soca case, BkgErrFILT change occurs first then BkgErroSOCA, VertConvSOCA, and BalanceSOCA?

travissluka (approved):

guillaumevernieres:
> @guillaumevernieres just out of curiosity, do variable_change applications follow a sequential order? e.g. in the 3dvar_soca case, BkgErrFILT change occurs first then BkgErroSOCA, VertConvSOCA, and BalanceSOCA?

Thanks for your review @liuxiao37k . That's right, see [oops/base/ModelSpaceCovarianceBase.h](https://github.com/JCSDA/oops/blob/ff765be6035ec00ca322818113f0a02c1e0a69e5/src/oops/base/ModelSpaceCovarianceBase.h#L198) , that is where the "magic" happens.

PR #411, opened 2020-07-18
Options to mask out balances

### LINKED ISSUE #410: Specify regions where the balance will not be applied
## Description
Current issues when assimilating adt and using the balance as a strong constraint, which is what is needed to constrain the mesoscale :
1. Unrealistic T and S in the mixed layer
2. "Regional Instabilities"

## Solution
1. Can be somewhat mitigated by assimilating sst and sss when the DA window is at least 24hr, but not ideal.
2. Zero out the Jacobian in critical places: top of the mixed layer and identified problematic regions (currently SW of Australia).
 
## Alternatives
1. Derive a better parametric background error
2. May not be an issue when using ensemble based DA ... may ...
3. Increase the unbalanced background error for ssh in problematic regions, but ssh is never fed back to the model. Good for making pretty analysis plot but practically useless.
4. Pass the errors or representation from the model to a ufo qc filter? I don't think this is implemented in ufo yet ...

guillaumevernieres:
## Description
The issue is described in #410. 
This PR adds the following feature:
- Read dc/dt from file, this will change the GODAS results ... for the better dc/dt=-0.01K^-1 everywhere wasn't meant to be
- Several ways to zero out the dynamic height jacobians:
   - First nlayers 
   - Entire water column where masked  
   - deta/dt and/or deta/ds globally
Not an ideal solution, but a useful tool to have when instabilities develop in some regions due to the combined assimilation of adt and the use of the balance. 

Example of usage of the mask in varchange_balance_TSSSH.yml:
```yaml
  ...
  jac_mask:
    filename: ./Data/kmask.nc
    name: kmask
    threshold: 0.3
    detadt: false
    detads: false
    nlayers: 10
```
will zero out the Jacobian of the balance where mask<0.3 and in the first 10 upper layers.
and,
```yaml
  dcdt:
    filename: ./Data/kmask.nc
    name: dcdt
```
will read dc/dt from ./Data/kmask.nc

### Issue(s) addressed
closes #410 

## Testing
What compilers / HPCs was it tested with? **gnu & intel**, **orion & singularity**
Are the changes covered by ctests? (If not, tests should be added first.) **yes**

## Dependencies
None


clouden90 (approved):
Codes look good by eyes. Tested on Hera and passed. May not be a final solution but it's ok for now.

travissluka (approved):
seems fine from a coding perspective... i'll just go ahead and assume you verified it from a science perspective.

test/CMakeLists.txt
```diff
@@ -409,6 +412,18 @@ soca_add_test( NAME static_socaerror_init
...
+# Use the soca_dirac.x application to create a mask for the dynamic height Jacobians
+soca_add_test( NAME balance_mask
+               EXE  soca_dirac.x
+               TEST_DEPENDS test_soca_gridgen )
+
+# Create dc/dt file for ocean/ice balance and convert dirac file into a mask
+add_test ( test_soca_create_kmask
+           python
+           ${CMAKE_CURRENT_SOURCE_DIR}/Data/create_kmask.py
+           TEST_DEPENDS test_static_socaerror_init
```
liuxiao37k, line 424: @guillaumevernieres I don't know if this should be the proper place to ask this question (sorry if it's not). Is this possibly a bug? Once I tried to run `test_soca_create_kmask`, it complained something because the dependency tests `test_static_socaerror_init` and `test_balance_mask` don't exist. I think `test_soca_static_socaerror_init` and `test_soca_balance_mask` do?

travissluka: dont worry  about the kmask ctest... its about to be deleted, I will have a PR that cleans up the ctests quite a bit

PR #415, opened 2020-07-31
Bugfix in ATLAS function space

benjaminmenetrier:
Minor bugfix, for a bug that was curiously silent... I also added `only: [...]` with the required objects after the `use atlas_module` occurrences.

Attempted explanation for the silence of this bug: https://github.com/JCSDA/mpas-jedi/pull/344#issuecomment-667249862


benjaminmenetrier:
3 tests fail here whereas they all pass on my laptop. Any idea?

travissluka:
The first failure is in `test_soca_3dvar_godas`, the second two failures depend on the output of this test. I'm guessing something unrelated is using uninitialized memory and the error has been hidden until now. I'll debug it this week.

On a side note,  I noticed recently that  `test_soca_3dvar_godas` has been going an order of magnitude slower than it should be. I'm hoping these two issues are related and I can knock out two birds with one stone.

travissluka (approved):
.... still have no idea why it was failing at one point. We have gremlins in the system.

PR #416, opened 2020-08-03
speedup kdtree creation used by godas bgerr

travissluka:
## Description

when the kdtree used by soca was switched to the atlas one, any test that used the godas bgerr went a lot slower (though we didn't notice it at the time)

Looking back over the TravisCI logs:
Before switch to atlas:
```
28/48 Test #28: test_soca_varchange_bkgerrgodas .....   Passed    4.00 sec
````

and after:
```
32/55 Test #32: test_soca_varchange_bkgerrgodas .....   Passed   39.02 sec
```

This speeds up the kdtree build by reserving the correct amount of space beforehand

### DIFF
--- a/src/soca/Utils/soca_utils.F90
+++ b/src/soca/Utils/soca_utils.F90
@@ -187,6 +187,7 @@ subroutine soca_remap_idw(lon_src, lat_src, data_src, lon_dst, lat_dst, data_dst
   n_src = size(lon_src)
   ageometry = atlas_geometry("UnitSphere")
   kd = atlas_indexkdtree(ageometry)
+  call kd%reserve(n_src)
   call kd%build(n_src, lon_src, lat_src)
 
   ! remap


ytremolet:
Just curious: what's the run time now?

travissluka:
> Just curious: what's the run time now?

@ytremolet much better now. I'm not sure if the remaining 1 sec difference is kdtree related, or other stuff in the code that has been changed since then.
``` 32/55 Test #32: test_soca_varchange_bkgerrgodas .....   Passed    5.06 sec```

Turns out that the kdtree was used by a lot of the tests, since it was also used to remap the rossby radius fields. So overall ctests time on TravisCI is down from `229.37 sec` to `125.90 sec` now.

travissluka:
@benjaminmenetrier I don't know if you used the atlas kdtree in any of the other repos, perhaps it would be best to do an inerhent `reserve()` inside the `build()` or `insert()` , not sure if that's best on the fortran or c side, i haven't looked at the atlas code much.

guillaumevernieres (approved):

benjaminmenetrier:
Thanks @travissluka, I'll add this call to `%reserve` in the BUMP source `type_tree.F90`.

PR #423, opened 2020-08-25
DA at multiple resolution

### LINKED ISSUE #417: DA at Multiple resolution
## Description
Use different geometry for B and h(x) in the DA applications.

## Solution
Expand the use of the change of resolution to Increment.

### LINKED ISSUE #436: VertConv and BkgErr will currently not work with different geometries between bkg (or traj) and incr.
## Description
- VertConv: Missing layer depth after interpolation
- BkgErr: Not sure what the issue is, but it fails when interpolating the background. Investigation needed, but that variable chnage needs to be simplified and cleaned up.

guillaumevernieres:
## Description
This feature will allow different geometries to be used for h(x) and B.
Features:
-  Variable changes: Use the correct geometry and interpolate bkg/traj to the "B" geometry 
- Increment interpolation: Similar to State, but only uses a 2d interpolation
- Addition of 3dvar test using 2 geometries  

### Issue(s) addressed
closes #417
closes #436

### Issues **NOT** addressed
- VertConv and BkgErr will currently not work with different geometries between bkg (or traj) and incr. (#436 )
- Duplication of States in Increment "diff"

## Testing

How were these changes tested? **in container only**
Are the changes covered by ctests? **addition of a 3dvar test with 2 geometries**


travissluka:
answers for the new test don't match https://travis-ci.com/github/JCSDA/soca/builds/181405749#L4062
Did you run on your machine with `-DSOCA_TESTS_FORC_DEFAULT_TOL=ON" ?

guillaumevernieres:
> answers for the new test don't match https://travis-ci.com/github/JCSDA/soca/builds/181405749#L4062
> Did you run on your machine with `-DSOCA_TESTS_FORC_DEFAULT_TOL=ON" ?

nope ...
Looks like this is the difference between release and debug. 1e-9 is too tight for sure, but some of the differences are quite large. Will investigate.

src/soca/Increment/Increment.cc
```diff
@@ -82,11 +84,10 @@ namespace soca {
   void Increment::diff(const State & x1, const State & x2) {
     ASSERT(this->validTime() == x1.validTime());
     ASSERT(this->validTime() == x2.validTime());
-    Log::debug() << "Increment:diff incr " << *this << std::endl;
-    Log::debug() << "Increment:diff x1 " << x1 << std::endl;
-    Log::debug() << "Increment:diff x2 " << x2 << std::endl;
-    soca_increment_diff_incr_f90(toFortran(), x1.toFortran(),
-                             x2.toFortran());
+    State x1_lr(*geom_, x1);
+    State x2_lr(*geom_, x2);
```
travissluka: I don't like this. If no conversion is needed then we are unnecessarily wasting memory by having extra State copies.

(I might be able to be convinced to ignore it and we'll deal with it later when we're ready to do some hardcore soca optimization)

src/soca/State/State.cc
```diff
@@ -39,7 +39,9 @@ namespace soca {
   }
   // -----------------------------------------------------------------------------
   State::State(const Geometry & geom, const eckit::Configuration & file)
-    : time_(), vars_(file, "state variables"), geom_(new Geometry(geom))
+    : time_(9999, 1, 1, 1, 0, 0),
```
travissluka: I guess we're in trouble if someone is still using this code 7979 years from now??

travissluka (commented):
by changing   `const State & traj_;` inside the filters to `State traj_lr;` (i.e. dropping the `&`) an entirely new state is created and saved in each filter. Memory is wasted. 

The `test_soca_3dvar_godas` before:
```
46: OOPS_STATS soca::State                     : simultaneous =     5, total =     10
```

and now:
```
47: OOPS_STATS soca::State                     : simultaneous =     8, total =     20
```

Also I wouldn't call it `_lr`, its not necessarily a lower resolution. A single state should be created and passed around with either smart pointers, or stored in a single location and passed around via references (as before)

src/soca/Increment/Increment.cc
guillaumevernieres: > by changing `const State & traj_;` inside the filters to `State traj_lr;` (i.e. dropping the `&`) an entirely new state is created and saved in each filter. Memory is wasted.
> 
> The `test_soca_3dvar_godas` before:
> 
> ```
> 46: OOPS_STATS soca::State                     : simultaneous =     5, total =     10
> ```
> 
> and now:
> 
> ```
> 47: OOPS_STATS soca::State                     : simultaneous =     8, total =     20
> ```
> 
> Also I wouldn't call it `_lr`, its not necessarily a lower resolution. A single state should be created and passed around with either smart pointers, or stored in a single location and passed around via references (as before)

All valid point. I'm actually not sure I need traj, bkg and geom as data members in any of the transforms. I'll check  and switch to smart pointers or simply remove them. Thanks for pointing it out.

src/soca/State/State.cc
guillaumevernieres: Easy pr for the Guillaume of 9999 ... Thinking ahead.

guillaumevernieres:
@travissluka : Just addressed travis-ci's tantrum about uncovered code. I think this is ready for review ... And as I write this, I realized I didn't check valgrind. Oops ... I'm sure it's fine!

travissluka:
i'm running it through valgrind now

test/testinput/3dvarlowres_soca.yml
```diff
@@ -0,0 +1,202 @@
...
+    num_ice_lev: 4
+    num_sno_lev: 1
+    geom_grid_file: soca_gridspec.nc
+    mom6_input_nml: ./inputnml/input.nml
+
+  model:
+    name: SOCA
+    tstep: PT1H
+    advance_mom6: 0
+    model variables: *soca_vars
```
travissluka: model isn't needed in yamls anymore for 3dvar

test/testinput/3dvarlowres_soca.yml
```diff
@@ -0,0 +1,202 @@
...
+      num_ice_cat: 5
+      num_ice_lev: 4
+      num_sno_lev: 1
+      geom_grid_file: soca_gridspec.small.nc
+      mom6_input_nml: ./inputnml/input_small.nml
+    linear model:
+      variable change: Identity
+      name: IdTLM
+      tstep: PT1H
+      lm variables: [tocn, socn, ssh, hocn]
```
travissluka: not needed anymore for 3dvar

travissluka (commented):
... I'm still waiting to test under valgrind. Looks good, though I'm sure I'll complain in the future when we start doing profiling/optimizations

travissluka (approved):
unsurprisingly, valgrind has issues with the mom6 regridding. I added supressions to handle this. all good.

PR #427, opened 2020-09-01
use crtm_COEFFICIENT_DIR to link crtm coeff

mer-a-o:
## Description

`crtm_COEFFICIENT_DIR` can now be used to link crtm coefficients for testing. This has been used in UFO and tested when building inside and outside of the bundle. 

## Testing
I tested this branch using soca-bundle and everything passed on Cheyenne with gnu/openmpi. 

There are failures in the clang container when I used the bundle, Travis-CI method, and jedi-build-package for building. These issues are not caused by the changes in this PR. I will issue a separate PR for that.


travissluka (approved):
thanks @mer-a-o !

PR #440, opened 2020-09-15
Feature/3dbump

### LINKED ISSUE #439: Add back 3dbump
## Description
Use 3D bump from the saber covariance factory as an alternative to using 2d bump combined with the vertical convolution.
- Add a test to initialize BUMP at LOW resolution
- Add a 3DVAR test with B at LOW resolution

## Related to
#423

guillaumevernieres:
## Description
Minor updates to allow for the use of the saber covariance factory with the variational application. I still don't think we will be able to initialize bump correlations at full resolution, but perhaps at a sub-sampled resolution. The motivation for that work is that bump is perfectly normalized, the soca VertConv is not ... 

### Addition of 2 tests:
1. Initialization of BUMP correlation operator. Use of unrealistic scales to speed up the testing
2. Example of the use of 1. for a 3dvar case at low resolution and only a sub-sample of the possible variables (no sea-ice, no chl, ...)

### Issue(s) addressed
closes #439 

## Testing
How were these changes tested? **Use of saber covariance factory tested with a 3dvar**
What compilers / HPCs was it tested with? **gnu/container**
Are the changes covered by ctests? **Yes, 2 tests added**

### DIFF
--- a/src/mains/3DVar.cc
+++ b/src/mains/3DVar.cc
@@ -17,6 +17,7 @@
 #include "oops/runs/Variational.h"
 #include "oops/generic/instantiateModelFactory.h"
 #include "saber/oops/instantiateLocalizationFactory.h"
+#include "saber/oops/instantiateCovarFactory.h"
 
 int main(int argc,  char ** argv) {
   oops::Run run(argc, argv);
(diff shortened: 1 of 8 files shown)


benjaminmenetrier (approved):
Looks good! I like the mention that "bump is perfectly normalized" in the PR description!

travissluka (approved):

travissluka (approved):

travissluka:
... for the record, i'm not 2x my normal excitement about approving this! Github was being glitchy.

PR #446, opened 2020-09-18
Removal of obsolete repo from the bundle

guillaumevernieres:
## Description
soca-config is obsolete. This PR replaces it with soca-science

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ env:
   global:
     - MAIN_REPO=soca
     - LIB_REPOS="fms gsw mom6 crtm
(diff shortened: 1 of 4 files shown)


travissluka:
`soca-config` was in that bundle because at one point the cycling scripts were contained in this repo. I'm not sure if we should put `soca-science` in this bundle anymore because nothing here uses `soca-science`. The dependency is the other way around, `soca-science` depends on `soca`

Also, you need to change this 
https://github.com/JCSDA/soca/blob/develop/.travis.yml#L31

travissluka (approved):

PR #456, opened 2020-10-03
Fix after j cleanup 3

travissluka:
## Description

A fix after https://github.com/JCSDA/ioda/pull/484 broke soca.... I guess because we do weird things with soca that no one else is doing ?? :man_shrugging:


guillaumevernieres (approved):
Some of soca's finest coding history right there! Thanks for the fix.

PR #457, opened 2020-10-06
fix interface and geometry iterator test

travissluka:
## Description

someone made a change to the yaml syntax required for the interface and geometry iterator tests. Good to know. All tests should pass.... for now.... the day is early. (https://github.com/JCSDA/oops/pull/868)

I guess I forgot to change branches before doing that fix, and my `accumul` fix snuck in here. Might as well merge that as well because it will get tested soon anyway. `Increment::accumul()` was just using the `axpy` interface, but it can't do that because `axpy` takes an increment, whereas `accumul` takes a state. as a result it was using the wrong registry in the interface . Luckily it seems nothing we have was actually using `accumul` . This will get tested later.

### DIFF
--- a/src/soca/Increment/Increment.cc
+++ b/src/soca/Increment/Increment.cc
@@ -132,7 +132,7 @@ namespace soca {
   }
   // -----------------------------------------------------------------------------
   void Increment::accumul(const double & zz, const State & xx) {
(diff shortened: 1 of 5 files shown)


guillaumevernieres (approved):
The new code coverage is make me cry a little :cry:

PR #462, opened 2020-10-12
update letkf tests for oops/ioda fix to letkf

travissluka:
## Description

This PR accompanies the matching oops and ioda branches that fix the use of QC flags with LETKF.
- small change to the obs filters and localization radius used in the LETKF test... not really necessary, but it increases the number of obs passing QC enough so that the O-A RMS is actually smaller than the O-B RMS... just to make me feel warm and fuzzy.
- fix `random seed` syntax so that the random seed is indeed set to zero now
- update letkf test answers

## Testing

- ctests are happy
- checked the obs output files to confirm `@ObsError` is now masked out for obs that failed QC (it wasn't before)

## Dependencies

- depends on ioda and oops branches of same name (https://github.com/JCSDA/ioda/pull/489 , https://github.com/JCSDA/oops/pull/1089 )
- all need to be merged at the same time


guillaumevernieres (approved):
Thanks @travissluka . I wonder how many of the legacy `_` we are still using our yaml files ...

frolovsa (approved):
LGTM (a new acronym i learned last week).

travissluka:
> LGTM (a new acronym i learned last week).

"Let's grab those marbles?"

shlyaeva:
oops and ioda PRs have been merged; this can be merged now.

Issue #429, opened 2020-09-04
test_soca_addincrement dependency


travissluka:
## Description
`test_soca_addincrement` depends on output files from the previous `test_soca_3dvar_godas`, however those same files are overwritten by following ctests. So if `test_soca_addincrement` is rerun separately after doing a complete `ctest` the test will fail. The tests should make sure their output file names are unique.

### To Reproduce:
the following will fail
```
ctest -R test_soca_3dvarfgat
ctest -R test_soca_addincrement
```

## Output
```
47: FATAL from PE     0: fms_io(get_field_id): field uocn NOT found in file ./incr.1.nc
```

travissluka:
this was fixed with https://github.com/JCSDA-internal/soca/pull/485

Issue #422, opened 2020-08-21
u/v not working with checkpoint


travissluka:
## Description
When running checkpoint with `uocn` and `vocn` enabled, `vocn` values are wrong. The pattern is mostly correct, but values are off by a factor of 1e10, and one of the PEs is missing (or that PE is in the only one in the correct range, i forgot to check) see image below

**update:** actually seems to be broken with 3dvar as well, seeing ` => WARNING:  latitute of dest grid exceeds src` in the output
### To Reproduce:
Running on my machine with intel, 8 PEs, 1 degree grid

## Output
![image](https://user-images.githubusercontent.com/12516201/90879861-c9cb4c80-e364-11ea-900a-f3eac79b437a.png)

```
Test     : input background:  
Test     :   Valid time: 2018-04-15T00:00:00Z
Test     :    socn   min=    6.492088   max=   38.000000   mean=   34.439682
Test     :    tocn   min=   -1.900000   max=   33.000000   mean=    8.811991
Test     :     ssh   min=   -1.902829   max=    1.075345   mean=   -0.184206
Test     :    hocn   min=    0.001000   max=  250.031217   mean=   44.566022
Test     :    uocn   min=   -1.148872   max=    1.229909   mean=    0.002323
Test     :    vocn   min=   -1.041607   max=    1.432261   mean=   -0.000246
Test     : analysis: 
Test     :   Valid time: 2018-04-15T00:00:00Z 
Test     :    socn   min=    6.498823   max=   38.014423   mean=   34.440624
Test     :    tocn   min=   -4.961783   max=   33.201462   mean=    8.790240
Test     :     ssh   min=   -1.902829   max=    1.075345   mean=   -0.184206 
Test     :    hocn   min=    0.001000   max=  250.031217   mean=   44.566022 
Test     :    uocn   min=   -1.179814   max=    1.234843   mean=    0.002854 
Test     :    vocn   min=-8297000084.811797   max=9401853709.984081   mean=-977162.795067
```

travissluka:
same thing with `uocn` actually. Fun pattern
![image](https://user-images.githubusercontent.com/12516201/90880915-7b1eb200-e366-11ea-8098-1c38128939ba.png)

PR #465, opened 2020-10-13
Bye bye geostrophy

guillaumevernieres:
## Description
**Issue**
The `u,v` increment was overwritten and computed from geostrophy, which may be OK for the 3dvar, but not so much for the ensemble applications. This was a temporary feature that was supposed to be tested before moving to the balance variable change.

**Solution**
This feature branch removes the geostrophic increment in `add_increment` method of State. The geostrophic module has been removed as well but saved in the `feature/geostrophy` branch to be revisited in the future.

### Issue(s) addressed
Related to #422, #429


guillaumevernieres:
@travissluka , tolerance issues, will fix.

frolovsa:
Good thing i read the description of this PR. Otherwise i was worried that
Earth had stopped rotating!

src/soca/State/soca_state_mod.F90
```diff
@@ -10,7 +10,7 @@ module soca_state_mod
 use soca_increment_mod
 use soca_convert_state_mod
 use oops_variables_mod
-use soca_geostrophy_mod
+!use soca_geostrophy_mod
```
travissluka:
```suggestion
```

travissluka (approved):

PR #473, opened 2020-10-21
cram u/v into letkf state

### LINKED ISSUE #422: u/v not working with checkpoint

travissluka:
## Description

While I had known that LETKF was not updating u/v, it turns out it was worse than this: u/v analysis was being created with zero spread. oops. This PR adds u/v to the LETKF state

- `ucon` and `vocn` were added to `Increment::getpoint()/setpoint()`. Eventually this needs to be cleaned up properly so that the variable names are not hardcoded there. beyond the scope of this PR
- remove the call to `colocate('h')` in `soca_state_add_incr`. This was breaking letkf and ensvariance applications. Those applications require that increment and state dont mess with locations of the increment. Not a problem at the moment since var is not using u/v, but we'll have to implement this correctly at some point. Then again, since colocate was backward, removing it is actually closer to being correct. Beyond the scope of this PR. (closes #422)
- update tests to use u/v. (enspert, ensvariance, letkf)

additional plots will be shown with a subsequent soca-science PR, but looking at LETKF analysis U spread after 14 days.
**before**
image

**after**
image
... small difference


guillaumevernieres (approved):
:+1:

PR #476, opened 2020-10-26
log() and exp() transformations

### LINKED ISSUE #475: Variable change: log() and exp() transformations
## Description
Some ocean variables (e.g. chlorophyll) are always positive and reasonably follow a log-normal distribution on a global-scale. Logarithmic (forward) and exponential (inverse) transformations are thus needed when assimilating these variables in order to 1) normalize the data, and 2) avoid creating negative values after applying the increments.

## Solution
The solution would be to add a new "variable change" option in SOCA to apply logarithmic (forward) and exponential (inverse) transformations to a set of variables that need such transformations. The proposed code change/addition can be found in this [feature branch](https://github.com/JCSDA-internal/soca/tree/feature/ana2model_log). A new ctest would also be needed to test out the applications of this variable change.

liuxiao37k:
## Description

Some ocean variables (e.g. chlorophyll) are always positive and reasonably follow a log-normal distribution on a global-scale. Logarithmic (forward) and exponential (inverse) transformations are thus needed when assimilating these variables in order to 1) normalize the data, and 2) avoid creating negative values after applying the increments.

This PR adds a new "variable change" option within Ana2Model to apply logarithmic (forward) and exponential (inverse) transformations to a set of variables that need such transformations. varchange_ana2model.yml is updated to test this new feature.

### Issue(s) addressed
- fixes #475 

## Testing
- compiled and tested on Hera


src/soca/State/soca_state_mod.F90
```diff
@@ -221,5 +222,48 @@ subroutine soca_state_convert(self, rhs)
...
+      call fckit_log%info("not transforming "//trim(tr_names))
+      cycle
+    end if
+    allocate(trn(size(trocn%val,1),size(trocn%val,2),size(trocn%val,3)))
+    trn = trocn%val
+
+    select case(trim(transfunc))
+    case("log")   ! apply logarithmic transformation
+      do z=1,trocn%nz
+        trocn%val(:,:,z) = log10(max(trn(:,:,z),0.001)) * 2.30258509299 * trocn%mask(:,:)
```
guillaumevernieres: 0.001 seems large?

src/soca/Transforms/LogExpon/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+soca_target_sources(
```
guillaumevernieres: Not needed for this PR

src/soca/Transforms/LogExpon/LogExpon.cc
```diff
@@ -0,0 +1,60 @@
+/*
```
guillaumevernieres: Same here, not needed

src/soca/Transforms/LogExpon/LogExpon.h
```diff
@@ -0,0 +1,56 @@
+/*
```
guillaumevernieres: ... and this file too needs to go.

guillaumevernieres:
@liuxiao37k seems fine again at quick glance. I have a few minor comments mostly about removing left over files from your other work.

src/soca/State/soca_state_mod.F90
liuxiao37k: > 0.001 seems large?

I chose 0.001 to be consistent with the QC-ed observations (globally, chlorophyll mean is around 0.1 and I'd like to set the maximum to be 10 and minimum to be 0.001 such that the log transformed values center around the mean of 0.1). I agree that mathematically it's less accurate but I am worried that having some data points of log(0.0001)=-9.21 will introduce a large bias to the increments, but I guess it also depends on how we define error functions for transformed data.

BTW, do you know how we can avoid having the fortran built-in log() function overwritten by oops:log?

liuxiao37k:
> @liuxiao37k seems fine again at quick glance. I have a few minor comments mostly about removing left over files from your other work.

Oops sorry I forgot to delete the un-needed files. Updated, but some soca tests failed (not ana2model). Seem to be unrelated to the feature branch. Thanks!

src/soca/State/soca_state_mod.F90
guillaumevernieres: If you have a min and a max, you need to use a logit transform, not a log. But if using log, use something like 
x_log = log(x+eps), that should be perfectly invertible, unlike what you currently coded.
The issue is that the tracers that were zero in the model state are now going to be set to a non-zero value, and this is going to be done at least once per DA cycle. I don't think this is what we want to be doing.

guillaumevernieres:
Could you merge develop in your branch @liuxiao37k ? @travissluka fixed the issue yesterday.

src/soca/State/soca_state_mod.F90
liuxiao37k: Thanks. I agree that epsilon is a better option.

src/soca/State/soca_state_mod.F90
```diff
@@ -221,5 +222,48 @@ subroutine soca_state_convert(self, rhs)
...
+      call fckit_log%info("not transforming "//trim(tr_names))
+      cycle
+    end if
+    allocate(trn(size(trocn%val,1),size(trocn%val,2),size(trocn%val,3)))
+    trn = trocn%val
+
+    select case(trim(transfunc))
+    case("log")   ! apply logarithmic transformation
+      do z=1,trocn%nz
+        trocn%val(:,:,z) = log10(trn(:,:,z)+epsilon(trn(:,:,z))) * 2.30258509299 * trocn%mask(:,:)
```
guillaumevernieres: replace by log(trn(:,:,z)+eps) epsilon doesn't need to be an array, and use log, not log10 * 2.3.....

src/soca/State/soca_state_mod.F90
```diff
@@ -221,5 +222,48 @@ subroutine soca_state_convert(self, rhs)
...
+      end do
+    case("expon") ! Apply exponential transformation
+      do z=1,trocn%nz
+        trocn%val(:,:,z) = exp(trn(:,:,z)) * trocn%mask(:,:)
```
guillaumevernieres: that is not the inverse of what's above.

guillaumevernieres (changes requested):

guillaumevernieres:
@liuxiao37k not quite there yet. Your inverse is wrong and replace log10 by log.

src/soca/State/soca_state_mod.F90
guillaumevernieres: > BTW, do you know how we can avoid having the fortran built-in log() function overwritten by oops:log?
Not quite sure what you mean. It's the fortran built in function that is called, nothing to do with oops. What is the issue?

liuxiao37k: Here is the issue: when compiling it reports `This name has not been declared as an array or a function.   [LOG]` because `use fckit_log_module, only: log, fckit_log`. The intrinsic Fortran function log() is overridden.

guillaumevernieres: Ho, I see, sorry, I'm slow @liuxiao37k ... 
`fckit log` is not used, just remove it. `use fckit_log_module, only: fckit_log` will be fine.

src/soca/State/soca_state_mod.F90
```diff
@@ -221,5 +222,44 @@ subroutine soca_state_convert(self, rhs)
...
+      ! Skip if no variable found.
+      call fckit_log%info("not transforming "//trim(tr_names))
+      cycle
+    end if
+    allocate(trn(size(trocn%val,1),size(trocn%val,2),size(trocn%val,3)))
+    trn = trocn%val
+
+    select case(trim(transfunc))
+    case("log")   ! apply logarithmic transformation
+      trocn%val = log10(trn+epsilon(trn)) * 2.30258509299
```
guillaumevernieres: I meant declare a real scalar variable, call it `eps` or something meaningful and set it to a small number ... say 1e-6, we can adjust it later. I don't know what the built in epsilon does, but I don't want to know and I don't want to differentiate it, which will be needed in the near future.

guillaumevernieres (changes requested):
Getting closer, I swear.

src/soca/State/soca_state_mod.F90
liuxiao37k: @guillaumevernieres  Got it! updated. Thanks!

src/soca/State/soca_state_mod.F90
```diff
@@ -221,5 +222,45 @@ subroutine soca_state_convert(self, rhs)
...
+    allocate(trn(size(trocn%val,1),size(trocn%val,2),size(trocn%val,3)))
+    trn = trocn%val
+
+    select case(trim(transfunc))
+    case("log")   ! apply logarithmic transformation
+      trocn%val = log(trn + min_val)
+    case("expon") ! Apply exponential transformation
+      trocn%val = exp(trn) - min_val
+    end select
+    deallocate(trn)
```
guillaumevernieres: `trn` 's scope is only within that routine, no need to deallocate. Not a big deal, we used to this ... just because, but it became tedious to maintain. You can safely remove that line

guillaumevernieres (commented):
... one minor comment. I'll approve after the change.

guillaumevernieres (approved):
Thanks @liuxiao37k

PR #477, opened 2020-10-27
fix failing travisci

### LINKED ISSUE #474: hofx answers have changed
## Description
answers have suddenly changed for hofx in soca
It might be due to https://github.com/JCSDA/oops/pull/4, eventhough i could have sworn that i tested to make sure answers wouldnt change with that pr. (oops!)

kdtree answers are finicky when retrieving N closest points but there happen to be > N points that are equidistant. perhaps calling `reserve()` first changes the way the tree happens to be partitioned at nodes with all equidistant points??

anyway, @danholdaway this might slightly impact your answers as well, if that pr is indeed the cause.

### LINKED ISSUE #478: CRTM binary files are gone, breaking soca travisci.
## Description
TravisCI tests for soca are currently broken because  https://github.com/JCSDA/crtm/pull/9 has removed the binary files we need for our tests.
Either download directly the files we need, or (easier) grab our copy of the binary files we need

travissluka:
## Description
I'm not entirely sure why the answers for hofx (and therefore the 3dvars) changed. Might have been due to #474 . I would have expected all obs to be impacted, not just insitu T/S. Anyway... moving on.

Also, we have our own copy of the CRTM binary files now because these are no longer in the git-lfs of the repository (yay!) but I also don't want to download ALL the coefficients for the 7MB of files that we need.

### Issue(s) addressed

- closes #474
- closes #478 

## Testing

TravisCI tests pass now


travissluka:
ugh, closing to figure out what's going on.

travissluka:
and as a bonus we shaved another 2 minutes off our fastest travisci runtimes, down from 11 min to 9 min!

shlyaeva (approved):
glad you found a quick way to recover from crtm removing the coefficients!

approving part on the crtm coefficients; not entirely sure why the answers have changed..

guillaumevernieres (approved):
:+1:

PR #485, opened 2020-11-11
Writing the increment is now optional

### LINKED ISSUE #451: Add a flag for, or remove, the increment output.
## Description
soca applications producing an increment generate `incr.X.nc` files. This results in a lot of disk space being used for files i'm not using, especially for ensemble applications. enspert/letkf for the 40 member 1 degree MOM6 generates ~30GB of these files for a single cycle

## Solution
- add flag to enable/disable increment output
- or rip it out completely

guillaumevernieres:
## Description
Thanks to @fabiolrdiniz 's https://github.com/JCSDA-internal/oops/pull/955, we can now remove the hard-coded writing of the increment in add_increment.
What was done:
- By default, the increment is not written. Happy  @travissluka ?
- Removed a `save` statement!
- Copy/pasted the increment write example in l95 into soca 3dvar_godas (but I did change the date ... hard work)

The example implemented if for one outer loop only, but the increment can be saved at every outer loop.

### Issue(s) addressed
closes #451

## Testing
What compilers / HPCs was it tested with? **gnu/container**
Are the changes covered by ctests? **added to the 3dvar_godas ctest**

### DIFF
--- a/src/soca/Fields/soca_fields_mod.F90
+++ b/src/soca/Fields/soca_fields_mod.F90
@@ -856,22 +856,6 @@ subroutine soca_fields_read(fld, f_conf, vdate)
     return
   end if
 
-  ! Read diagnostic file
-  if (iread==2) then
-    call f_conf%get_or_die("filename", str)
-    incr_filename = str
-    call fms_io_init()
-    do ii = 1, size(fld%fields)
-      field => fld%fields(ii)
-      if (field%io_name == "" ) cycle
-      if ( field%nz == 1) then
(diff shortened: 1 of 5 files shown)


travissluka (approved):
... so the lesson learned here is that if we wait long enough other people will fix our issues?

guillaumevernieres:
> ... so the lesson learned here is that if we wait long enough other people will fix our issues?

I'm thinking of making a label out of that comment ... The `fakenews` one is pretty handy, but a `if we wait long enough other people will fix this issue` sounds good too. Applies to 50% of the soca issues?

PR #495, opened 2020-11-25
Simplify sea ice io

### LINKED ISSUE #493: Simplify sea ice io
## Description
This is a continuation of #490 and probably should have been done in #490, but, here we are. IO field will be expected to be the same as the sea ice state in soca, no more differentiating between sis2 and cice. Expected field will be ice fraction, thickness and snow depth for 1 category.

## Solution
Delete more stuff.

guillaumevernieres:
## Description
Further simplification of how we handle the sea ice variables in soca.
What was done:
- Removal of io specific to cice5/6 or sis2. `soca` now only reads 2D fields for fraction, thickness and snow depth.
- Sea ice geometry specifications are removed from Geometry

This PR will require changes to `soca-science`

### Issue(s) addressed
closes #493


travissluka:
Not that it matters, but I thought it was funny that the lat/lon on the ice restart file are wrong, making it look like we have Sub-Saharan and Mediterranean ice.
image

travissluka (approved):
👍

PR #509, opened 2021-01-14
fancier correlation length specifications

travissluka:
## Description

The Rossby radius based correlation lengths weren't behaving quite as expected. Assuming we want that radius to be a Gaussian standard deviation, it should have been multiplied by 3.57 to convert it to a Gaspari-Cohn cutoff distance. I took this opportunity to add more options to how the correlation lengths are specified. The new format is as follows:

``` yaml
corr_scales:
  <domain>:
    base value:    # base value of rh (default: 0)
    rossby mult:   # scaling factor for adding the rossby radius on top of "base value" (default: 0)
    min grid mult: # min value of rh, determined by "min grid mult" * grid_size (default: 1)
    min value:     # global min value of rh (default: 0)
    max value:     # global max value of rh (default: something really big)
```

All of the above lengths are given as a Gaussian standard deviation, not GC cutoff. 
These changes should allow us to have ocean correlation lengths that are closer to rossby radius in the extra tropics, ~~and also have some crude equatorial stretching~~.

**don't merge until there is a corresponding PR in soca-science**

## Testing

ctests have been updated to be as close as possible, within rounding error, to what was there before. Unfortunately the slight difference means anything ensemble related is really different, as usual. I'm currently running tests with the 1 deg on soca science using:

``` yaml
corr_scales:
  ocn:
    rossby mult: 1.0
    min grid mult: 1.0
```

I'll post dirac plots once I get a chance to rerun them on discover.


guillaumevernieres (approved):
Nice and clean. Thanks.

PR #519, opened 2021-01-29
add ocean "biomass" as a soca state

### LINKED ISSUE #512: add ocean "biomass" as a soca state
## Description
Previous studies suggest that the assimilation of both chlorophyll and biomass (proxied by particulate organic carbon, or POC) into ocean bigeochem models may yield better results than that of chlorophyll alone. This is very likely to be true for BLING given that chlorophyll is a diagnostic and thus is not connected or feedback to the evaluation of biogeochem states in the model. Here I propose to add ocean "biomass" as a soca state, in the same manner with that of adding chlorophyll, in preparation for the assimilation of satellite-derived POC observations.
Of course, it is important to develop appropriate balance operators between chlorophyll, biomass, nutrients, and other ocean biogeochem variables, but that would be addressed separately at a later stage.

liuxiao37k:
## Description

This PR adds ocean `biop` as a new soca state, following what has been done with `chl`. `biop` is the molar concentration of biomass in sea water in phosphorus units. Phosphorus (instead of nitrogen) is the currency used in BLING.

Also, 3dvar_soca ctest is updated to include the assimilation of `biop`. In this exercise, VIIRS particulate organic carbon (POC) observations are used as a proxy for `biop` after a simple conversion following Redfield ratio of C:P=106:1. Due to its simplicity, this conversion is implemented in ioda-converter, instead of by creating a new ufo. In this test, `Identity` is used as the ufo for `biop`.

A few related obs and model files are updated.

### Issue(s) addressed

- fixes #512 

## Testing

On Hera, with intel18.


guillaumevernieres (commented):
@liuxiao37k looks fine except for the comparison test in one of the 3dvar. You need to build with gcc/debug when you update the reference values.
Looks like a simple edit:
```
Float mismatch at line 40: 17928.009127  not equal to  17928.009125  with max relative difference  1e-12  Actual relative difference =  1.1155732143999785e-10
```
Just swap out the above value and that should be good to go.

guillaumevernieres (approved):
:+1:

PR #528, opened 2021-02-15
Initialize BUMP instance index

benjaminmenetrier:
This is the smallest PR I've ever written: "1 changed file with 1 addition and 1 deletion".
It initialized the BUMP instance index to zero, which is required if the BUMP instrumentation is activated.
Absolutely no impact expected!


guillaumevernieres (approved):
Merci @benjaminmenetrier

PR #533, opened 2021-02-24
add distance_from_coast field

### LINKED ISSUE #532: Add a "distance_from_coast" field so that QC can remove SSS obs near the coastlines
## Description
use kdtree lookup to generate a `distance_from_coast` field during `gridGen`, to be used by the QC later

travissluka:
## Description

`distance_from_coast@GeoVaLs` can now be used in the yamls. gridgen will calculate this for all ocean points when it is run.

image

In addition to removing SSS obs too close to land, this could also be used to do the obs error inflation for insitu profiles near land, which is currently being done in preprocessing by my horrible nest of WOD related scripts

### Issue(s) addressed
- closes #532 

## Testing

tests were updated for 3dvar to use this, instead of the land mask, removing all obs <= 1500km from coast


travissluka:
@kbhargava I'm going to start adding you as a reviewer for soca PRs. Don't feel pressure to approve anything though, but it might be useful to look at these to help you slowly get more familiar with the code

frolovsa:
@travissluka NCODA uses distance to coast to also change the correlation scale (gets shorter and asymmetric as you get closer to the coast). Your image looks very coarse though? Will it look cleaner if one runs at 1/4 resolution?

travissluka:
@frolovsa Yes, it's calculated on the fly for the given grid. This is the 3deg grid we use for the ctests.

guillaumevernieres (approved):
Nice !
:+1:

PR #538, opened 2021-03-10
install of test data

travissluka:
## Description

several of the data files (model `INPUT/`, `godas_sst_bgerr.nc`, and `rossrad.dat`) are now installed if `make install` is done on soca. This is needed for the soca-science TravisCI builds. It should have no impact on anything else


guillaumevernieres (approved):
:+1:

PR #544, opened 2021-03-19
Enable ensemble perturbation and LETKF for chl and biop

liuxiao37k:
## Description

This PR 1) enables ensemble perturbations for chlorophyll and biomass_p (`soca_covariance_mod.F90`), and adds these two variables to enspert and LETKF ctests, and 2) adds background error checkpoints (`soca_bkgerrutil_mod.F90`) for chlorophyll and biomass_p and updates all ctests and their testrefs that call this application.

### Issue(s) addressed

- fixes #537 

## Testing

Hera, intel

PR #548, opened 2021-03-25
defaults for bkgerr min/max

travissluka:
## Description
https://github.com/JCSDA-internal/soca/pull/544 broke soca-science because it added required fields to the yamls for `chl`/`biop` bkg error bounds.

This provides default values for those parameters instead, because I don't think I should have to add yaml parameters for bgc or ice fields if i'm not actually using those fields.

This is a temporary fix. Eventually I will be removing all of these hardcoded field names from this section of code and make it more generic.


guillaumevernieres (approved):
:+1:

liuxiao37k:
@travissluka thanks for fixing it. it sure makes sense!

PR #549, opened 2021-03-31
Fix TravisCI

travissluka:
## Description

Something changed in the docker images that broke our builds.
Something dealing with creating users... i dunno. This fixes it though

TravisCI build should pass with this PR. I'll do the same thing with soca-science later

edit: updated some reference answers as well, not sure if thats due to jedi or container changes... eh, who cares.


guillaumevernieres (approved):
:+1:

PR #551, opened 2021-04-05
Removed GeneralizedDepartures

ytremolet:
## Description

Removed increment's inheritance on GeneralizedDepartures which is not needed and will cause trouble in future PRs.
Part of JCSDA-internal/oops/issues/1137

## Dependencies

None


travissluka (approved):
... I'm always a fan of deleting code. :+1:

PR #553, opened 2021-04-06
Penultimate cleanup for generalization of fields

### LINKED ISSUE #540: implement proper "fast" hofx
## Description

https://github.com/JCSDA-internal/soca/pull/539 implements a quick fix for https://github.com/JCSDA-internal/oops/pull/1098

In order to fix it correctly and switch to the not-slow hofx we need to do some cleaning up, including removing our bad habit of calculating derived fields in `GetValues`.

travissluka:
## Description

Implements the variable change class needed to correctly do hofx3D, and removes almost all hardcoded field information and places it in a yaml configuration file ( see [test/Data/fields_metadata.yml](https://github.com/JCSDA-internal/soca/blob/1bb5bdc68c09de2f78f9b453809af34aa1bae825/test/Data/fields_metadata.yml)).  This should also make adding additional tracers in the future a lot simpler

closes #540 (I can't wait to see how many ms/ns faster our hofx is !!)

**do not merge until appropriate changes are made to soca-science**

### Details
 - Removes all manipulation of field data, other than interpolation, from `GetValues` and `LinearGetValues`. It has been moved to `Model2GeoVaLs` and `LinearModel2GeoVaLs`. `LinearModel2GeoVaLs` simply does a mapping from internal soca variable names to the names expected by UFO, and 2D slicing of the surface of 3D variables. `Model2GeoVaLs` additionally creates several derived variables that had been previously created in `GetValues`
- added `soca_fields_metadata` and `soca_field_metadata` Fortran classes. A single ` soca_fields_metadata`  is created and lives inside `soca_geom`, it is populated by reading in the `fields_metadata.yml` config file. A `soca_field_metadata` is attached to each field of `State` or `Increment` and stores the user definable parameters for the field.
- remove `soca::instantiateBalanceOpFactory();`, it's not really needed. Each variable change class now registers itself inside its own `.cc` file.
- in order for the hack used for CRTM to keep working, I had to add dummy atmospheric fields, which are just 2D fields that aren't actually used for anything. These are at the bottom of `fields_metadata.yml` and should be removed once a proper coupled hofx is implemented

## Testing
There should be no change of answers.

There is a slight bug with the masking on the surface flux variables, this will be fixed in a subsequent PR because it changes the answers.


travissluka:
@liuxiao37k, FYI, I don't know if you were planning on adding any more BGC tracers in the future, but if you do, after this PR you should be able to do so without any code changes. Just add them into  [test/Data/fields_metadata.yml](https://github.com/JCSDA-internal/soca/blob/1bb5bdc68c09de2f78f9b453809af34aa1bae825/test/Data/fields_metadata.yml)

test/testinput/hofx_3d.yml
```diff
@@ -85,28 +86,3 @@ observations:
       simulated variables: [mass_concentration_of_chlorophyll_in_sea_water]
     obs operator:
       name: Chlorophyll Ocean Color
```
guillaumevernieres: @travissluka , why did you remove the filters below?

travissluka: laziness... forgot to put them back after removing them during the initial dev

guillaumevernieres (approved):
:+1:

liuxiao37k:
> @liuxiao37k, FYI, I don't know if you were planning on adding any more BGC tracers in the future, but if you do, after this PR you should be able to do so without any code changes. Just add them into [test/Data/fields_metadata.yml](https://github.com/JCSDA-internal/soca/blob/1bb5bdc68c09de2f78f9b453809af34aa1bae825/test/Data/fields_metadata.yml)

@travissluka This is awesome!!! thank you for adding this capacity!!!

PR #561, opened 2021-04-26
Add the model factory to the enspert application

### LINKED ISSUE #560: Add the possibility of using PseudoModel when randomizing B
## Description
It doesn't really make much sense, ideally we should have a B randomize that is independent of the forecast ... But we don't.

## Solution
Add the model factory in `genpert`

## Alternatives
Create a new application, or change the existing one to not run a forecast.

guillaumevernieres:
## Description
See #560 

### Issue(s) addressed
- closes #560 

## Testing
Not part of the CI testing, do we want to add a test for this?

### DIFF
--- a/src/mains/GenEnsPertB.cc
+++ b/src/mains/GenEnsPertB.cc
@@ -6,11 +6,13 @@
  */
 
 #include "soca/Traits.h"
+#include "oops/generic/instantiateModelFactory.h"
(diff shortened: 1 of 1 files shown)


travissluka (approved):
> Not part of the CI testing, do we want to add a test for this?
Eh, i'll ignore it.

PR #577, opened 2021-05-19
Bugfix for the interpolation of positive definite variables

### LINKED ISSUE #576: Positive definiteness not conserved in interpolation
## Description
Picked up while implementing the dual res 3dvar at 1/4 deg. 3 salinity gridpoint of the sub grid are <0, which the GSW libray doesn't like. Oops 

### To Reproduce:
... don't

guillaumevernieres:
## Description
Stuff's broken for a change. 
I added a `property` field in the metadata for field, it's values can be `none` or `positive_definite` for now. We might extend to `bounded` at some point.

### Issue(s) addressed
- closes #576 
- addresses [soca-science #260](https://github.com/JCSDA-internal/soca-science/issues/260)

## Testing
Unfortunately not picked up by the ctests. The bug appears withing the sub-sampled 1/4 degree grid.


src/soca/State/soca_state_mod.F90
```diff
@@ -166,17 +166,29 @@ subroutine soca_state_convert(self, rhs)
...
   call convert_state%setup(rhs%geom, self%geom, hocn1, hocn2)
   do n = 1, size(rhs%fields)
+    if (rhs%fields(n)%name=='layer_depth') cycle ! skip layer_depth interpolation, faster to re-create
     field1 => rhs%fields(n)
     call self%get(trim(field1%name),field2)
     if (field1%metadata%io_file=="ocn" .or. field1%metadata%io_file=="sfc" .or. field1%metadata%io_file=="ice")  &
     call convert_state%change_resol(field1, field2, rhs%geom, self%geom)
+    ! Insure that positive definite variables are still >0
+    ! TODO (Xiao) add other tracers
+    if (rhs%fields(n)%name=='socn') then
```
travissluka: FYI, I will be rejecting this. 
I've been busy _removing_ unnecessarily hardcoded variable names, so I'm not going to let them start slipping back in. Proper way is to add a `positive definite` variable to the field metadata, with the default set to `false`, and then explicitly setting to true in the yaml file for salinity et al. We could use that flag as well for determining which variables need a log transform if we want that on.

guillaumevernieres: yup, sounds good.

guillaumevernieres:
@travissluka I still have `layer_depth` hard coded in the interpolation. I figured this is fine since we shouldn't be interpolating without `layer_depth` as part of the fields ... Happy to hear suggestions.

travissluka:
> @travissluka I still have `layer_depth` hard coded in the interpolation. I figured this is fine since we shouldn't be interpolating without `layer_depth` as part of the fields ... Happy to hear suggestions.

That's fine, `layer_depth` is special.
I'm not sure I like the name `property`, but it's fine for now and we'll deal with it later if we have other `properties` we want to deal with.

You should also update the soca-science `fields_metadata.yml` before you forget

travissluka (approved):

guillaumevernieres:
> You should also update the soca-science `fields_metadata.yml` before you forget

yes, on it.

PR #578, opened 2021-05-20
Add jedi-cmake to bundle.

### LINKED ISSUE #579: Build soca with static netCDF library.
## Description
This is not necessarily a bug or feature in SOCA.  
SOCA bundle will not build completely when the netcdf libraries it finds are static.
This is due to a limitation in ecbuild's FindNetCDF.cmake
It has been fixed in jedi-cmake.

## Solution
Add jedi-cmake to the bundle.

## Alternatives
Open a PR into ecbuild from jedi-cmake and bring in the FindNetCDF.cmake from there.

aerorahul:
## Description

Add `jedi-cmake` to the bundle

What bug does it fix, or what feature does it add?
makes soca build with static netCDF libraries.

Is a change of answers expected from this PR?
No.

Issues:
closes #579 

## Testing

In CI
Tested on machines with static netcdf libraries.
The changes do not impact containers or stacks built with jedi-stack.
Are the changes covered by ctests? (If not, tests should be added first.) - N/A

### DIFF
--- a/bundle/CMakeLists.txt
+++ b/bundle/CMakeLists.txt
@@ -14,6 +14,8 @@ include( ecbuild_bundle )
 
 ecbuild_bundle_initialize()
 
+ecbuild_bundle( PROJECT jedicmake GIT "https://github.com/jcsda-internal/jedi-cmake.git" BRANCH develop UPDATE )
+include( jedicmake/cmake/Functions/git_functions.cmake  )
 
 # ECMWF repos that you probably dont need to build yourself because they
 #  should be in jedi-stack and the containers
--- a/bundle/CMakeLists.txt
+++ b/bundle/CMakeLists.txt
@@ -14,7 +14,7 @@ include( ecbuild_bundle )
 
 ecbuild_bundle_initialize()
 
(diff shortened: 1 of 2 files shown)


travissluka:
eh, that's not going to work. that include location is only valid if you build everything as a single bundle, which is not how our travisci tests work. also, i think the LIB_REPOS names should match the project name in the bundle, so use `jedicmake`, or change it to `jedi-cmake` in the bundle

bundle/CMakeLists.txt
```diff
@@ -14,6 +14,8 @@ include( ecbuild_bundle )
 
 ecbuild_bundle_initialize()
 
+ecbuild_bundle( PROJECT jedicmake GIT "https://github.com/jcsda-internal/jedi-cmake.git" BRANCH develop UPDATE )
```
travissluka:
```suggestion
ecbuild_bundle( PROJECT jedicmake GIT "https://github.com/jcsda-internal/jedi-cmake.git" UPDATE BRANCH develop  )
```

aerorahul: yup noticed that and done.

travissluka:
I guess it actually worked? I'm assuming cmake will complain if the `include()` path wasn't actually found.

travissluka (approved):

travissluka (approved):

PR #583, opened 2021-05-24
fix soca when ioda-converters is part of the bundle

travissluka:
## Description

`ioda-converters` is messing up soca when it is built as part of the bundle. This is because the file `ioda-converters/src/ncar-bufr2nc-fortran/kinds.f90` generates a `kinds.mod` that conflicts with that same module in oops. Unlike other jedi repos, ioda-converts is not placing its built fortran modules in a project specific output directory (and neither was soca).

This PR makes soca place its compiled fortran modules in a soca specific directory, so that we never have to see those rogue ioda-converter modules. 

Tested on orion: you can now build the `feature/orion_modules` branch without having to specify the `soca-science` target.
Tested on travisCI with https://travis-ci.com/github/JCSDA-internal/soca-science/jobs/506964372


guillaumevernieres (approved):
That works, thanks @travissluka

PR #592, opened 2021-06-15
remove hardcoded masked interp for atmospheric flux fields

travissluka:
## Description

PR #553 cleaned up a lot of the hardcoded variable names in `GetValues`, but one small section for atmospheric fluxes remained because it would cause a change in answers. (There was an inconsistency in masking for these fields: they were masked for interpolation, but not masked for the `gpnorm` prints ).

This removes that exception and forces those variables to use the masked configuration in the metadata file. Answers have changed anywhere the `coolskin` operator was used.


guillaumevernieres (approved):
:+1:

PR #603, opened 2021-07-06
Fixes for tests

ytremolet:
## Description

I needed to add NOTRAPFPE to several tests in soca for them to pass on my machine (on Mac with clang + gfortran). Doesn't seem like  a good idea to ignore divisions by 0 for tests but ocean dwellers are a different breed ;-) Adding Benjamin since it looks like it's happening in BUMP. 

I also needed to explicitly call python3 in two places.

I hope all that works on all platforms (or there is a better solution).

## Dependencies

None.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -450,11 +450,13 @@ soca_add_test( NAME forecast_pseudo
 soca_add_test( NAME static_socaerror_init
                EXE  soca_staticbinit.x
                NOCOMPARE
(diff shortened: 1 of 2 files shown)


guillaumevernieres (approved):
Thanks for the fix, although I can't reproduce the issue you are seeing. Are the saber tests all OK with clang?

ytremolet:
Yes, all the saber tests pass with clang.

PR #605, opened 2021-07-07
Generalize covariance code

### LINKED ISSUE #602: Generalize Covariance class code
## Description
The covariance class uses a lot of hardcoded variable names for
- perturbation scales
- choice of convolution

Generalizing this so that all the information is instead obtained from the yaml configurations will make things easier in the future.

travissluka:
## Description

_**do not merge until soca-science PR is ready**_

Generalizes the background error covariance code to remove hardcoded variable names. All information is obtained from the yaml files.

- all the bump related yaml for `SocaError` covariance has been moved under a `bump:` section to make it cleaner (and consistent with `BUMP` covariance class)
- the `correlation` section is required to inform which variables belong to which correlation lengths generated by socaerror_init.
- As a fun side effect of the generalization, we can have any number of correlation lengths now. So we could setup different lengths for ssh, or various tracers.
- `Covariance` checks to make sure all variables are on the h grid, so u/v are still rejected for now.

I might do a followup PR to generalize the error variance transform, I think it is the last remaining place with hardcoded variable names preventing us from being able to add new tracers simply by changing yaml files.

### Issue(s) addressed

- closes #602

### DIFF
--- a/src/soca/Covariance/soca_covariance_mod.F90
+++ b/src/soca/Covariance/soca_covariance_mod.F90
@@ -31,13 +31,13 @@ public :: soca_cov, soca_cov_setup, soca_cov_delete, &
 
 !> Fortran derived type to hold configuration data for the SOCA background/model covariance
 type :: soca_cov
-   type(bump_type),     pointer :: ocean_conv(:)  !< Ocean convolution op from bump
-   type(bump_type),     pointer :: seaice_conv(:) !< Seaice convolution op from bump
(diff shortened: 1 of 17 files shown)


guillaumevernieres (approved):
Nice! Thanks @travissluka

travissluka:
FYI, i've added back in a line to ignore `hocn` in covariance.
My battle to remove `hocn` as a required field in `Increment`  will continue later.

PR #624, opened 2021-08-12
Feature/compute sqrt vert loc

### LINKED ISSUE #623: generate eigen vectors of vertical localization
## Description
Use new driver in oops to generate eigen vectors of the vertical localization function

frolovsa:
## Description
Adds a driver and the ctest to compute square root of the vertical localization function. This is the first step towards running getkf in soca. 

### Issue(s) addressed
fixes #623 

## Testing
added a new ctests. Examined output and the correlation structure looks reasonable (see image attached). y axis are levels for ssh, T, and S respectively.  x-axis is the correlation between the ob location and the other levels. 

image

## Dependencies
- waiting on JCSDA/oops/pull/#1553


test/testinput/sqrtvertloc.yml
```diff
@@ -0,0 +1,110 @@
...
+truncation tolerence: .7
+
+output:
+ frequency: PT6H
+ datadir: Data
+ exp: sqrtvert 
+ type: ens
+ date: *date
+
+#test:
```
guillaumevernieres: remove commented out lines

guillaumevernieres:
@frolovsa , couldn't you compare to B^1/2  and its sampled reconstruction as a test? That would need to be in oops I guess, impulse response of B and its proxy, at a few places? 
For ex, I wouldn't mind seeing the same plot that you made above but for the original B.

frolovsa:
> @frolovsa , couldn't you compare to B^1/2 and its sampled reconstruction as a test? That would need to be in oops I guess, impulse response of B and its proxy, at a few places?
> For ex, I wouldn't mind seeing the same plot that you made above but for the original B.

I am going to loop @shlyaeva in this discussion as she is starting to review the oops part of this PR. Currently, there is no test driver for sqrtVertLoc. I just make the eigen vectors and compare them to the reference i made before. I assume that an "expert" has looked at the original reference and called them good. 

I can see two ways to accomplish what you are asking for. 
1) I can compute a similar plot for Sqrt(B)*diarcFunction and we can decide that the output is valid and commit the reference solution. I will have to normalize Sqrt(B)*diarcFunction by the 1/sqrt(B.getVariance()). 
2) I can also write a test fixture in oops that will automate the same computation as above. Because the reference solution will be approximate, one would need to set the tolerance value. So an "expert' will have to examine the solution to determine the appropriate tolerance. Also, because we cant disable horizontal correlations at the moment (except we can do it for soca), we will have to compute the approximation at a few locations that are far enough from each other (i.e. we could not just compute some global norm). 

I think option (2) is a valid target. It will take some time to put it together. I might not be able to get to it until mid-September. Should we try to commit this version of the code and then have a follow-up PR to develop a proper test like @guillaumevernieres is suggesting?

frolovsa:
Hm, this does not fail when i run on hera. @travissluka any idea how to debug this further?

travissluka:
> Hm, this does not fail when i run on hera. @travissluka any idea how to debug this further?

The develop branch of soca is failing on travisci because of yaml schema validation stuff from the parameters class code sprint that is going on. I'm probably not going to deal with that until the code sprint is done

frolovsa:
@travissluka i cant reproduce this failure with either gnu or intel on orion. I wonder if this is using an old cashed version of the oops branch feature/computeSqrtVertLoc rather then the oops develop.

travissluka:
@frolovsa it claims its using develop https://app.travis-ci.com/github/JCSDA-internal/soca/builds/241881335#L270
I'll clear the cache and restart the job, on the off chance that works.

travissluka (approved):

guillaumevernieres (commented):
:+1:

PR #633, opened 2021-08-25
[Bugfix] Saving of local geometry

### LINKED ISSUE #632: Local geometry output is wrong
## Description
The array bounds for the local geometry written to file is wrong.

Plotting the pe's domains currently looks like this:
bad-domain-decomp

which isn't correct. It should lokk like this:
good-domain-decomp

guillaumevernieres:
## Description
closes #632


clouden90 (approved):
Looks good to me!

Issue #622, opened 2021-08-03
Implement 3D LETKF (add vertical obs. localization)


frolovsa:
## Description
Add vertical obs. localization to the soca letkf.

## Solution
The following steps will probably need to be added. 
in oops:
[] *GeometryIterator returns point3 object not point2. 
[] Will require changes to all models, l95, qg, fv3, soca, shallow_water, neptune. 

in soca: 
[] Add 3D option to geometry iterator. 
- Probably pass the switch through the geometry object. 
- Add test to soca to test 3D iterator. 

[] Add new method int = GeometryIterator.numberOfDimensions()
[] Modify soca::increment.setLocal() and soca::increment.getLocal() to work with 3D iterator. 
[] Add new obs localization to soca. something like soca::Gc99_3D. 

## Alternatives
GETKF localization is under the development in a parallel issue. It would be nice to have two options that we can compare. Testing in the HAT10 domain will be ideal.

frolovsa:
@travissluka i ran in to a science decision that i need help with. 
So when we iterate in 3D, we loop through all vertical levels for the 3D variables. Each level is assigned a discrete depth value. So far so good. But what happens for the 2D variables? Should one of the iterations be reserved for things like surface  2D variables with the depth assigned to 0? 
So far we have the following 2D variables= {ssh cicen hicen hsnon} all of those are on the surface, correct?

travissluka:
yes, those are all on the surface, and so we could probably give them a depth of 0.0.

we can discuss this in person, perhaps, but it might be a good time to hash this out:
I haven't looked too closely at what you have been implementing, but Instead of thinking in terms of "levels" it might be better to think in more generic terms of "localization groups". Where it is up to the model (or more specific ObsLocalization classes) to determine the number of localization groups, and how obs are localized within those groups. 

E.g. if I want 
1) vertical localization at every level, soca would create 75 "localization groups", one for each level in the column, and modulate the localization depending on dz between grid/obs.
2) to speed things up, only create 2 "localization groups": above MLD, and below. where sst obs only impact the top group, and ssh and insitu T/S impacts both groups.
3) variable localization, expanding on number 2 above, create 3 groups, where s/T are localized independently in the top group, and they both impact the bottom group the same.

[...]

PR #636, opened 2021-09-22
Feature/soca letkf localization

weihuang-jedi:
NOTE: This is a draft PR. It does everything that we need but we might want to iterate on some design choices. 

Introduces 3D iterator in soca
-- *GeometryIterator returns Point3D (lat, lon, dep/alt) instead of Point2D (lat, lon).
-- One can control GeometryIterator through the geometry object. 
-- ++ operation for the iterator will loop either over 2D or 3D points. 
-- added iterator 3d and iterator 2d tests. 
-- also checked that letkf produces exactly the same result with 3D iterator as with 2D iterator. 
-- in the 3D iterator, we have nzo+1 vertical iterations, with surface variables returned together and separatley from the 3D variables. 

Some design considerations: 
i considered different ways of controlling iterator behavior. I think it is cleanest to control it through the config of the geometry object. we often initialize iterator using Geometry.begin() method. Any other method i can think off would be more intrusive or would require a recompile. 

I am not very happy with the nested if statements in the fortran code. Maybe we refactor it when we introduced the iterators following Travis's suggestion on variable groupings. 

Closes:
https://github.com/JCSDA-internal/soca/issues/622

Dependency
- [ ] waiting on JCSDA-internal/oops/pull/1388
- [ ] waiting on JCSDA-internal/ufo/pull/1528


src/soca/GeometryIterator/GeometryIterator.cc
```diff
@@ -72,9 +73,10 @@ GeometryIterator& GeometryIterator::operator++() {
 // -----------------------------------------------------------------------------
 
 void GeometryIterator::print(std::ostream & os) const {
-  double lat, lon;
-  soca_geom_iter_current_f90(keyIter_, lon, lat);
-  os << "GeometryIterator, lat/lon: " << lat << " / " << lon << std::endl;
+  double lat, lon, dep;
+  soca_geom_iter_current_f90(keyIter_, lon, lat, dep);
+  os << "GeometryIterator, lat/lon: " << lat << " / " << lon
+     << " / " << dep << std::endl;
```
frolovsa, line 101: lets call this depth. both in code and in print.

frolovsa (changes requested):
Wei, i think we need to support two cases. One case of a 2D iterator (the default) and the second case of 3D iterator. The best way to trigger this change is to add a property to the geometry yaml and geometry object.

frolovsa:
@weihuang-jedi i cant tag you as a reviewer in this PR. Can you take a look through this code and let me know if you have any feedback?

test/testinput/obslocalization.yml
```diff
@@ -1,6 +1,7 @@
 geometry:
   mom6_input_nml: ./inputnml/input.nml
   fields metadata: ./fields_metadata.yml
+  interator dimension: 2
```
travissluka: "interator" ?

src/soca/GeometryIterator/soca_geom_iter_mod.F90
```diff
@@ -105,8 +108,16 @@ subroutine soca_geom_iter_equals(self, other, equals)
...
-  if (associated(self%geom, other%geom) .and. (self%iind==other%iind) &
-      .and. (self%jind==other%jind)) equals = 1
+  if (associated(self%geom, other%geom)) then
+    if (self%geom%iterator_dimension .eq. 2) then
+      if ((self%iindex==other%iindex) .and. (self%jindex==other%jindex)) equals = 1
+    elseif (self%geom%iterator_dimension .eq. 3) then
+      if ((self%iindex==other%iindex) .and. (self%jindex==other%jindex) .and. &
+          (self%kindex==other%kindex) ) equals = 1
+    else
+      call abor1_ftn('soca_geom_iter_equals: unknown geom%iterator_dimension')
```
guillaumevernieres, line 119: That reminds me that we've been meaning to add an "==" operator in Geometry for a while ... 
That's quite a few nested ifs. You could remove one level by testing the non-association on line 111 and abort. I would suggest to use "case" for the `iterator_dimension` to make the code more readable, cosmetic at this point and up to you @frolovsa and @weihuang-jedi

src/soca/GeometryIterator/soca_geom_iter_mod.F90
```diff
@@ -116,24 +127,52 @@ end subroutine soca_geom_iter_equals
...
+    elseif (self%kindex < 0 .OR. self%kindex > self%geom%kec) then
+      ! out of range
+      call abor1_ftn('soca_geom_iter_current: depth iterator out of bounds')
+    else
+      ! inside of the 3D grid
+      depth = depth1d(1,1,self%kindex)
+    endif
+  else
+    call abor1_ftn('soca_geom_iter_current: unknown geom%iterator_dimension')
   endif
```
guillaumevernieres: Same as above. I think you could reduce the number of nested if's by using cases, it could just be me but I get easily confused when I see too many if's. Cosmetic again, so up to you.

src/soca/Increment/Increment.cc
```diff
@@ -156,26 +158,56 @@ namespace soca {
...
+        else if (vars_[ii] == "ssh")  varlens[ii]=1;
+        else if (vars_[ii] == "cicen") varlens[ii]=1;
+        else if (vars_[ii] == "hicen") varlens[ii]=1;
+        else if (vars_[ii] == "hsnon") varlens[ii]=1;
+        else if (vars_[ii] == "chl") varlens[ii]=nzo;
+        else if (vars_[ii] == "biop") varlens[ii]=nzo;
+        else
+            varlens[ii] = 0;
+      }
     }
```
guillaumevernieres, line 212: Aaaaaa ... Could that be done differently? use "switch/case" maybe?

guillaumevernieres (commented):
@frolovsa ... Lot's of nested if's in your code, any chance you could try to reduce the number of nested levels where you can?

frolovsa:
> @frolovsa ... Lot's of nested if's in your code, any chance you could try to reduce the number of nested levels where you can?

@guillaumevernieres i agree not a big fan of all the nests. I like your suggestion of re-writing as "switch/case". I will give it a go.

src/soca/GeometryIterator/soca_geom_iter_interface.F90
```diff
@@ -0,0 +1,175 @@
+!
```
travissluka: There seems to be a problem with your recent merge of `develop`. This file should be named `soca_geom_iter.interface.F90`

src/soca/Geometry/soca_geom_mod.F90
```diff
@@ -46,27 +46,32 @@ module soca_geom_mod
...
     !> \name local compute domain indices
     !! \{
-    integer :: iscl, iecl, jscl, jecl
+    integer :: iscl, iecl, jscl, jecl, kscl, kecl
     !> \}
 
     !> \name local data domain indices
     !! \{
-    integer :: isdl, iedl, jsdl, jedl
+    integer :: isdl, iedl, jsdl, jedl, ksdl, kedl
```
travissluka: can we get rid of all these `k***` variables and just use `1` and `nco` instead? They're redundant

frolovsa: done

src/soca/GeometryIterator/soca_geom_iter_interface.F90
frolovsa: removed this file.

test/testinput/obslocalization.yml
frolovsa: done

src/soca/GeometryIterator/soca_geom_iter_mod.F90
frolovsa: switch top level ifs to select/case. Hopefully that twill make code more readable.

src/soca/GeometryIterator/soca_geom_iter_mod.F90
frolovsa: switch top level ifs to select/case. Hopefully that twill make code more readable.

src/soca/Increment/Increment.cc
frolovsa: switch top level ifs to select/case. Hopefully that twill make code more readable.

frolovsa:
@guillaumevernieres , @travissluka i followed up on your suggestions. See if the code is up to snuff now.

frolovsa:
@guillaumevernieres @travissluka this is ready to re review. see my comment and request from 6 days ago.

frolovsa (approved):
Not sure if i should be a reviewer on this one since i wrote most of it. But sure, i will approve it.

src/soca/GeometryIterator/GeometryIterator.cc
```diff
@@ -83,13 +84,21 @@ GeometryIterator& GeometryIterator::operator++() {
   soca_geom_iter_next_f90(keyIter_);
   return *this;
 }
+// -----------------------------------------------------------------------------
+
+int GeometryIterator::iteratorDimension() const {
```
travissluka, line 89: i'm waiting for the ctests to pass on my computer, then i'll merge this PR.
But just as a sanity check, the ctests *should* be using `iteratorDimension()`, right? codecov is complaining that it is not being used, but I suspect it's wrong.

travissluka (approved):

YongzuoLi-NOAA:
@travissluka I now can test LETKF vertical localization, right ? If so, I will use insitu T/S profile to test it.

frolovsa:
Not yet.
Still need to add an implementation for the vertical localization. Should
be done in a few weeks.

This PR just allows you loop through 3d points instead of 2d points. but
the result will be the same
------------------------------
Sergey Frolov (he/him)
PSL/OAR/NOAA

YongzuoLi-NOAA:
@frolovsa Thank you. I will wait.

PR #648, opened 2021-10-12
Trigger travis-ci for release/soca-1.0.0

guillaumevernieres:
## Description
What the title says.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ branches:
   only:
     - develop
     - master
+    - release/soca-1.0.0
     - /.*/travisci_.*$/
 
 language: cpp
--- a/bundle/CMakeLists.txt
+++ b/bundle/CMakeLists.txt
@@ -14,7 +14,7 @@ include( ecbuild_bundle )
 
 ecbuild_bundle_initialize()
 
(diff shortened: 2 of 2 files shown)


guillaumevernieres:
Oops ... prep.sh does not handle "TAG" keywords in the bundle.

travissluka:
I thought I fixed that. Maybe I fixed it in the soca-science prep

guillaumevernieres:
> I thought I fixed that. Maybe I fixed it in the soca-science prep

Ha, ok. When using tag we have 1 less field, so ``` awk '{print $8}'``` returns `)` . Cool branch name!
Going to add `UPDATE` in front of tag ... nobody will notice.

guillaumevernieres:
> I thought I fixed that. Maybe I fixed it in the soca-science prep

I don't know what the logic of `UPDATE TAG` would be instead of just `TAG` when checking out a tag ... but **if** it works, I'm not going to argue. I had a mix of both anyways ...

guillaumevernieres:
... 3rd or 4th time in a row that the build stalls and is terminated.
soca-ci

Could you do a mental ctest @travissluka and merge if you don't see anything wrong?

travissluka:
it is possible that a tag could be moved to point to a different commit, so UPDATE kinda makes sense. Though that would clearly be bad style on our part for a release tag.

travissluka (approved):

PR #654, opened 2021-10-20
fix travisci issues

travissluka:
## Description

same as https://github.com/JCSDA-internal/soca/pull/653
keep the build from freezing up on travisCI when it gets near the end


guillaumevernieres (approved):
Oops!
:+1:

PR #665, opened 2021-10-22
Hard code number of pe's used in the tutorial

guillaumevernieres:
## Description
Using all the cores in the node doesn't work when a node has ~20+ core, as the low resolution model used doesn't scale well beyond ~20pes. 

swapping `mpirun` with `mpirun -np 2` in the tutorial scripts.


travissluka (approved):
Travis~CI~ mental ctests are passing

PR #668, opened 2021-10-25
fix yaml schema / and obsloc test

### LINKED ISSUE #667: Remove garbage from yaml files
## Description
some of the ctests are failing because of the yaml schema validation that was added in the recent parameters class code sprint.
Remove the unused garbage from these files so that they work again.

travissluka:
## Description
- garbage removed from the yamls so that the yaml schema checker is happy.
- update the obslocalization test rms reference values because the default search method was changed from `brute force` to `kd tree` and there were tiny differences. remove the search method in the yaml, because the default of `kd tree` is what we want.

### Issue(s) addressed

- fixes #667


guillaumevernieres (approved):
Thanks @travissluka

PR #670, opened 2021-10-26
Empty stubs for the UFS MOM6-CICE6 model

### LINKED ISSUE #669: Empty stubs for the UFS MOM6-CICE6 model
## Description
Preliminary technical work towards interfacing with the UFS MOM6-CICE6 model: Add an empty class for the UFS MOM6-CICE6 model

guillaumevernieres:
## Description
Preparation work for the addition of the UFS MOM6-CICE to the Model factory. 

### What was done
- Empty stubs inside of Model/ufsm6c6/ to define a new model factory member
- Moved and renamed the MOM6 solo model from "SOCA" to MOM6solo 
- Optional build of the the above
- Addition of specific model ctests for the UFS

The added model doesn't do anything, but the implementation should hopefully facilitate the work of adding the UFS to soca.

### Issue(s) addressed
-  fixes #669


src/soca/Model/CMakeLists.txt
```diff
@@ -1,8 +1,4 @@
...
-  ModelFortran.h
-  soca_model_mod.F90
-  soca_model.interface.F90
-  soca_mom6.F90
-)
\ No newline at end of file
+add_subdirectory(mom6solo)
+if (BUILD_UFSM6C6)
+  add_subdirectory(ufsm6c6)
+endif ()
```
aerorahul, line 4: I don't think this is remotely possible since the UFS is not an `ecbuild` project.
Not to mention all the UFS dependencies that will need to be found and checked.

I would suggest an option that effectively does something along the lines of:
```
if(FORECAST_MODEL MATCHES "UFS")
    find_package(ufs REQUIRED)
endif()
```

and replace the option for `BUILD_UFSM6C6` in the top-level `CMakeLists.txt` to:
```
set(FORECAST_MODEL "MOM6Solo" CACHE STRING "Choose Forecast Model options.")
set_property(CACHE FORECAST_MODEL PROPERTY STRINGS "MOM6Solo" "UFS")
```

src/soca/Model/ufsm6c6/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+soca_target_sources(
+  ModelUFSm6c6.cc
+  ModelUFSm6c6.h
```
aerorahul, line 3: I think we should think of these files in the context of what is in fv3-jedi.
In fv3-jedi, they are `ModelUFS.cc` and `ModelUFS.h`.
We should pay a little close attention and see if we can combine these two into a library to interface with or something like that.
It may or may not be possible, so this might be the way to go.  Just thinking out loud.  To be discussed.

aerorahul (commented):
Some preliminary comments to think about.

rsdunlapiv:
@guillaumevernieres I cannot add myself as a reviewer, but I can see the PR

src/soca/Model/ufsm6c6/CMakeLists.txt
rsdunlapiv: That's right.  The "UFS" interface is the same for SOCA and FV3JEDI, at least in terms of the model interface that the OOPS layer sees.  It might be that in the short to medium term, SOCA still interprets "UFS" as having MOM6/CICE6 state while FV3JEDI interprets is as having atmosphere state.  Eventually these distinctions will need to diminish and each system would request whatever state is needed from the coupled model as a whole.

rsdunlapiv: The idea is to extend this interface:
https://github.com/NOAA-EMC/fv3-jedi/tree/develop/src/fv3jedi/Model/ufs
to also bring out the ocean/ice states.  It should be configurable so it can run in SOCA mode without any additional overhead of atmospheric fields being allocated or exported.

rsdunlapiv: We need to figure out where this "UFS" model interface actually lives.  In some sense it does not belong in SOCA or FV3JEDI, but some intermediate place where it would be accessed by both systems.

src/soca/Model/CMakeLists.txt
rsdunlapiv: @mark-a-potts invested a large amount of time in building JEDI with the UFS as a dependency.  We should definitely leverage his experience with this.  Ideally there is one way to build JEDI and link against configurations of the UFS.  (Some of the subcomponents might differ, e.g., I might build UFS with atm only, or datm-ocn-ice, etc., but the umbrella build and linking should be reusable.

src/soca/Model/ufsm6c6/CMakeLists.txt
mark-a-potts: Initially, I was thinking that the model interface would be a base class with SOCA and FV3JEDI being polymorphic child classes that would implement their own specifics, but I am not sure if that will be the right approach for a unified coupling.

src/soca/Model/CMakeLists.txt
mark-a-potts: So, I just re-visited this over the last couple of days and I now have added the UFS as an external project (and dependency) to the ecbuild framework for JEDI. I will try to clean that up and submit it as a PR to JEDI to see if we can get that into the develop branch. Having a single build for the entire system will make this whole integration much, much easier, so I am kine of excited that this seems to be working.

aerorahul: We can let @mark-a-potts take this up when he starts working on this project.

src/soca/Model/ufsm6c6/CMakeLists.txt
aerorahul: When @mark-a-potts starts working, he can provide a path forward.  Will revisit then,

aerorahul (approved):
I resolved my comments.
I reviewed the code visually and this looks promising.

Issue #639, opened 2021-09-29
Fix GetValues tests


travissluka:
## Description
GetValues Interpolation fails because we never bothered to implement an AnalyticInit. It has been temporarily turned off so that the CI passes.

See https://github.com/JCSDA-internal/soca/issues/637

Issue #608, opened 2021-07-12
Use analitical solutions to test GetValue 


guillaumevernieres:
## Description
We are currently reading a state instead of generating an analytical solution. Not ideal.
Thanks @hga007 for pointing out the issue.

## Solution
Implement **AnaliticInit** (See ufo #1332 )[https://github.com/JCSDA-internal/ufo/issues/1332] . See implementation example in https://github.com/JCSDA-internal/shallow-water , and soon in https://github.com/JCSDA-internal/roms-jedi

## Alternatives
Do nothing.

PR #679, opened 2021-11-11
implement analytic init

travissluka:
## Description

- Analytic init class is added to generate an analytic state and analytic geovals.
- GetValues test has been re-enabled

### Issue(s) addressed
- https://github.com/JCSDA-internal/soca/issues/608
- https://github.com/JCSDA-internal/soca/issues/639


travissluka:
@mer-a-o do you know what are we going to do with TravisCI? I see that our plan expired 3 days ago, and so we currently have no way of testing any of the soca repos

PR #692, opened 2021-12-16
removed the read geovals

### LINKED ISSUE #691: Read geoval is "broken"
## Description
I am getting rid of the method that reads the atmospheric geovals in getvalues . Are you OK with that @HamidehGMAO ? That would mean not being able to use the crtm ufo from soca anymore.
@shlyaeva and @ytremolet are close to a 2 model h(x), which is what we will be using in the long term.

guillaumevernieres:
## Description
Removal of read geoval which was needed to use the crtm ufo.

### Issue(s) addressed
fixes #691


travissluka:
there are two references to `crtm` in `.travis.yml` that need to be removed

travissluka (approved):

PR #697, opened 2021-12-22
remove fortran backtrace

travissluka:
## Description

`call backtrace` while useful, is apparently a GNU extension and so doesn't work with intel. Removing


ulmononian (approved):
this builds successfully on orion using machine.orion.intel from soca-science. thanks, travis!

kbhargava (approved):
Tested building soca and soca-science using intel and this branch. Both build fine, some ctests fail but I don't think this PR has anything to do with it. Thanks @travissluka

PR #710, opened 2022-01-31
UFO PR185 broke soca: fix "assign error" in ctest yaml

travissluka:
## Description

https://github.com/JCSDA-internal/ufo/pull/1856 changed the way that `assign error` worked, which broke our answers in our ctests.

I'm assuming that what was happening now is that the error was assigned to adt obs with values between -2.0 and 2.0, but obs outside that range were not being removed (as they were before).
fixed the yaml file to replicate the old behavior

~soca-science should be checked and fixed as well~ (mesoscale error stuff is disabled currently in soca-science)

### DIFF
--- a/test/testinput/3dvar_godas.yml
+++ b/test/testinput/3dvar_godas.yml
@@ -137,6 +137,7 @@ cost function:
     - filter: Bounds Check
       minvalue: -2.0
       maxvalue: 2.0
+    - filter: Perform Action
       action:
         name: assign error
         error function:


gthompsnJCSDA (approved):
Sorry Travis if my PR caused a hiccup to you.  Truthfully the Bounds Check or older Blacklist filters were used for actions like assign long before we had this newer and greater Perform Action contributed by Wojciech.  So any Bounds check that was previously doing a little double-duty of bounding while also assigning error seems wise to update.  Thanks!

PR #725, opened 2022-03-14
fix single ob bug in getvalues, add tests

### LINKED ISSUE #702: run breaks when the number of observations are less than PEs used
## Description
When the number of observations are less than the number of PEs used to run experiments in soca-science the steps run.var/hofx fails. It hangs at the step `GetValues<MODEL, OBS>::fillGeoVaLs`
@travissluka mentioned that it could be because noaa19 has fewer obs than PEs, and something in `fillGeoVaLs` is trying to do an MPI reduce with all PEs but the PEs with 0 obs are skipping the MPI call

### To Reproduce:
Run any DA or noDA experiment for 2015092012 using the sst_noaa19 obs from the R2D2 database located at `/work/noaa/marine/marineda/r2d2/obs/jcsda_soca/ob/benchmark_v1` on orion. The experiment breaks at run.var/hofx step. 

## Output
**output logs**
On orion the log can be seen at `/work/noaa/marine/kritib/runs/HAT10/1135x633x50/test_sstnoaa19/logs/2015092012/run.var  `

travissluka:
## Description
Fixes a bug where GetValues hangs if the number of observations are less than the number of PEs used. This was happening because the initialization of the interpolation requires a global MPI communication, but some of the PEs were exiting early if they had no obs in the obsspace.

I also added two tests, a single ob 3dvar test, and a zero ob 3dvar test. The zero ob test is disable because it causes IODA to crash.... another issue for another day.

### Issue(s) addressed

closes #702

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -3,7 +3,9 @@ set( soca_test_input
   testinput/3dhybfgat.yml
   testinput/3dvarbump.yml
   testinput/3dvar_godas.yml
+  testinput/3dvar_single_ob.yml
   testinput/3dvar_soca.yml
+  testinput/3dvar_zero_ob.yml
   testinput/3dvarlowres_soca.yml
   testinput/3dvarfgat.yml
   testinput/3dvarfgat_pseudo.yml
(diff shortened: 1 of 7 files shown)


guillaumevernieres (approved):
:+1:

test/CMakeLists.txt
```diff
@@ -159,7 +163,9 @@ set( soca_obs
   Data/Obs/gmi_gpm_obs.nc
   Data/Obs/icec.nc
   Data/Obs/icefb.nc
+  Data/Obs/zero_ob.nc
```
kbhargava: Just minor change here, if you can move this one down to follow the alphabetical order you have.

PR #731, opened 2022-03-31
add test for the letkf with halo distribution

### LINKED ISSUE #686: Rossby localization will not work properly with Halo distribution
## Description
I am working on modifying how obs localizations are specified in yamls. I noticed that Rossby localization does not have parameter "lengthscale". This parameter is currently needed to specify how many obs are kept on each PE in a "halo" around the mpi tile. This is done by specifying
`driver: update obs config with geometry info: true`
The update is performed in: 
https://github.com/JCSDA-internal/oops/blob/4315d3d2c3b23cb4c8057857425c5d24be7855ec/src/oops/runs/LocalEnsembleDA.h#L90

## Proposed Solution
I am still working on the new way to specify the lengthscale for Halo distribution. So lets hold tight and see what might work out. Most likely, you will specify something like:
`obs localization: lengthscale for horizontal halo distribution: (in meters)`
This would be a global value that will not change with altitude so most tiles will have to overprovision obs. Would that work, @travissluka ?

## what to do next
I dont think you guys need to do anything at the moment. just dont expect Rossby to work with Halo distribution. GC99 should still work.

travissluka:
## Description

- added the `letkf` ctest to test the use of the letkf _without_ having to do the split observer / solver by using the halo distribution
- the existing ctests were renamed `letkf_split_...`

closes #686

## Testing

There are slight differences in the answers, compared with the split observer/solver, ~but I think this is due to slight interpolation differences due to the way the kdtree is built. Ignoring.~ because thinning will not produce the identical set of observations.

(note that there are large differences shown for the `letkf.test` file... that file was never being used, so ignore the diffs!)


test/testinput/letkf.yml
```diff
@@ -0,0 +1,101 @@
...
+window length: P2D
+
+background:
+  variables: &soca_vars [socn, tocn, ssh, uocn, vocn, hocn, chl, biop]
+  _: &state
+    date: 2018-04-15T00:00:00Z
+    read_from_file: 1
+    basename: ./Data/
+    state variables: *soca_vars
+  members:
```
frolovsa: you could use new templated filenames. see for example. 
https://github.com/JCSDA-internal/oops/blob/a5a30e8edbdf5a1058927f3dbd3f27296be6318d/l95/test/testinput/letkf.yaml#L9

frolovsa (approved):
Cool. I am glad this is working now. 
See a small comment about templated filenames for your consideration.

test/testinput/letkf.yml
travissluka: nice, thanks. I didn't know about that feature

PR #734, opened 2022-04-12
use local copy of UnstructuredInterpolator

travissluka:
## Description

Use a local copy of the unstructured interpolator class that was in OOPS, until such time that the generic interpolator can handle masked/unmasked combinations of variables.

### Issue(s) addressed
- needed for https://github.com/JCSDA-internal/oops/pull/1689

## Testing
all answers are the same


ytremolet (approved):
Looks good. Thanks!

PR #743, opened 2022-05-03
Update test answers for OOPS PR#1709

travissluka:
## Description

- stop using the `compare` script for the test comparison, all tests now use the built in test comparison using the reference file specified in the yaml
- update the reference files to work with pending OOPS PR 

closes #742

NOTE: I still need to update the tolerances.

PR #745, opened 2022-05-05
Addition of the ConvertIncrement  application

### LINKED ISSUE #744: Compute a dynamic height increment from T, S
## Description
This feature would add the possibility of checking the amplitude of the ssh increment resulting from a T, S increment. 

The Jacobian used in the balance operator of the static **B** may have been zero-ed out in some places, thus reducing the amplitude of the associated ssh increment.  

## Solution
Recompute the ssh increment by implementing the ``oops`` ``ConvertIncrement`` application in ``soca``.

guillaumevernieres:
## Description
See issue #744 for more details. This feature branch adds the `ConvertIncrement` `oops` application and implement a test that applies the dynamic height balance to the T, S increment field from the 3dvar_godas test.
The example isn't what we would end up doing exactly, since the ssh field in the input increment would need to be zero-ed out. The output test for ssh is the initial ssh field + the output of the dynamic height calculation. Not what we want but a good enough demonstration of the working plumbing.
 

### Issue(s) addressed

Link the issues to be closed with this PR, whether in this repository, or in another repository.
(Remember, issues should always be created before starting work on a PR branch!)
- fixes #744

## Testing
Unit test and visual inspection of the output increment.

## Dependencies

- [oops PR #1724](https://github.com/JCSDA-internal/oops/pull/1724)

### DIFF
--- a/src/mains/CMakeLists.txt
+++ b/src/mains/CMakeLists.txt
@@ -80,6 +80,12 @@ ecbuild_add_executable( TARGET  soca_letkf.x
                         SOURCES LETKF.cc
                         LIBS    soca
                       )
+
+ecbuild_add_executable( TARGET  soca_convertincrement.x
+                        SOURCES ConvertIncrement.cc
+                        LIBS    soca
+                        )
+
 ecbuild_add_executable( TARGET  soca_convertstate.x
                         SOURCES ConvertState.cc
                         LIBS    soca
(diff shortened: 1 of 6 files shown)


travissluka (approved):
Note to self that the answers will have to be updated if https://github.com/JCSDA-internal/soca/pull/743 is merged in before this PR

PR #747, opened 2022-05-18
re-enable getvalues test

travissluka:
## Description

re-enable the getvalues unit test which was made possible after https://github.com/JCSDA-internal/oops/pull/1731 was merged

the tolerances are slightly smaller now, 1.5 !   (yes, that's huge, I should probably figure out why at some point)

### DIFF
--- a/test/testinput/getvalues.yml
+++ b/test/testinput/getvalues.yml
@@ -2,39 +2,36 @@ geometry:
   mom6_input_nml: ./inputnml/input.nml
   fields metadata: ./fields_metadata.yml
 
-state variables: &geovals [sea_water_potential_temperature,
-                           sea_water_salinity,
-                           sea_water_cell_thickness,
-                           sea_surface_height_above_geoid,
-                           sea_surface_temperature,
-                           sea_ice_category_area_fraction,
-                           sea_ice_category_thickness,
-                           sea_surface_wave_significant_height,
-                           net_downwelling_shortwave_radiation,
-                           upward_latent_heat_flux_in_air,
-                           upward_sensible_heat_flux_in_air,
-                           net_downwelling_longwave_radiation,
-                           friction_velocity_over_water,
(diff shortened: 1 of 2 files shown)


kbhargava (commented):
Builds fine and and all ctests pass except the ones below:
1. test_soca_convertstate
2. test_soca_3dvarbump 
3. test_soca_addincrement 
4. test_soca_convertincrement 
5. test_soca_3dvarlowres_soca 

The above tests fail with `terminate called after throwing an instance of 'oops::TestReferenceFloatMismatchError' what():  Test reference Float mismatch`

test/testinput/getvalues.yml
```diff
@@ -44,3 +41,5 @@ locations:
         lon1: -192.5
         lon2: -57
       obs errors: [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]
+
+tolerance interpolation: 1.5  # still really huge... should probably figure out why sometime.
```
fmahebert, line 45: This test compares an interpolation to an analytic evaluation, so is sensitive to the fact that the "interpolation" algorithm is quite flawed. We can re-evaluate this after my PR fixing the interpolation gets in!

fmahebert (approved):

fmahebert:
@kbhargava I expect the failures you're seeing would be unrelated to this PR, but they may be related to the recent change to output test information with more digits... it's possible some of the test tolerances need adjusting.

test/testinput/getvalues.yml
travissluka: @fmahebert yes, I suspect it's the issue of the interpolation pulling the far away "nearby" points. But on the plus side, here is a great example for you to use when testing the new interpolation!

PR #752, opened 2022-06-03
Feature/parametric b

### LINKED ISSUE #751: Save the parametric background error
## Description
Allow the possibility to save the parametric background error.

## Solution
Use the `convertincrement` applications applied to a unit increment with the appropriate linear variable change.
**What needs to be done:**
- optional generation of a unit increment instead of reading
- remove the saving of the background error in BkgErrFilt and BkgErrGodas (can be obtained by using the `convertincrement` app)

guillaumevernieres:
## Description

This PR enables writing to file the standard deviation of the parametric background error, or your favorite change of variable applied to a unit increment.

Changes:
- Removal of the intermediate "background errors" that were written to hard-coded file names  in `BkgErrFilt` and `BkgErrGodas`. If one wishes to, these can be saved again using the `convertincrement` application and the desiered variable change.
- Added the possibility of generating a unit increment instead of reading from file
- Added a `ctest` that applies `BkgErrFilt` and `BkgErrGodas` to a unit increment and outputs the estimated parametric background error in a filename specified in the yaml configuration

### Issue(s) addressed

- fixes #751 

## Testing

Tested through the added `ctest` inside a container.


guillaumevernieres:
It seems like we could get rid of the case where we set the increment to zero when reading. It doesn't seem to be used anywhere ... and codecov is screaming ...

test/CMakeLists.txt
```diff
@@ -583,6 +585,12 @@ soca_add_test( NAME ensvariance
                EXE  soca_ensvariance.x
                TEST_DEPENDS test_soca_enspert )
 
+soca_add_test( NAME parametric_stddev
+               EXE  soca_convertincrement.x
+               TOL  2e-5 0
```
travissluka: `TOL` is not used anymore, tolerances can be set, if needed, inside the `test` section of the yaml

travissluka (changes requested):
small change, remove `TOL` in the cmake.
Feel free to remove the  `iread=0` section, i dont see anywhere it is being used either.

travissluka (approved):

PR #764, opened 2022-06-27
use localinterp fieldset interface

### LINKED ISSUE #748: [Bug] Interpolation doesn't work
## Description
... And hasn't for a while. I waited for the last oops PR and the updated atlas before writing this ticket:
We are having issues with the 1/4 degree global MOM6 when assimilating more than 1 obs space, it has been this way for quite a few commits, each giving a traceback slightly different than the last. 

### To Reproduce:
You can use my `scratch` here:
```console
/work/noaa/marine/Guillaume.Vernieres/runs/test-interp/scratch-ref
```
I won't erase it for a while, but I might change the `var.yaml`.

1. build the latest `soca` develop 
2. make a copy of `scratch-ref` somewhere
3. `srun -n 480 soca_var.x var.yaml` will crash with the following as an example of traceback:
```console
[...]
OOPS_TRACE LocalUnstructuredInterpolator::LocalUnstructuredInterpolator start
OOPS_TRACE LocalUnstructuredInterpolator::LocalUnstructuredInterpolator done
OOPS_TRACE LocalInterpolator<MODEL>::LocalInterpolator done
OOPS_TRACE LocalInterpolator<MODEL>::LocalInterpolator starting
==== backtrace (tid: 231470) ====
 0 0x000000000024a510 eckit::Value::operator+()  ???:0
 1 0x000000000024b123 eckit::Value::makeOrderedMap()  ???:0
 2 0x00000000000e7507 eckit::LocalConfiguration::LocalConfiguration()  ???:0
 3 0x0000000000629fbc atlas::functionspace::FunctionSpaceImpl::FunctionSpaceImpl()  ???:0
 4 0x000000000061b185 atlas::functionspace::detail::PointCloud::PointCloud()  ???:0
 5 0x000000000061b0e9 atlas::functionspace::PointCloud::PointCloud()  ???:0
 6 0x0000000000198b6d soca::Geometry::Geometry()  /work/noaa/marine/Guillaume.Vernieres/runs/test-interp/soca-science/bundle/soca/src/soca/Geometry/Geometry.cc:52
 7 0x0000000000198b6d std::_Tuple_impl<0ul, atlas::functionspace::PointCloud*, std::default_delete<atlas::functionspace::PointCloud> >::_M_head()  /apps/gcc-10.2.0/gcc-10.2.0/include/c+\
+/10.2.0/tuple:201
 8 0x0000000000198b6d std::__get_helper<0ul, atlas::functionspace::PointCloud*, std::default_delete<atlas::functionspace::PointCloud> >()  /apps/gcc-10.2.0/gcc-10.2.0/include/c++/10.2.0\
/tuple:1284
[...]
```

travissluka:
## Description

Fix the interpolation by not doing 10k+ halo updates.

Short description: the interpolation `apply()` and `applyAD()` needed to take a `FieldSet` instead of `Increment` or `State` so that we aren't continuously converting to/from fieldsets and so that the halo exchange and halo adjoint were not being called so much.

long description:
- uses the changes in pending OOPS PR that allows us to use an interpolator class that takes `FieldSet`
- our `toFieldSet()` and `fromFieldSet()` all operate on **_unmasked_** fields now. Before, SABER was using unmasked fields, and our interpolation was using masked. Since those two interfaces were merged into one recently, I had to change this behavior. To accomplish this, the following were also changed:
  -  The kd tree for lookup of closest points of masked fields was changed to store indexes pointing to gridpoints on the _**unmasked**_ grid.
  - when `soca_gridgen.x` is run, the mask has not yet been created, so `Geometry` constructor would fail when trying to generate the masked kdtree. Instead of a separate `Geometry::gridgen()` method, the constructor now takes a flag indicating if the grid is being generated and kdtree initialization should be performed. (Seems messy, I'll clean that up at some later date.)

### Issue(s) addressed

- fixes #748

## Testing

- ctests pass
- tested on var on  orion with 1/4 degree and normal obs counts

## Dependencies

- requires oops branch of same name
-  https://github.com/JCSDA-internal/oops/pull/1804 needs to be merged first


travissluka:
sigh... and of course I'm getting out-of-memory issues again when testing it this morning :face_exhaling:

liuxiao37k:
> sigh... and of course I'm getting out-of-memory issues again when testing it this morning 😮‍💨

@travissluka Should I plan to test on Gaea, then? There may be other issues with Gaea but I'll try if Orion has been complaining.

travissluka:
@liuxiao37k either should work. If the branches with the temporary fix that @guillaumevernieres has been using work on orion, then these should to. I'm mostly interested in seeing if the runtime and memory usage is different between those two versions

frolovsa:
@travissluka at some point @weihuang-jedi and would want to run this branch with letkf to see if we can come down from the 3 hour runtime mark. Let us know when things start to firm up with this branch so we can give it a go.

travissluka:
@frolovsa this branch should be ready to test. I don't know yet if it solves 100% of our problems, but it should definitely help

travissluka:
I found a small bug in this PR that was causing random failures, I'll take the PR out of draft after I get it fixed

travissluka:
and now I'm back to believeing that this PR works... until someone proves me wrong

liuxiao37k:
> and now I'm back to believeing that this PR works... until someone proves me wrong

@travissluka Are we still good with this PR given the discussion following [oops#1804](https://github.com/JCSDA-internal/oops/pull/1804)? If so, I'll continue testing. Thanks!

shlyaeva:
FYI: no expected changes in oops (lfric-lite was fixed), this PR should be good to go from that point of view.

travissluka:
> > and now I'm back to believeing that this PR works... until someone proves me wrong
> 
> @travissluka Are we still good with this PR given the discussion following [oops#1804](https://github.com/JCSDA-internal/oops/pull/1804)? If so, I'll continue testing. Thanks!

@liuxiao37k yes this should be good to go. Unless anyone finds any major problems, I would like to merge this by the end of the day because it is needed for the skylab release.

src/soca/Geometry/Geometry.cc
```diff
@@ -40,6 +40,14 @@ namespace soca {
     // Fill ATLAS fieldset
     soca_geo_to_fieldset_f90(keyGeom_, extraFields_.get());
 
+    // messy, fix this
+    // generate the grid ONLY if being run under the gridgen application.
+    // also, if true, then don't bother with the kdtree generation in the next step.
+    if (gen) {
```
guillaumevernieres, line 46: that's kind of messy/ugly ... why remove gridgen at all from Geometry?

travissluka: the way the kd tree is implemented in the constructor now, it cannot be built if the masks have not already been created by gridgen, so that section of code was failing for `soca_gridgen.x`. Ideally gridgen should take place in the constructor so that the kdtree after it can work. And since it takes place in the constructor now, there is no need for a separate `Geometry::gridgen()` class method. (you'll notice that the kdtree generate is skipped during gridgen, this is out of laziness because for some reason the mask wasn't being loaded at that point in the code, it doesn't matter for now because we don't need the kdtree in gridgen.)

guillaumevernieres (approved):

liuxiao37k:
@guillaumevernieres @travissluka I tested this soca with an actual 0.25deg case with a full set of obs, and let it run through a full `soca-science` cycle (3DVAR scheme). The `run.var` step took 512 seconds w/ `soca-science` pointing to this soca branch (tested on Gaea) compared to 525 seconds w/ `feature/4emc` (tested on Orion). I didn't compare results, though.

guillaumevernieres:
> [...] I didn't compare results, though.

_I am_ applying the ostrich policy here @liuxiao37k ...

travissluka:
Thank you @liuxiao37k for taking the time to test (especially since orion is down today!)
Glad to see it's 2.5% faster, definitely statistically significant.

liuxiao37k:
> Thank you @liuxiao37k for taking the time to test (especially since orion is down today!)
> Glad to see it's 2.5% faster, definitely statistically significant.

I neglected to mention the reduction from the previous 1129s `run.var` (feature/s2s_dev), so this additional 2.5% improvement is definitely magnificent... I am super excited about a revived develop branch!

PR #766, opened 2022-06-30
update version number and dependencies numbers

travissluka:
## Description

update version numbers for skylab v1.0.0.0.0-rc1

the following upstream still need to be updated when they are updated.
- [ ] saber
- [ ] ioda
- [x] fms ? (probably not)
- [x] mom6 ? (maybe)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# (C) Copyright 2017-2021 UCAR.
+# (C) Copyright 2017-2022 UCAR.
 #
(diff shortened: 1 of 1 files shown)


climbfuji:
Do we need to do all these adventures version number updates? Risk of breaking something if we don't? spack-stack has only those recent versions anyway.

shlyaeva (approved):

travissluka:
... didn't feel like waiting for ioda/saber numbers to be bumped up

PR #769, opened 2022-07-08
The covariance factory is missing in genpert

guillaumevernieres:
## Description
Now that we have discovered that we could split the bump training per variable (it only took 4 years), thanks @benjaminmenetrier!  We are going to be making liberal use of BUMP_NICAS.

`genpert` is  missing the covariance factory and can't use SABER :cry:  , this PR fixes the problem.


travissluka (approved):
TravisCS mental compiler (which is apparently faster than TravisCI) has finished compilation and all tests pass

guillaumevernieres:
> TravisCS mental compiler (which is apparently faster than TravisCI) has finished compilation and all tests pass

Looks like it's stuck after the ctest again.

PR #796, opened 2022-08-04
Make h incr optional

guillaumevernieres:
## Description
What the title says ...

### Issue(s) addressed
- fixes #795 
- fixes #794 

## Testing
Removed the `hocn` increment in :
- `3dvar_soca.yml` , same results.
- `3dvarlowres_soca.yml` ... not the same results as develop ... Needs attention. 

## Dependencies
- waiting on JCSDA/ufo/pull/2190

PR #799, opened 2022-08-22
Speed up and revive soca_covariance_training ctest(s)

### LINKED ISSUE #782: Speed up and revive a few soca_covariance_training ctests
## Description
A few soca_covariance_training ctests have been previously turned off due to slow execution.

## Solution
Try to speed up these tests using coarsened grid and by reducing number of variables. Consolidate the tests if possible.

liuxiao37k:
## Description

This PR adds a bump covariance training ctest back to the list.

1. yaml `parameters_bump_cov.yml` is updated to reflect the `bump:` parameters used [here](https://github.com/JCSDA-internal/soca-science/blob/feature/bump3d-gv/configs/soca/static-b_var-hdiag.yaml.sh);
2. low-res grid is used, and ctest takes about 5-6 seconds;
3. pre-generated pert.lowres.ens.*.an files (~900k, 4 files) are handled via git-lfs.
4. attempted to consolidate existing (turned off) ctests.

For further details please refer to the issue/discussion linked below.

## Issues addressed:
- fixes #782


test/testinput/parameters_bump_cov.yml
```diff
@@ -0,0 +1,64 @@
...
+    parameter: stddev
+    type: an
+
+ensemble:
+  members from template:
+    template:
+      date: *date
+      basename: ./Data/
+      ocn_filename: ocn.pert.lowres.ens.%mem%.2018-04-15T00:00:00Z.PT6H.nc
+      state variables:  [tocn, hocn]
```
liuxiao37k, line 59: @guillaumevernieres after #796 is merged, we won't need `hocn` any more in cases like this, do we? thanks.

guillaumevernieres: @liuxiao37k , depends how you read these files. If reading as a state, `hocn` will still be mandatory, if reading as an increment, you will be able to skip `hocn`. But looking at your yaml, it seems that the yaml key `read_from_file` is missing. A value of `1` reads an increment, `3` reads a state (or the opposite ...). Not quite sure what the default is.

guillaumevernieres (approved):
Looks good. Thanks @liuxiao37k

test/testinput/parameters_bump_cov.yml
liuxiao37k: @guillaumevernieres thank you! looks like the default for `read_from_file` is `1`, per [this piece of code](https://github.com/JCSDA-internal/oops/blob/02d639c574f90241ad8d1131051b499589b086be/qg/model/StateQG.cc#L64), which reads in a state.

ShastriPaturi: You are right @liuxiao37k. I ckecked too.

liuxiao37k: @ShastriPaturi thanks for confirming!

travissluka: The default for `read_from_file` is actually `0`, so @guillaumevernieres is right that it needs to be added
https://github.com/JCSDA-internal/soca/blob/a532e3e55d70ac5b4b3ff57a59bb3654d77d3dc6/src/soca/Fields/soca_fields_mod.F90#L757

https://github.com/JCSDA-internal/soca/blob/a532e3e55d70ac5b4b3ff57a59bb3654d77d3dc6/src/soca/Fields/soca_fields_mod.F90#L775

liuxiao37k: @travissluka Thanks so much for pointing me to the correct source code. I wasn't sure where to look. I am still a bit confused so please help me understand: 1) is this also the place where `read_from_file` key is coded for `ensemble: `? (I do have `read_from_file: 1` for `Background: `) I am confused because I stole `ensemble:` config from a saber [ctest](https://github.com/JCSDA-internal/saber/blob/57ed5b74a57cbe7b930eace0da5a536f47f2304b/test/testinput/qg_error_covariance_training_bump_cov.yaml#L61); I am guessing the interfaces for soca and saber are coded differently for `read_from_file`? 2) How exactly is the state "invented" when `read_from_file` is missing (`0`)? I am not sure I've seen this from soca_fields_mod.F90 but I probably need to look more carefully. I am very curious how we managed to obtain reasonable length scales if the ensemble files are actually not read in correctly (e.g. initialized with zeros??). I was studying @guillaumevernieres's code [here](https://github.com/JCSDA-internal/soca-science/blob/feature/bump3d-gv/configs/soca/static-b_var-hdiag.yaml.sh) and didn't see a second `read_from_file` key beside the one for `Background: `. Thanks again!

guillaumevernieres: @liuxiao37k , what's under template needs to be `soca` specific. Just add `read_from_file: 1` under template and that should be fine ... It doesn't look like `0` or `2` are implemented ... Oops! That chunk of code is so great it probably belongs in a museum.

liuxiao37k: > @liuxiao37k , what's under template needs to be `soca` specific. Just add `read_from_file: 1` under template and that should be fine ... It doesn't look like `0` or `2` are implemented ... Oops! That chunk of code is so great it probably belongs in a museum.

@guillaumevernieres yes sir.

liuxiao37k: @guillaumevernieres I pushed a commit to just add `read_from_file: 1` under `template` without changing testref. Looks like it didn't change the bump output (all ctests passed). A bit odd to me, but as long as the reviewers are happy!

travissluka:
Hmm, odd, yeah I would have expected that to change things. It is possible
the "test" output from bump in this case is meaningless and wouldn't change
.. I'll take a closer look at the output before merging... but otherwise PR
looks fine

guillaumevernieres:
@liuxiao37k , @travissluka , it looks like the `read_from_file` key is set to `1`, so the results are the same. I was going to track down where but I got distracted by a meeting. There's probably an other `read_from_file` passed to the config somewhere ...

liuxiao37k:
> @liuxiao37k , @travissluka , it looks like the `read_from_file` key is set to `1`, so the results are the same. I was going to track down where but I got distracted by a meeting. There's probably an other `read_from_file` passed to the config somewhere ...

Sounds good. thanks! Please let me know when you track it down because I am curious. I was thinking the same yesterday so I started randomly looking into oops code. I also saw some discussion somewhere about `read_from_file` no longer being needed if to read a state, but forgot where. thanks again.

guillaumevernieres:
> > @liuxiao37k , @travissluka , it looks like the `read_from_file` key is set to `1`, so the results are the same. I was going to track down where but I got distracted by a meeting. There's probably an other `read_from_file` passed to the config somewhere ...
> 
> Sounds good. thanks! Please let me know when you track it down because I am curious. I was thinking the same yesterday so I started randomly looking into oops code. I also saw some discussion somewhere about `read_from_file` no longer being needed if to read a state, but forgot where. thanks again.

I probably copied what was done for `qg` in `oops` ages ago, that's where `read_from_file` is coming from. But it is definitely specific to `soca`, not something generic in oops. That part of the code is overdue for a major cleanup!

kbhargava (commented):
Tests pass on Orion. Sorry for the delay, keep getting kicked out of orion.

PR #808, opened 2022-09-16
use composite operator for insituTS in two tests

travissluka:
## Description

Uses the `Composite` obs operator for two of the ctests (added to `hofx_3d`, and replaces existing insitu obs for `3dvar_soca`).
Requires matching UFO branch https://github.com/JCSDA-internal/ufo/pull/2271 

The existing functionality for `MarineVertInterp` and `InsituTemperature` are unaffected. Test answers are the same within rounding error (except for the `Nonlinear Jo`  values which will of course be different.)

We can replace the rest of the tests when we are no longer using the older way of having separate T/S obs operator

## Dependencies
- requires https://github.com/JCSDA-internal/ufo/pull/2271, which can be merged first


guillaumevernieres (approved):
Nice! Thanks for doing this @travissluka

kbhargava (approved):
Builds on orion and ctests pass.

PR #812, opened 2022-09-22
add ice concentation to ewok state variables

travissluka:
## Description

add `cicen` to the state variables in ewok so that https://github.com/JCSDA-internal/ufo/pull/2296 will work


travissluka:
assuming "this can't possibly break anything, and I don't want to wait on anyone" so i'm force merging

PR #816, opened 2022-09-27
bugfix for soca 3dvar under ewok

travissluka:
## Description

- use relative instead of absolute paths for the soca executables. The filenames produced by `soca_var.x` were not correct, and were being truncated, because of the well known and loved "fms has filename lengths that are too short" issue.
- The number of vertical levels for the 5 deg case was wrong (should be 25, it _was_ 33), which seemed to run fine until NaNs appears in the var

## Testing

This PR has been tested under https://github.com/JCSDA-internal/ewok/pull/415

### DIFF
--- a/ewok/static/5deg/MOM_input
+++ b/ewok/static/5deg/MOM_input
@@ -1,3 +1,3 @@
 NIGLOBAL = 72
 NJGLOBAL = 35
-NK = 33
+NK = 25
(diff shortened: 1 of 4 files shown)


travissluka:
note to self: this PR is into `develop`, since nothing else has been merged into `develop`, I will also push this to `release/skylab-v2` afterward

PR #818, opened 2022-10-04
SABER blocks refactoring

benjaminmenetrier:
## Description

Companion PR of https://github.com/JCSDA-internal/saber/pull/327.
The value of nonlinear Jb (and J) might be different because the calculation of the static B inverse changed for SABER blocks.


benjaminmenetrier:
This PR can be merged now.

PR #819, opened 2022-10-06
fix travis CI

travissluka:
## Description

fix TravisCI after the recent docker image update.

on a side note: 
- when did our ctests become so slow??? I remember back when they used to take 1-2 minutes total
- we should probably work on switching over to AWS codebuild, at some point


guillaumevernieres (approved):
ignoring the timeout ...

mer-a-o (approved):

climbfuji:
Any idea why the tests take so much longer nowadays?

kbhargava (approved):

PR #826, opened 2022-12-07
Adapt to ewok changes in passing dates

shlyaeva:
Companion PR for https://github.com/JCSDA-internal/ewok/pull/489

Tested both 3DVar and HofX marine ewok experiments with this branch, first cycle completed for both.

### DIFF
--- a/ewok/tasks/runSaveAnalysis.py
+++ b/ewok/tasks/runSaveAnalysis.py
@@ -20,7 +20,7 @@ if not os.path.exists(workdir):
 os.chdir(workdir)
 
 # Date
-andate = conf['date']
+andate = conf['an']['date']
 base = conf['experiment']['expid'] + '.an' # + andate
 filename = base + '.$(file_type).nc'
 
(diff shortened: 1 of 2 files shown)


ytremolet (approved):
Thanks.

ewok/tasks/runSaveAnalysis.py
```diff
@@ -20,7 +20,7 @@
 os.chdir(workdir)
 
 # Date
-andate = conf['date']
+andate = conf['an']['date']
```
cmgas, line 23: This was conf['an']['datetime'] in fv3, is the difference normal here?

cmgas (commented):
Just a question

ewok/tasks/runSaveAnalysis.py
shlyaeva: Good eye! Yes, this is expected. (I started with `datetime` :))
soca specifies date in the analysis yaml section differently from fv3-jedi (this is model-controlled yaml section):
in fv3-jedi `datetime` key is used: https://github.com/JCSDA-internal/fv3-jedi/blob/develop/ewok/templates/analysis.yaml#L7
in soca `date` key is used: https://github.com/JCSDA-internal/soca/blob/89fbf9bf88e7e66e2de9668b1b00ab1c0c4827a7/ewok/templates/analysis.yaml#L2

cmgas (approved):

PR #829, opened 2023-01-09
Rename plot tasks

cmgas:
## Description
Companion of https://github.com/JCSDA-internal/ewok/pull/535


shlyaeva (approved):
Approving based on consistency with fv3-jedi changes. I didn't test.

cmgas:
This built successfully and the tests all passed, the error is 

```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

travissluka (approved):

ytremolet (approved):

travissluka:
sigh, it looks like someone (@ytremolet) did a normal merge, and not a squash merge.... there goes my clean git commit log

climbfuji:
If it's important, you can restore the branch, check out the hash before the PR was merged, then merge it back in as squash commit ...

ytremolet:
I saw the default was a normal merge and not a squash merge so I left it like that because I thought that's what you wanted. I'll try to remember you also want a clean history!

travissluka:
@climbfuji, just giving @ytremolet a hard time, it's fine.
I guess the default merge method is set per-user, because my default is always squash. I disabled the other merge methods now so it shouldn't happen again.

If I could get SOCA cleaned up enough so that a minor inconsistency in the git log was the messiest thing, that would be a good day!!

climbfuji:
> @climbfuji, just giving @ytremolet a hard time, it's fine. I guess the default merge method is set per-user, because my default is always squash. I disabled the other merge methods now so it shouldn't happen again.
> 
> If I could get SOCA cleaned up enough so that a minor inconsistency in the git log was the messiest thing, that would be a good day!!

We need to be able to do merge commits when we merge release branches into develop, but it's ok to turn it on for just that one merge, then turn it off again.

PR #843, opened 2023-02-15
ctests were failing if run in a different order

travissluka:
## Description
Some of the ctests (e.g. `dirac_soca_cov`, and `enspert`) would fail if you ran them individually after having previously run all the ctests sequentially. This is because two of the tests (`dirac_soca_mask` and `dirac_soca_nomask`) were overwriting the bump files created by the earlier `static_socaerror_init`

This PR
1. modifies the yaml of those dirac tests to NOT overwrite the existing bump files
2. changes the answers for subsequent tests that were using the wrong bump files.

You can now rerun the ctests individually without the answers changing and driving you crazy for a whole day!


kbhargava (approved):
I tested on orion. Build and ran ctests in the order first and then a few randomly. All tests passed

PR #847, opened 2023-02-28
A few fixes to the Model to Analysis variable change

### LINKED ISSUE #842: Co location is not implemented/commented out:
Add a co-locate option to the Ana2Model change of variable.
One might want to rotate, but not co-locate when using the LETKF for ex, but we have to at least colocate to be able to use the variational application.

I could work on this Friday @travissluka if that helps.

guillaumevernieres:
### Description
This was supposed to be straight forward, but then it wasn't.

**Stuff fixed/added:**
- State method added/implemented: `tohgrid` (`tocgrid` doesn't do anything yet)
- Renamed `Ana2Model` to `Model2Ana` to make it less confusing. The variable change under `cost function` is expecting a change of variable from the model to the analysis ... Which we coded backward, or maybe we dropped the inverse switch at some point ... I don't know, but it should be fixed now
- I was hopping to just turn on the interp from uv-points to h-points, but it turned out that thing wasn't working. I added a simple interp based on the 6 neighbors of a point. Simple and a lot faster than a kdtree.

### Notes
- I'm backtracking whatever schematics I added to issue #842 , h(x) should be done on the native grid!
- (u,v) increment on the h-grid and interpolated back to the c-grid then added to the background, or fed to mom6 through iau is fine 
- (u, v) state interpolated from h-grid back to c-grid and directly inserted in the restarts at every cycle is a bad idea (interp=diffusion). 
- I'll add the `tocgrid` method later. We plan to use this work to construct our static-**B** and we only need `tohgrid` to do that work.

### Issue(s) addressed
- fixes #842 

### Testing
- Added the `interp` switch ito the `convertstate_changevar` test 
- Tested at 5deg and 0.25deg.


kbhargava:
Builds fine and the related tests pass on orion. I am going to have a deeper look tomorrow at the files.

kbhargava (changes requested):
Looks good. This is what the interpolation does in this test case, thought will just put it here. 
image

kbhargava (approved):

kbhargava (approved):
I accidentally clicked on "request changes" PR oops!

guillaumevernieres:
Should I just merge this once I resolve the conflicts or do you want to review @travissluka ?

travissluka:
I'll look at it today, I promise

src/soca/Utils/soca_utils.F90
```diff
@@ -257,5 +259,78 @@ subroutine soca_remap_idw(lon_src, lat_src, data_src, lon_dst, lat_dst, data_dst
...
+     ij(1,6) = i-1; ij(2,6) = j-1
+  case default
+     call fckit_exception%abort(fromto, "option is not implemented")
+  end select
+
+end subroutine soca_stencil_neighbors
+
+! ------------------------------------------------------------------------------
+!> idw interpolation given known neighbors
+subroutine soca_stencil_interp(lon_src, lat_src, lon_dst, lat_dst, data, data_out, nn)
```
travissluka, line 296: fine for now, but I have big dreams of someday removing all our interp code and using what is available from oops.... someday

travissluka (approved):

PR #851, opened 2023-03-07
update letkf answers due to ADT UFO fix

travissluka:
## Description

The ADT observation operator was wrong when using halo distribution (see https://github.com/JCSDA-internal/ufo/issues/2692), UFO PR will fix that

ADT was not being tested with LETKF / halo distribution, so ADT obs were added to the LETKF ctests and answers updated. See UFO PR for additional proof that it works.

### Dependencies
Merge after:
* https://github.com/JCSDA-internal/ufo/pull/2695


kbhargava (approved):
This looks fine. Here is a plot of difference between SST analysis after 1 cycle with split solver vs halo. Don't know what the large(ish)  difference at the southern borders are.

image

guillaumevernieres (approved):
:tada:

travissluka:
> This looks fine. Here is a plot of difference between SST analysis after 1 cycle with split solver vs halo. Don't know what the large(ish) difference at the southern borders are.
> 

the kd tree used can be a bit random sometimes. Also, if the halo is set smaller than the largest localization radius, it might remove observations that would otherwise be included in the non-halo distribution, but those obs are far enough away that they don't really matter

PR #857, opened 2023-03-23
fix in getBackground

cmgas:
## Description
Fixes issue for mom6 experiments


fcvdb (approved):
The problem is fixed. Thanks Clementine!

travissluka:
I haven't had a chance to test, so I'll take @fcvdb 's word that it works
THanks @cmgas !

PR #861, opened 2023-04-12
Adjust tolerances

ytremolet:
## Description

I had to adjust tolerance for some tests to pass on a Mac M1 with clang 14.0.3 and gfortran 12.2.0

Is this acceptable? Or should the soca team dig deeper? :-)

## Dependencies

None


travissluka (approved):
We have some tolerances in places that are > 1.0. So I think this is fine!

PR #864, opened 2023-04-17
Merge develop in `feature/ufs_dom`

cmgas:
Update with develop - I couldn't test the build


shlyaeva (approved):
did not test (tried but it doesn't work on the platform/compiler of choice), approving based on commit history -- just merging develop in.

climbfuji:
I'll test it in a bit on my macos where the shared vs static linker issue does not occur (at least it didn't last time!)

climbfuji:
What do you mean with "I don't think this works yet"?

I can compile the NG-GODAS configuration on my macOS, but the tests are failing (as before, so that is expected).

cmgas:
At the time I was getting an error on my Mac, I have not tested since updating with the newest develop

climbfuji (approved):
I tested the NG-GODAS build.

PR #885, opened 2023-05-11
Remove discontinuities in the balance op 

### LINKED ISSUE #883: Fix the spatial discontinuity in the balance operator
## Description
We're zeroing out the jacobians instead of setting them to min/max values.

## Related to
- [JCSDA-internal/soca-science/issues/578](https://github.com/JCSDA-internal/soca-science/issues/578)

### LINKED ISSUE #884: Add an option to use a vanilla version of the vertical convolution
## Description
When I coded that thing eons ago, it was meant to be simple and based on layer thicknesses as a basis for decorrelation length scale. Not great, but good enough as part of a static B. We need that simple feature back in. It's also too easy to trip over yourself while trying to configure the parameters of that change of variable! 

~Also, I think there's a bug somewhere between lines 103-114 of [soca_vertconv_mod.F90](https://github.com/JCSDA-internal/soca/blob/develop/src/soca/LinearVariableChange/VertConv/soca_vertconv_mod.F90).~ We're seeing increment from surface temp larger at depth than at the surface, I assumed VertConv might have been the issue, but it seems to be coming from SocaError ... Which will be deprecated or transformed into something else soon, so ... ignoring the bug.

## Solution
Add a flag to exit `soca_calc_z` [here](https://github.com/JCSDA-internal/soca/blob/8fd638dcdf034bebaf855c0b1fc1e731150d7488/src/soca/LinearVariableChange/VertConv/soca_vertconv_mod.F90#L103).

## Related to
- [soca-science/578](https://github.com/JCSDA-internal/soca-science/issues/578)

guillaumevernieres:
## Description
This is not going to suddenly massively improve our forecast skills or analysis, but the filtering of the jacobians of the linearized balances was not handled properly. 
What this PR does:
- replaces the box filter to a tanh filter when zeroing out jacobians over part of the water column
- the handling of the min/max limits for dS(T)/dT was wrong (set to 0 instead of the limits). This issue is illustrated [here](https://github.com/JCSDA-internal/soca-science/issues/578), thanks to @hyunsookkim-NOAA for the plot.
- limiting the application of dS/dT to below the mixed layer is also done with the same `tanh` filter
- a slight reorganization of the yaml config for the balance operator, it looks something like this now:
```yaml
       - linear variable change name: BalanceSOCA
        kst:
          dsdtmax: 0.1
          dsdzmin: 3.0e-6
          dtdzmin: 1.0e-6
          nlayers: 999
        ksshts:
          nlayers: 10
        dcdt:
          filename: data_static/72x35x25/dcdt.nc
          name: dcdt
```
where `nlayers` is the index at which the filter = 0.5 ... No parameter to configure the slope.
- A "bit" out of scope for this PR, but I also added a `vanilla` option for the vertical convolution. It's meant to by-pass a chunk of configuration that was meant to confuse people with slight dyslexia ... 

### Issue(s) addressed
- fixes #883 
- fixes #884 

## Testing

- ctest/5deg
- 1/4 deg global
- some earlier versions were tested by @hyunsookkim-NOAA  
- intel
- gnu
Are the changes covered by ctests? (If not, tests should be added first.)
- yes

## Dependencies
A change of `yaml` config will be necessary in a few downstream repos (soca-sci, gdasapp, ...)


src/soca/LinearVariableChange/Balance/soca_balance_mod.F90
```diff
@@ -139,27 +118,45 @@ subroutine soca_balance_setup(self, f_conf, traj, geom)
...
              &hocn%val(i,j,:),&
              &self%kst%dsdtmax, self%kst%dsdzmin, self%kst%dtdzmin)
-        ! Set Jacobian to 0 above mixed layer
+
+        ! filter out the Jacobian as specified in the configuration
         do k=1,nl
-           if (layer_depth%val(i,j,k) < mld%val(i,j,1)) then
-              jac(k) = 0.0_kind_Real
-           end if
+           coef_mld = soca_tanh_filt(layer_depth%val(i,j,k),mld%val(i,j,1))
```
travissluka: I know you don't totally agree with how the nemovar balances were implemented, but I believe they set the jacobian to 0 at the surface and linearly interpolate down to the MLD, instead of such a sharp cutoff. Thoughts on trying that (someday?)

src/soca/LinearVariableChange/VertConv/soca_vertconv_mod.F90
```diff
@@ -100,6 +105,9 @@ subroutine soca_calc_lz(self, i, j, lz)
   lz = self%lz_min
   lz = max(lz, self%scale_layer_thick*abs(hocn%val(i,j,:)))
 
+  ! if doing vanilla vertconv, we're done!
+  if (self%vanilla) return
```
travissluka: whats the point of adding this flag? It's the same effect as setting  `Lz_mld: 0` in the yaml

test/testinput/3dvar_godas.yml
```diff
@@ -72,10 +72,13 @@ cost function:
...
       - linear variable change name: BalanceSOCA
-        dsdtmax: 0.1
-        dsdzmin: 3.0e-6
-        dtdzmin: 1.0e-6
-        nlayers: 2
+        kst:
+          dsdtmax: 0.1
+          dsdzmin: 3.0e-6
+          dtdzmin: 1.0e-6
+          nlayers: 999
```
travissluka, line 79: @guillaumevernieres I see several different values of `nlayers` depending on the tests here. What values would you recommend for the real-world config?

guillaumevernieres: @travissluka We need to turn it off for the global until we properly test it and add a high latitude limit. 999 would turn it off. Omitting the kst config in the yaml to turnt kst off might be a better idea.
10 maybe in the hat10?

src/soca/LinearVariableChange/Balance/soca_balance_mod.F90
guillaumevernieres: Sure.

src/soca/LinearVariableChange/VertConv/soca_vertconv_mod.F90
guillaumevernieres: :man_facepalming: 
:ice_cream: ?

travissluka (changes requested):
Thanks for doing this. 2 minor things, but otherwise looks good. (also, would it be worthwhile to add the capability, someday, to output the jacobians so that we can look at them to help with debugging?)

* remove the `vanilla` stuff and use the existing paramters
* instead of cramming in a `999`, make it so that you can omit `kst` or `ksshts` in the yaml to turn off those sections?

guillaumevernieres:
> Thanks for doing this. 2 minor things, but otherwise looks good. (also, would it be worthwhile to add the capability, someday, to output the jacobians so that we can look at them to help with debugging?)
> 
> * remove the `vanilla` stuff and use the existing paramters
> * instead of cramming in a `999`, make it so that you can omit `kst` or `ksshts` in the yaml to turn off those sections?

Will do. Thanks for the review @travissluka .

travissluka (approved):
Thanks, @guillaumevernieres 
Since soca-science CI is currently broken, I'll go ahead and merge this and try to get the associate soca-sci changes merged in later today.

PR #887, opened 2023-05-22
add VADER to SOCA

### LINKED ISSUE #879: Add VADER to SOCA build procces
## Description
Add VADER to SOCA build process, actually hooking it up and using it will be handled later.

* [x] add VADER to bundle / CMakeLists so that it compiles
* [x] make sure VADER is built as part of CI tests

that's it, VADER won't do anything yet

travissluka:
## Description

* closes #879 

Adds VADER to SOCA, and is "used" in the nonlinear variable change, (We don't have any recipes that do anything we would want, so it currently does nothing), but compiling with VADER is now part of the TravisCI.

I had to add the `soca::State::fromFieldSet()` method for this.

## testing
The changes do nothing. No change in ctests answers. We'll actually start making use of VADER at a later date (soonish?)


guillaumevernieres (approved):
:+1:

src/soca/VariableChange/VariableChange.cc
```diff
@@ -22,6 +22,9 @@ namespace soca {
 
 VariableChange::VariableChange(const Parameters_ & params,
                                const Geometry & geometry) {
+  // setup vader
+  vader_.reset(new vader::Vader(params.vader));
```
svahl991, line 26: You might consider passing Vader a defined cookbook when you construct it, maybe just including a single recipe for now. (I haven't tested Vader with an empty cookbook.) That way you control it. Otherwise Vader will use its default cookbook, which gets updated from time to time. But since Vader doesn't currently have any recipes (I don't think) for variables you're producing, it probably doesn't matter much either way right now.

svahl991 (approved):

src/soca/VariableChange/VariableChange.cc
travissluka: @ svahl991 Thanks I thought about that, but I figured would take my chances for now since there are no relevant recipes that could break anything for me. I'll add a proper cookbook when I (soonish?) work on getting some marine var changes into vader.

PR #897, opened 2023-06-07
Just trying to fix the ci, don't merge

guillaumevernieres:


guillaumevernieres:
I thought I was loosing my mind, but nope. Lost it long ago!
`saber` PR merged, so we should be all good.

PR #920, opened 2023-07-13
changes to answers due to BUMP c0 -> si

### LINKED ISSUE #919: Update answers due to saber PR#602
default bump interpolation was changed with 
* https://github.com/JCSDA-internal/saber/pull/602

update the reference answers

travissluka:
## Description
* closes https://github.com/JCSDA-internal/soca/issues/919

bump interpolation default changed from `c0` to `si`.
Since we should be using the new default anyway, I just updated the reference answers. A quick look at the dirac output looks reasonable.


kbhargava (approved):
🎉 thanks Travis!

travissluka:
@guillaumevernieres @kbhargava, I don't think the saber pr will break anything for either the regional or global 1/4, but if you notice bump diracs looking weird you can force bump to use the old `c0` method

PR #922, opened 2023-07-17
Change the mask value

### LINKED ISSUE #915: 3D hybrid fgat with pseudo model is broken
## Description
The 3D hybrid fgat with a pseudo model is the configuration that I use for the `soca` implementation in the global-workflow. I tested with the `3D-FGAT` and `4D-Var` cost function and it worked fine until approximately early last week (~June 20th). I'm not 100% sure of the date, I was working on other unrelated "science" issues when things started failing after rebuilding. I wasted most of my time trying to figure out what was wrong in the yaml config, but I think the issue is in `soca` and/or `oops`.  

### To Reproduce:
I'll push a branch and probably issue a draft PR that will include a test with the EMC configuration. I should have done this earlier! 

**A problem for July 5th and beyond!**

guillaumevernieres:
## Description
This draft PR is only meant to expose a few bugs, it does not fix anything, beyond allowing the EMC's implementation of the hybrid envar to not produce NaN's ... I have yet to check that the results are not sensitive to the mask values.  

Changing the mask value changes the results and a few of the tests are failing for other reasons.

### Issue(s) addressed
- closes #915


guillaumevernieres:
@travissluka : Food for thoughts for this afternoon's meeting. I'll submit an other draft PR that doesn't fix anything but does run a hybrid envar that looks similar to the implementation we have at EMC, and fails.

src/soca/Fields/soca_fields_mod.F90
```diff
@@ -1019,6 +1049,7 @@ subroutine soca_fields_read(self, f_conf, vdate)
       call self%get("mld", mld)
       do i = isc, iec
         do j = jsc, jec
+            if (self%geom%mask2d(i,j)==1) cycle
```
travissluka: @guillaumevernieres btw, this is why nothing was working. The mask is 1 on water, you're putting garbage values or zeros for the mld everywhere!

guillaumevernieres: Ha ... pfff, good catch @travissluka . Not quite sure why it was working before.

travissluka: If I fix that, and put in crazy values for the masked areas, I AM getting floating point exceptions in a lot of places, and the 3dhyb diagnostics are struggling. I'll see what happens if I fix those things

travissluka:
@guillaumevernieres after poking around a bit and setting the masked values to something absurd like 1e33, I've come to the following conclusions:

1. changing `tocn` or `socn` affects the lowres var (the output is the same, but the test answers are different) it seems the mask is not updated correctly when calculating state min/max/mean?
```
	  61: Test Line: '   tocn   min=-1.8106069927198700   max=1000000000000000377920795214610432.0000000000000000   mean=273880909674851778267203825041408.0000000000000000' 
          61: Ref Line : '   tocn   min=-1.9799978062827919   max=32.0274073197037552   mean=6.0264913585459716'  
```
2. changing `ssh` affects `convertstate` and changes the interplated state values of ssh near the coastline. I guess the mom6 interp looks at ssh regardless of if it's masked??
3. changing `uocn` or `vocn` breaks the model forecast, something bad happens at the northern boundary. Also, the `hyb` tests complain because the diagnostics are messed up. Since the hybrid diagnostics start getting messed up, it is conceivable that with NaNs it gets so bad that it breaks completely
4. changing `cicen` changes the ice area everywhere :man_shrugging: 

I'll submit a PR that fixes a few divide by zero issues in a couple of places, and I'll adopt your "setting the mask value" code. I guess we'll just assume that the state changeres has some issues that we should look at if we ever start using it for real, otherwise we'll brush all the other issues under the rug by setting masked values to zero.

guillaumevernieres:
Thanks for looking into this @travissluka .

travissluka:
I pushed a couple of changes to this branch, @guillaumevernieres let me know if it looks okay, and if so I'll merge.
I fixed a couple of bugs found in the balance and mld calculations, they were trying to calculate over land. I also changed the default fill value to 0.0, and removed the explicit fill values from the yaml file. If this fixes your problems can we use it to close #915 ?

guillaumevernieres:
> I pushed a couple of changes to this branch, @guillaumevernieres let me know if it looks okay, and if so I'll merge. I fixed a couple of bugs found in the balance and mld calculations, they were trying to calculate over land. I also changed the default fill value to 0.0, and removed the explicit fill values from the yaml file. If this fixes your problems can we use it to close #915 ?

Thanks @travissluka , the changes look good. I'm going to assume this should still fix our current problem without testing! So merge away when you feel like it.

travissluka (approved):

PR #933, opened 2023-07-26
cleanup soca bundle

### LINKED ISSUE #821: cleanup bundle
## Description
cleanup bundle and ci build similar to
https://github.com/JCSDA-internal/soca-science/pull/524

travissluka:
## Description
* closes https://github.com/JCSDA-internal/soca/issues/821

makes a couple of changes to the bundle that we've been meaning to do, similar to changes that were already done to `soca-science`:
* eckit/fckit/atlas are removed. These are handled by the pre-built spack/modules now. The versions that were listed here were too old to work, which is proof that nobody was building them as part of the bundle anyway.
* the test data for ufo, ioda, saber are optionally included if the `-DTEST_UFO=ON`, `-DTEST_IODA=ON`, `-DTEST_SABER=ON` flags are set when running `ecbuild `
* FMS is off in the bundle by default. It should be loaded as a module now instead `module load fms/release-jcsda`


HamidehGMAO (approved):
Looks good. Thanks

PR #935, opened 2023-07-26
add dummy field support

### LINKED ISSUE #926: Add capability to create dummy atmospheric fields with fixed values.
## Description
We'll temporarily need to add atmospheric fields to SOCA that are needed to start hooking up OASIM. I'd rather not go through the trouble of creating interpolated 2D atmospheric fields.

* [ ] get a list of the atmospheric variables needed, and representative dummy global values
* [ ] add the capability to set these fields to a single global value via the fields metadata yaml

travissluka:
## Description
* closes https://github.com/JCSDA-internal/soca/issues/926

Instead of reading a state variable from a file, we can now add a "dummy" field with a constant value defined in  `fields_metadata.yml` (see `dummy_atm1`). This is useful for making progress on OASIM where we need atmospheric fields, but don't want to take the time to interpolate test atm fields to the MOM6 grid, especially given that coupled hofx will be available. We'll be able to run OASIM tests with standalone SOCA.
(This "constant value field" feature might have other uses in the future that I can't think of at the moment?)

## testing
A dummy field `dummy_atm1` with fixed value of `5.0` has been added to `fields_metadata.yml`.
`hofx_3d` has `dummy_atm1` added to its state, which can be seen in the test reference output as being correctly set to `5.0`


HamidehGMAO (approved):
Thanks Travis! Looks good to me for now. I will add the atmospheric variables like this. Thanks.

PR #942, opened 2023-08-04
Limit the vertical remapping log to 1 pe only.

### LINKED ISSUE #941: Log info fo the vertical remapping should only be done on 1 pe
## Description
Just annoyed by the zillions lines of "vertically remapping stuff".

guillaumevernieres:
## Description
I've had it!!! Working with ensembles and a few 100 pe's makes the log output unreadable. @kbhargava asked me to add a log of what variables were remapped in a previous PR, which is a good idea, but I forgot  to limit the output to 1 pe only. I don't know or remember if there is a better solution, but this feature branch fixes that problem.

### Issue(s) addressed
- fixes #941

### DIFF
--- a/src/soca/Fields/soca_fields_mod.F90
+++ b/src/soca/Fields/soca_fields_mod.F90
@@ -906,7 +906,7 @@ subroutine soca_fields_read(self, f_conf, vdate)
 
     ! built-in variables
     do i=1,size(self%fields)
-      
+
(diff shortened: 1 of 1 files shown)


kbhargava (approved):
Looks good! Thanks for catching this 😄  Builds and test pass on orion. I haven't actually looked at the log files. Below is the screenshot of log files with this branch (on top) and develop (below). 🎉 
image

PR #959, opened 2023-09-25
Updates for updated CI containers

shlyaeva:
## Description

- Updates to CI by Travis
- A non-intelligent reference update that I needed for the soca tests to pass on orion with gnu with spack-stack/1.5.0. Not sure why updates are needed, but all changes are small as far as I can tell.


travissluka:
my untested code change worked on the first try, huzzah!

travissluka (approved):

travissluka:
FYI @shlyaeva the actual reference changes required might have been fewer than you think. I suspect that some answers changed at some point but were still within tolerance so they never got updated.

shlyaeva:
Great @travissluka, thank you for confirming!

PR #967, opened 2023-10-23
Use icepack_FOUND in code CMakeLists

danholdaway:
## Description

Use icepack_FOUND in the code and BUILD_ICEPACK only in the bundle. This prevents other people from having to add BUILD_ICEPACK in addition to putting it in the bundle.

### Issue(s) addressed

N/A

## Testing

Icepack test builds without needing to specify 

## Dependencies

N/A


travissluka:
hmm, the reference answers don't match for the test using icepack.... have we not been running that test this whole time, and @danholdaway fixed it to actually run that test!?!? oops

Dooruk:
This works for me, should I update the test reference for "Travis" to work 😄

Dooruk (approved):

travissluka:
> This works for me, should I update the test reference for "Travis" to work 😄

sure, if you could do that, that would be great.
It seems the other non-TravisCI tests are not compiling with icepack, so i'll have to fix that in a separate PR

danholdaway:
I just pushed the ref change.

guillaumevernieres:
That's weird that the test failed. The test ref was created with gnu and was still OK the last time I checked.

PR #974, opened 2023-11-13
explicit diffusion part 2 - variable groups

### LINKED ISSUE #972: Add support for groups to EXPLICIT_DIFFUSION
Currently, the EXPLICIT_DIFFUSION block only has a single correlation length for all the variables.

We should be able to specify groups.

travissluka:
## Description
- closes #972 

Previously the `EXPLICIT_DIFFUSION` block could only use a single correlation length for all the variables.
This PR adds "variable groups" so that multiple different lengths can be used for different variables.

Don't pay too much attention to the code changes, since some of that will change again once 3D support is added (it was mainly wrapping parts of the existing code in loops to handle groups). Instead, make sure the changes to the yaml input files make sense.

1. Changes to the yamls:
The calibration step now takes a list of groups that have their scales defined (note that this is just a group name, it is not actually associated with any model variables at this point). All the resulting parameters are saved to the same output file.
image
When using, for example in the dirac app, `groups` is required and does a mapping between the group name in the parameters input file (generated in the above step) and the actual variables in the model. If a model variable that the diffusion operator is supposed to process is not in any of the groups, a error will be raised.
image

2. If you look at the test diracs, T  (left) and S are the same as they were before, SSH dirac (right) has been added and uses a smaller length
image


kbhargava (approved):
Looks good. Builds and runs fine on orion. Will test a realistic case after the `part3` PR. Thanks Travis!!

guillaumevernieres (approved):
Thanks @travissluka

PR #984, opened 2023-12-08
disable enshofx

### LINKED ISSUE #981: soca enshofx test bug? each ensemble member tries to write to the same file independently
## Description

I am going down a rabbit hole ...in https://github.com/JCSDA-internal/jedi-bundle/pull/76, I am reporting repeated test failures for only one test, `test_soca_enshofx`. I added print statements to the code that throws the error described in the PR (`FATAL: NETCDF ERROR: Permission denied File=./Vertical_coordinate.nc`):
```
vi ../../../../../jedi-bundle-no-gfdl-forks/mom6/src/framework/MOM_io.F90
# 
...
  write(0,*) "CREATING FILE " // trim(filename) // " with: ", domain_set, num_PEs(), thread, SINGLE_FILE, one_file
  if (one_file) then
    call open_file(IO_handle, filename, OVERWRITE_FILE, threading=thread)
  else
    call open_file(IO_handle, filename, OVERWRITE_FILE, MOM_domain=Domain)
  endif
  write(0,*) "CREATED FILE " // trim(filename)
...
#
I_MPI_FABRICS="shm" /opt/intel/mpi/2021.6.0/bin/mpiexec "-n" "6" "/mnt/experiments-efs/dom.heinzeller/jedi-bundle-fms-2023p03/build-no-gfdl-forks-debug/bin/soca_enshofx.x" "testinput/enshofx.yml"
...
 CREATING FILE ./Vertical_coordinate with:  F           1         400
         400 T
 CREATING FILE ./Vertical_coordinate with:  F           1         400
         400 T
 CREATING FILE ./Vertical_coordinate with:  F           1         400
         400 T
 CREATED FILE ./Vertical_coordinate
 FATAL: NETCDF ERROR: Permission denied File=./Vertical_coordinate.nc
```
As we can see, multiple processes (each member) is trying to create and write to the file, but since it's only one PE per ensemble member it does so serially. If I set the environment variable `HDF5_USE_FILE_LOCKING=FALSE` before running the test, then the test passes.

My question to @travissluka @shlyaeva @ytremolet is: I think we sometimes set this env variable (e.g. https://github.com/JCSDA-internal/jedi-tools/blob/develop/buildscripts/compiling/run_intel_build_S4.sh where it doesn't make sense, because this is for a compile-jedi jobcard), but so far we haven't had to set it on aws-pcluster. It makes sense to disable file locking if we can guarantee that we only read from the same file with multiple independent processes, but writing to it seems rather dangerous?

### To Reproduce:

See above

## Additional context
n/a

## Output

See above

travissluka:
## Description
- closes https://github.com/JCSDA-internal/soca/issues/981

Removes the enshofx ctest because it causes problems. We don't use enshofx in the real world anyway

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -645,12 +645,18 @@ soca_add_test( NAME hofx_4d_pseudo
                TEST_DEPENDS test_soca_gridgen
                             test_soca_forecast_mom6 )
 
-  # TODO Re-enable the comparison once that's fixed
-soca_add_test( NAME enshofx
-               EXE  soca_enshofx.x
-               MPI 3
-               NOTRAPFPE
-               TEST_DEPENDS test_soca_gridgen )
+# NOTE, enshofx is disabled because MOM6 always writes out some diagnostic
+# netcdf files (which can't be disabled). This is normally not a problem, except
+# with enshofx, because multiple instances of MOM6 are started under separate
+# MPI_COMMs, causing problems when they all try to write to the same file at the
+# same time. We'll need to take a closer look at MOM6 output when we start
+# dealing with in-core MOM6
+
+# soca_add_test( NAME enshofx
+#                EXE  soca_enshofx.x
+#                MPI 3
+#                NOTRAPFPE
+#                TEST_DEPENDS test_soca_gridgen )
 
 
 # variational methods


climbfuji (approved):
Thanks!

kbhargava (approved):
Looks god to me. Haven't tried building though.

PR #988, opened 2023-12-18
Add vertical component to explicit diffusion central block

### LINKED ISSUE #971: Add 3D support to EXPLICIT_DIFFUSION
Add an interleaved 2D hz + 1D vt mode to the explicit diffusion. Vertical and horizontal normalization is to be calculated separately. This should allow for:

- horizontal normalization can be calculated offline once (slowly)
- vertical normalization can be calculated every cycle if desired, it should be faster

Bonus points:
- test as a localization operator for EnVAR

travissluka:
## Description

- closes #971 

Add vertical diffusion to `EXPLICIT_DIFFUSION`. It should now have everything needed to be used in place of BUMP for the correlation operator.

**details:**
- follows the Weaver papers, look those up if you're curious.
- This has been implemented by splitting the horizontal and vertical steps as $C=C_V^{1/2} C_h^{1/2} C_h^{T/2} C_V^{T/2}$ This allows us to calculate the normalization separately. So, for example, horizontal normalization (which is slower) can be calculated once, and vertical (which is very fast) can be calculated more frequently (monthly, every cycle, etc)
- with the vertical and the horizontal separated, the normalization is no longer perfect. It's close enough, but just note that even with a `brute force` normalization method there are tiny errors in the normalization. This is to be expected.
- see the `parameters_diffusion_hz.yaml` and `parameters_diffusion_vt.yaml` files for examples of how to initialize the diffusion block.
- a script has been added to `tools/calc_scales.py` that I used for calculating the horizontal and vertical correlation input files for the HAT10 test run. I'll clean these up later, but they are useable now.

## Testing
- the adjoint test works, with the same caveat as last time that a halo update needs to be commented out in the code for the test to pass (but it can't be left commented out for the rest of the tests to work)
- dirac tests have been updated to do 3D, they look correct in the vertical, see for yourself.
- a 3dvar test was added that uses `EXPLICIT_DIFFUSION` instead of `SocaError`. The results look similar but are slightly different because the vertical scales are different.

I have run 2 months with the hat10, and while the vertical correlation lengths need tuning (they're probably too big), things seem better. For example, off the coast of Brasil with `VertConv / BUMP` (left) there was a horrible warm blob that is gone with `EXPLICIT_DIFFUSION` (right). (I think this was caused by bugs with `VertConv` when the thin 1 mm levels were present, I don't feel like elaborating on what I think was wrong)
image

Looking at the vertical impact with the HAT10 regional: I used the following MLD (calculated from a July mean) to specify the vertical length scales (using the `calc_scales.py` script)
image

assimilating SSt and ADT on the first cycle gives the following T increment at 1m (left) and a little over 20m depth(right), showing the impact of specifying MLD on the vertical lengths
image

### Timing
Using the 1/4 deg regional HAT10, with correlation lengths of roughly 1 rossby, to check timing:

**Using BUMP/VertConv**
```
Name                              : max (ms)    avg (ms)     % total   imbal (%)
SocaError::inverseMultiply        :   150.39      141.63        0.16       12.52
SocaError::multiply               : 94575.19    94410.34       85.79        0.39
ErrorCovariance::doMultiply       : 75164.96    66793.77       60.70       19.63
LinearVariableChange::changeVarAD : 16549.46    11832.66       10.75      111.92
LinearVariableChange::changeVarTL : 15941.01    15795.20       14.35        1.62
Timers::Total                     :110072.44   110042.78      100.00        0.06
```
And with EXPLICIT_DIFFUSION for 3D diffusion
```
Name                              : max (ms)    avg (ms)     % total   imbal (%)
SABER::inverseMultiply            :  8550.12     8538.93       12.65        0.38
SABER::multiply                   : 39020.70    38707.42       57.33        2.00
LinearVariableChange::changeVarAD :  1783.66     1493.37        2.21       42.24
LinearVariableChange::changeVarTL :  1538.94     1200.01        1.78       61.97
Timers::Total                     : 67543.66    67513.31      100.00        0.10
```

quite a bit faster. Note that SABER with `EXPLICIT_DIFFUSION` is doing a `GMRESR` for the inverse multiply, whereas `SocaError` was doing nothing, so that unfairly adds 8 seconds of runtime! As you get away from the large scales required at the equator, I would expect `EXPLICIT_DIFFUSION` to outperform `BUMP_NICAS` considerably more.

##  next steps
- It "works" for localization, almost, but I need to add in a cross-variable part before it can be used for ensemble localization
- I might add in the capability to do a single 3D diffusion (instead of the implemented 2D + 1D) if we cared about the operator being perfectly normalized for localization.
- Several components can be deprecated/removed once downstream stuff is tested and switched over: `VertConvSOCA`, `SocaError`, `BkgErrSOCA`, `HorizFiltSOCA`
- It might run faster if we use a larger halo and do a halo exchange less frequently, maybe. The horizontal required 206 iterations of diffusion, and the vertical required 288, with a halo exchange each time.


kbhargava (approved):
Thank @travissluka !! Looks fantastic and all the tests pass with GNU on orion.

guillaumevernieres (approved):
:tada:

Dooruk:
@travissluka thank you for this feature, a great Xmas present 🎁  I'm excited to test/cycle with it!

In terms of the correlation scales and files created, you wrote in the previous diffusion [PR](https://github.com/JCSDA-internal/soca/pull/966) that vertical correlation (parameters_diffusion_vt) needs to be regenerated every so often (weekly, monthly?) and the horizontal correlation scale files would be a one time initialization, correct? Asking this so I can better implement this into our workflow.

Is [this](https://rmets.onlinelibrary.wiley.com/doi/full/10.1002/qj.3918) a good paper to have a general understanding of the methodology you implemented here or would you recommend reading other Weaver papers?

travissluka:
> @travissluka thank you for this feature, a great Xmas present 🎁 I'm excited to test/cycle with it!
> 
> In terms of the correlation scales and files created, you wrote in the previous diffusion [PR](https://github.com/JCSDA-internal/soca/pull/966) that vertical correlation (parameters_diffusion_vt) needs to be regenerated every so often (weekly, monthly?) and the horizontal correlation scale files would be a one time initialization, correct? Asking this so I can better implement this into our workflow.
> 
> Is [this](https://rmets.onlinelibrary.wiley.com/doi/full/10.1002/qj.3918) a good paper to have a general understanding of the methodology you implemented here or would you recommend reading other Weaver papers?

@Dooruk That's right, horizontal calibration only needs to be performed once, and the vertical will need to be done more often. I would plan on doing the vertical every cycle, and either using the instantaneous MLD from the restart files or performing some sort of running average. The `calc_scales.py` script I have in soca can be cleaned up a bit for generating those scales. I'll be issuing a PR soon-ish to create new static files on the HPCs for the 1/4 deg grid, so you can pull the horizontal parameters file from there.

Yes, that's the main paper I looked at. I used the 2D+1D method from that paper (without interleaving, for now, I might try interleaving someday to see if it makes a difference). Weaver's 2013 and 2001 papers are also useful. Just note that Weaver focuses mainly on implicit diffusion, but I've implemented explicit diffusion. Which is slower, but good enough for now.

Dooruk:
Sounds good, thanks. Just to reiterate, you are suggesting using `calc_scales.py` (well, a modified version of it) instead of `parameters_diffusion_vt` and stick with `parameters_diffusion_hz`, until you create the static files for 0.25 degrees, got it.

travissluka:
> Sounds good, thanks. Just to reiterate, you are suggesting using `calc_scales.py` (well, a modified version of it) instead of `parameters_diffusion_vt` and stick with `parameters_diffusion_hz`, until you create the static files for 0.25 degrees, got it.

@Dooruk there should now be files in `/discover/nobackup/projects/jcsda/s2127/static/skylab-7.0.0/soca/025deg/diffusion/` that you can use. The horizontal files and the vertical files for our test month are there. You can use those files as is to make sure things are working (I actually haven't taken a scientific look at the 1/4 deg global, so I'm eager to see how it works for you). You can see here https://github.com/JCSDA-internal/skylab/blob/develop/models/mom6/defaults/bstatic.yaml how i have the static b yaml setup for the skylab runs

Then, in your workflow, yes use `calc_scales.py` using the MLD of the day as input, then recalculate the vertical parameters with `soca_error_covariance_toolbox.x` (like is done for the `parameters_diffusion_vt` soca ctest)

Dooruk:
@travissluka thank you for providing those files. I'm able to use them with our workflow. I will try `calc_scales.py` next to obtain dynamic vertical parameters and start cycling. Here is the convergence plots with the observations below:

```
- adt_3a_egm2008
- adt_3b_egm2008
- adt_c2_egm2008
- adt_j3_egm2008
- adt_sa_egm2008
- salt_profile_fnmoc
- sst_ostia (active for this plot but passive while cycling)
- sss_smos_esa
- sss_trak_fnmoc
- sst_gmi_l3u (passive, but this is what i use while cycling)
- sst_ship_fnmoc
- sst_trak_fnmoc
- temp_profile_fnmoc
```

norm_gradient_reduction

One thing I noticed is that iteration part takes about 4-5 times longer when I use full set of the observations. I'm posting the `OOPS_STATS` outputs below, the only difference between the two is profilers and `background_error_model`. I'm currently trying another test with no profilers but it is still taking as long.

`BUMP` Stats:

Screenshot 2024-01-08 at 11 16 33 AM

`EXPLICIT_DIFFUSION` Stats:

Screenshot 2024-01-08 at 11 15 58 AM

travissluka:
@Dooruk hmm, that's weird. The difference in timing seems to be in the interpolation, and I don't think it should be taking that long so there is probably something wrong. I don't think it has anything to do with the static B settings.
If you're still having problems after spending some time diagnosing, please open a separate issue and we'll discuss there.

PR #1001, opened 2024-01-12
Enable use of EXPLICIT_DIFFUSION for localization

### LINKED ISSUE #998: Add ability to use EXPLICIT_DIFFUSION for localization
## Description
To use `EXPLICIT_DIFFUSION` for localization, the following need to be added

- an equivalent to the saber `multivariate strategy: duplicated` mode
- an equivalent to the duplication strategy for the vertical, in case we want to turn off vertical localization. Normally with `BUMP_NICAS` we would just set the vertical scale to something large, but this won't work with Diffusion because it hates large scales.

travissluka:
## Description
- closes #998 
- closes https://github.com/JCSDA-internal/soca-science/issues/615
- closes https://github.com/JCSDA-internal/soca-science/issues/616
- 
Several features were added to the EXPLICIT_DIFFUSION saber central block to allow us to use it for localization
- `multivariate strategy: duplicated`
  similar to BUMP_NICAS, setting this parameter (the default behavior is `univariate`) duplicates the multiplication across all the variables in a group. This is essentially done by first creating a new 3d field that is a summation of the input variables, doing the diffusion multiply, then copying the single field back into the input variables. 
- `vertical.strategy: duplicated`
  normally with BUMP_NICAS we would just set a large vertical length scale if we didn't want localization in the vertical. But, diffusion hates large scales, so this copies the "duplicated variable" strategy above, but for vertical levels. This essentially creates a 2d field that is a summation of the input levels of the 3d field, performs a horizontal diffusion, then copies back into the 3d field. In effect, this is the same as an infinitely long vertical length scale... but much faster.
- a new `3dhyb_diffusion` test was added, see the yaml for examples of how to use the new parameters

## Testing
- [x] Travis: test with the HAT10 before merging
- [x] ctest added.
- [x] previous ctests using diffusion for correlation are not affected

If you turn on those features when doing a dirac, you'll see the cross variable impact (not shown here)

Comparing the 3dhyb test with diffusion (right) with the original nicas (left), and setting the weights in the test to only use the ensemble portion of B, the increments look nearly identical

image
image
image
image


travissluka:
and testing with 1/4 deg HAT10. Increments after 1 cycle for NICAS (left) and diffusion (right). 3denvar with a 10 members, pure ensemble B. Lengthscale is set to
- rossby mult 1.5
- min grid mult = 1.0
- hz max = 250e3

T at lvl 20
image

U at lvl 20
image

SSH
image

They look pretty much the same to me!

### timing
The only difference in timing seems to be the initialization of the ensemble covariance, I think it was loading the states slowly for the NICAS run (orion just being orion?)

with NICAS (resolution set to 8.0)
```
OOPS_STATS Name                                                :     min (ms)    max (ms)    avg (ms)     % total   imbal (%) 
OOPS_STATS oops::Covariance::EnsembleCovariance                :     12195.86    12197.89    12197.06       38.77        0.02
OOPS_STATS oops::Covariance::HybridCovariance                  :     12207.50    12208.92    12208.08       38.80        0.01
OOPS_STATS oops::Covariance::ensemble::multiply                :      8139.09     8362.95     8256.86       26.24        2.71
OOPS_STATS oops::Covariance::hybrid::inverseMultiply           :      2214.91     2222.57     2219.32        7.05        0.35
OOPS_STATS oops::Covariance::hybrid::multiply                  :      8240.88     8444.88     8345.76       26.53        2.44
OOPS_STATS oops::Localization::Localization                    :      1217.60     1304.02     1239.01        3.94        6.98
OOPS_STATS oops::Localization::multiply                        :      7111.51     7452.73     7242.06       23.02        4.71
OOPS_STATS util::Timers::Total                                 :     31461.57    31464.64    31462.75      100.00        0.01
```

with diffusion
```
OOPS_STATS Name                                                :     min (ms)    max (ms)    avg (ms)     % total   imbal (%) 
OOPS_STATS oops::Covariance::EnsembleCovariance                :       605.42      608.18      607.30        3.70        0.45
OOPS_STATS oops::Covariance::HybridCovariance                  :       616.23      619.13      618.28        3.77        0.47
OOPS_STATS oops::Covariance::ensemble::multiply                :      7361.03     7514.55     7440.37       45.35        2.06
OOPS_STATS oops::Covariance::hybrid::inverseMultiply           :      1975.60     1980.99     1978.07       12.06        0.27
OOPS_STATS oops::Covariance::hybrid::multiply                  :      7452.55     7590.71     7523.99       45.86        1.84
OOPS_STATS oops::Localization::Localization                    :       183.05      202.75      192.12        1.17       10.26
OOPS_STATS oops::Localization::multiply                        :      6166.99     6619.16     6412.32       39.08        7.05
OOPS_STATS util::Timers::Total                                 :     16405.69    16408.42    16406.98      100.00        0.02
```

shlyaeva:
In the future we could implement `multivariate duplicated` vs `univariate` up a level in saber block chains (similar to what we do now for cross-time localization).

travissluka:
> In the future we could implement `multivariate duplicated` vs `univariate` up a level in saber block chains (similar to what we do now for cross-time localization).

yep, I was thinking about that... but for now the amount of code required was minimal.
We'll take a better look at it when we get around to making EXPLICIT_DIFFUSION a generic thing in the saber repo

travissluka:
also, as a nice side effect, i think this fixes the speckled patterns we were seeing in the envar!

image

It's hard to see, but there are "speckles" with NICAS in the T increment at some depth. NICAS is quite jagged, whereas diffusion is pretty smooth. Perhaps the NICAS would have looked better with an even higher resolution?  I think to get the resolution high enough to make it as smooth as diffusion, we would notice it running a lot slower.

guillaumevernieres (approved):
Nice! Thanks @travissluka .

kbhargava (commented):
🎉 Looks Good!! Thanks Travis :)

kbhargava (approved):

PR #1011, opened 2024-01-26
write out bkgerrgodas

travissluka:
## Description

I added the capability to write out the calculated parametric background error to a netcdf file.
See the `3dvar_godas` test now.

I still plan on rewriting the bkgerrgodas class, but this will let us know now just how noisy those fields are. I haven't looked yet with the 1/4 deg grid, but I suspect they are very noisy. @Dooruk would you mind taking a look if you have a chance?

### DIFF
--- a/src/soca/LinearVariableChange/BkgErrGodas/soca_bkgerrgodas_mod.F90
+++ b/src/soca/LinearVariableChange/BkgErrGodas/soca_bkgerrgodas_mod.F90
@@ -63,6 +63,7 @@ subroutine soca_bkgerrgodas_setup(self, f_conf, bkg, geom)
 
   type(soca_field), pointer :: field, field_bkg
   integer :: i
+  character(len=:), allocatable :: str
 
   ! Allocate memory for bkgerrgodasor
   call self%std_bkgerr%copy(bkg)
@@ -107,6 +108,12 @@ subroutine soca_bkgerrgodas_setup(self, f_conf, bkg, geom)
 
   ! Apply config bounds to background error
   call self%bounds%apply(self%std_bkgerr)
+
+  ! optionally write out the values
+  if( f_conf%has("output")) then
+    call f_conf%get_or_die("output.filename", str)
+    call self%std_bkgerr%write_file(str)
+  end if
 
 end subroutine soca_bkgerrgodas_setup
 
--- a/test/testinput/3dvar_godas.yml
+++ b/test/testinput/3dvar_godas.yml
@@ -51,6 +51,9 @@ cost function:
         efold_z: 2500.0       # [m]
 
       - linear variable change name: BkgErrGODAS
+        # optionally save the calculated parameters to a file
+        output:
+          filename: data_generated/3dvar_godas/bkgerrgodas.nc
         sst_bgerr_file: data_static/godas_sst_bgerr.nc
         t_min: 0.1
         t_max: 2.0


guillaumevernieres (approved):
:+1:

Dooruk (approved):
This works and here are outputs for 3dvar_diffusion and 3dvar_godas:

Temperature:
bkgerr_temp

travissluka:
> This works and here are outputs for 3dvar_diffusion and 3dvar_godas:
> 

Wow that is hideous, I definitely need to add in smoothing!

travissluka:
@Dooruk out of curiosity, can you see a checkerboard pattern in your input background where you see the checkerboard pattern here? If so, we've found that adding in `VERTEX_SHEAR=True` into the MOM_input file when running MOM helps with this. If you can convince the modeling people there to let you add it.

Dooruk:
> @Dooruk out of curiosity, can you see a checkerboard pattern in your input background where you see the checkerboard pattern here? If so, we've found that adding in `VERTEX_SHEAR=True` into the MOM_input file when running MOM helps with this. If you can convince the modeling people there to let you add it.

Hmm, not quite.. Perhaps @sanAkel can chime in about `VERTEX_SHEAR`. 

@travissluka so `horiz_filt` fixes that pattern with `3dvar_godas` and we don't feel the impact?

mom_res

sanAkel:
@travissluka Adding that to MOM config is fine with me!

But @Dooruk do you have a plot of it with/without it?

guillaumevernieres:
> This works and here are outputs for 3dvar_diffusion and 3dvar_godas:
> 
> Temperature: bkgerr_temp

Is that for the top layer @Dooruk ? If so, maybe there's something wrong in the configuration? The surface should look OK, it's based on omb's that @travissluka computed in his previous life.

guillaumevernieres:
We currently use a bkg error based on a 30 member ens. forecast that @liuxiao37k put together ages ago. The spread is a bit large, but the spatial patterns are what's expected:
sst-bkgerr

Dooruk:
@sanAkel I will produce the difference plot, thanks for the suggestion.

@guillaumevernieres, the config I use differ for `BalanceSOCA` and `BkgErrFILT`. Should I also not use `kst` for instance? I can just replicate what you have and make comparsisons again before `VERTEX_SHEAR`.

https://github.com/NOAA-EMC/GDASApp/blob/develop/parm/soca/berror/saber_blocks.yaml

sanAkel:
> We currently use a bkg error based on a 30 member ens. forecast that @liuxiao37k put together ages ago. The spread is a bit large, but the spatial patterns are what's expected:
> 
> sst-bkgerr
> 
> 
> 
>  

Sorry for opening another thread, but I must say @guillaumevernieres that spread looks nice 🙂

What's the min/max scale? Or color scale?

(Reminds me of a similar plot you made when you were in Bld 33!! Ages ago!)

guillaumevernieres:
You are plotting the diagonal of **B** @Dooruk , so not related to the possible issues in the balance terms like `Kst` ... Which I'm sure is a problem too! (I zero kst out for now btw). Can you compare your `BkgErrorGodas` config to  @kbhargava 's ? `BkgErrorGodas` should start with sst omb's at the surface and smoothly transition to the gradient of T. We've had it look OK in the past.

guillaumevernieres:
I closed my ncview session @sanAkel , but out of meomory it was [0, 12]C, which is a "bit" large!  I suspect it looks similar to what Eric is using? It's just the spread of a 30 some day forecast. All credits to @liuxiao37k ...

travissluka:
@Dooruk can you point me to the MOM.res.nc background file you are using, I'd like to poke around. I don't think there is something wrong with the BkgErrGODAS configuration, but rather the background.

sanAkel:
> @Dooruk can you point me to the MOM.res.nc background file you are using, I'd like to poke around. I don't think there is something wrong with the BkgErrGODAS configuration, but rather the background.

@guillaumevernieres Thanks!
Up to 12C is large! Though I haven't seen Eric's in a while- he changed few years ago (S2S v 2, v3 change- ocean model resolution went 2x).

@Dooruk Can you show me tomorrow what's all this that @travissluka is talking about? Those godas options? Ignorant here!

Dooruk:
Ok, I replicated the exact config [here](https://github.com/JCSDA-internal/skylab/blob/develop/models/mom6/defaults/bstatic.yaml) for static B and I'm getting the exact same result. Which is odd, isn't it? For now, I declare, defeat 😞 

@travissluka I put the MOM background in my (dardag) `$NOBACKUP`, in `4travis` let me know if you can't access it.

@sanAkel sure, I will be in tomorrow and I can show you what the settings are but I'm a bit monkey see, monkey do as far as the SABER config YAML specifics. I haven't changed what "works"

travissluka:
@Dooruk if you adjust your colorscale you'll see that the checkerboard pattern is actually present in the MOM.res.nc file.
image

Where are you getting this file from? Is it part of a soca analysis cycle?

I would try 
1) Adding back in the `BkgErrFILT` (two iterations maybe) as a temporary workaround. I'll be adding smoothing into `BkgErrGODAS` in February.
2) You can also try adding in the `VERTEX_SHEAR=True` option into `MOM_input` which should help those patterns from appearing in the first place. From past experience, MOM can produce those checkerboard patterns, but they tend not to be noticeable until the DA amplifies them.

Dooruk:
Thanks for looking into this. That is an output for default GEOSgcm 0.25-deg coupled model setup, so there is no SOCA involvement for this particular plot. 

I just had a discussion with @sanAkel  and he mentioned there are some other dependencies with that parameter. He wants to conduct some tests with `VERTEX_SHEAR` option turned on. I will keep you guys posted.

sanAkel:
@travissluka, As @Dooruk said, we had a chat about the model config (including that one you suggested).

I'm totally unclear of all these background error options.  @Dooruk said you are working on some examples and documentation? Or maybe it's already there? Can you point me to it?

Dooruk:
@sanAkel some relevant JEDI level documentation is [here](https://jointcenterforsatellitedataassimilation-jedi-docs.readthedocs-hosted.com/en/latest/inside/jedi-components/soca/classes.html).

PR #1013, opened 2024-01-29
Add State variable change constructor

### LINKED ISSUE #1012: Add new State Constructor required by OOPS
## Description
https://github.com/JCSDA-internal/oops/pull/2498 creates a new oops-level State constructor which must be implemented by the models.

## Requirements

The new constructor must construct a new state using an existing state, and then perform a model-to-analysis variable change. This is required so that this variable change is not needed in the FGAT cost function. (See https://github.com/JCSDA-internal/oops/issues/2484)

## Acceptance Criteria (Definition of Done)
All tests as before after the variable change has been removed from the FGAT cost function.

## Dependencies
Depends upon https://github.com/JCSDA-internal/oops/pull/2498

svahl991:
## Description

Adds the constructor described in #1012

## Issue(s) addressed

Resolves #1012

## Dependencies

build-group=https://github.com/JCSDA-internal/oops/pull/2498
build-group=https://github.com/JCSDA-internal/fv3-jedi/pull/1115
build-group=https://github.com/JCSDA-internal/mpas-jedi/pull/943

## Checklist


svahl991:
I noted that none of the tests appear to use the 3D-FGAT Cost Function, so the new constructor is not exercised in any tests; it is just required for compilation.

travissluka (approved):
Thanks @svahl991, at some point soon I will be cleaning up all the ctests, the 3dfgat cost function will be added then.

shlyaeva (approved):

svahl991:
I just added a commit to the OOPS PR which adds a test of the new State constructor. Unfortunately, I'm having some difficulty figuring out how to modify the state test yaml in soca so that the new test passes. I may need some help from soca people. I've added a commit to this PR to get you started, but the test isn't passing yet.

travissluka:
> I just added a commit to the OOPS PR which adds a test of the new State constructor. Unfortunately, I'm having some difficulty figuring out how to modify the state test yaml in soca so that the new test passes. I may need some help from soca people. I've added a commit to this PR to get you started, but the test isn't passing yet.

I'll take a look at it

src/soca/State/State.cc
```diff
@@ -67,6 +68,15 @@ namespace soca {
     Log::trace() << "State::State created by interpolation." << std::endl;
   }
   // -----------------------------------------------------------------------------
+  State::State(const oops::Variables & vars, const State & other) : State(other)
+  {
+    eckit::LocalConfiguration varChangeConfig;
```
travissluka: there are required parameters that need to be passed into `Model2Ana` (see https://github.com/JCSDA-internal/soca/blob/2db42a253e38cdfdf37572878a1d4f2bf1152efd/test/testinput/varchange_ana2model.yml#L17 "rotate" and "log"

I don't see any way to get those from the test input yaml.

svahl991: Is this a problem for more than just the test? Is this also a problem if/when soca wants to use the FGAT cost function, where this constructor is used? Can assumptions be made so that these parameters could be hard-coded in the constructor?

travissluka: @guillaumevernieres correct me if I'm wrong, but is no one actually using  `Model2Ana`?
If that's the case, @svahl991 I think we can just either disable that part of the ctest, or remove the variable transform from the constructor.... then add it to the long list of things about variable changes that I need to clean up and fix in soca.

svahl991: There is also a Parameter option to completely disable running the new test of the new state constructor. See `coupled/test/testinput/coupledstate.yaml` in the OOPS PR for an example. I just want to make sure we're not just sweeping a problem under the rug that will eventually come back to bite us.

svahl991: @travissluka Do you want me to try testing it without the variable transform in the constructor and see what happens? Or do you want to just take on the development of this PR from here?

travissluka: oh we're probably sweeping a problem under the rug... but I added a comment about it so we're okay 😆

src/soca/State/State.cc
```diff
@@ -67,6 +68,20 @@ namespace soca {
     Log::trace() << "State::State created by interpolation." << std::endl;
   }
   // -----------------------------------------------------------------------------
+  State::State(const oops::Variables & vars, const State & other) : State(other)
+  {
+    // TODO(Travis, maybe) The variable change needs to go here
+    //  (U/V rotate, etc), but since we don't really use it right now, this 
+    //  can wait until after the variable changes get cleaned up, (and
```
svahl991:
```suggestion
    //  (U/V rotate, etc), but since we don't really use it right now, this
    //  can wait until after the variable changes get cleaned up, (and
```
Removing end-of-line whitespace that  is causing the coding norms test to fail.

kbhargava (approved):
Thanks @svahl991 !! Builds fine on orion with GNU and passes the look at the code test :) That's the extent of what I tested.

svahl991:
Upstream OOPS PR is now merged.

Issue #1014, opened 2024-01-30
ML based balance


guillaumevernieres:
## Description
Develop an ML based balance operator as a saber outer block, I think that's what it should be, could you confirm @travissluka and/or @shlyaeva ?

Hard to tell yet in how many steps I will do this work, but the initial goal is to call pre-trained AI balance emulators (developed in the `daml` repo) from within `soca`. I'll add a training option later on.

## Help!
Build `libtorch` as part of the containers and on HPC. Help with this would be nice, the `gnu` version is easy to build, but I couldn't figure out how to build with intel.

shlyaeva:
Yes, I think this should be a saber outer block. If you have any questions on interfaces for saber blocks or related, let me know!

travissluka:
unfortunately, searching https://packages.spack.io/ it seems that `py-torch` is available, but not the C++ api

guillaumevernieres:
> unfortunately, searching https://packages.spack.io/ it seems that `py-torch` is available, but not the C++ api

I guess we'll stick with `gnu` for now then. Thanks for checking @travissluka .

PR #1015, opened 2024-01-31
First step towards a ML based balance operator

guillaumevernieres:
## Description
Not much was done (beyond lots of copy/paste/delete):
- creating a minimalist saber outer block that will eventually exercise the ML Balances.
- addition of a dirac test to trigger the adjoint test.

It's a draft because I don't know of a better way to select the sub-fields ... See code. 

- part of  https://github.com/JCSDA-internal/soca/issues/1014


src/soca/MLBalance/MLBalance.cc
```diff
@@ -0,0 +1,99 @@
...
+    }
+    if (field.name() == "tocn") {
+      dtIndex = index;
+    }
+    index += 1;
+  }
+
+  // Apply forward balance
+  // dc +=  kct * dsst
+  auto dcView = atlas::array::make_view<double, 2>(fset[dcIndex]);
```
shlyaeva: You can use `fset["cicen"]` here instead of finding indices, I think

src/soca/MLBalance/MLBalance.cc
```diff
@@ -0,0 +1,99 @@
...
+
+MLBalance::MLBalance(
+    const oops::GeometryData & geometryData,
+    const oops::Variables & vars,
+    const eckit::Configuration & mlbConf,
+    const Parameters_ & params,
+    const oops::FieldSet3D & xb,
+    const oops::FieldSet3D & fg)
+  : saber::SaberOuterBlockBase(params, xb.validTime()),
+    params_(params), innerVars_(vars), innerGeometryData_(geometryData), kct_(0.1)
```
shlyaeva: Do you expect that the balance operator will change the variables? If so, `innerVars_` will have to be different from `vars`.

src/soca/MLBalance/MLBalance.cc
guillaumevernieres: Ha! Thanks.

src/soca/MLBalance/MLBalance.cc
guillaumevernieres: We've always had `innerVars_=outerVars_` in our change of variables. I think that might change this year, I remember hearing @travissluka threatening to do something about it.

travissluka: I have threatened, and now the gears are in motion... prepare yourself.

travissluka:
@guillaumevernieres I talked with @shlyaeva and I think we agreed that we'll let you hack away in the soca repo with this. And once we have something that seems to work, I'll clean it up and move it into the saber repo.

src/soca/MLBalance/MLBalance.cc
guillaumevernieres: ... ok ... could you ship that medieval armor of yours my way @travissluka ?

guillaumevernieres:
> @guillaumevernieres I talked with @shlyaeva and I think we agreed that we'll let you hack away in the soca repo with this. And once we have something that seems to work, I'll clean it up and move it into the saber repo.

Happy doing my own cleanup if that's helpful. I'll try to push something a bit more complete next week, I was also thinking about dealing with ML and physics based balances in the same outer block. The only thing that will differ is how we construct the jacobian.

guillaumevernieres:
Un-drafting. I think I have enough of a skeleton of a saber outer block to start "playing" with the balances ... in future PR's (not this one).

guillaumevernieres:
I'm not sure what the jedi integration tests are supposed to be doing, but looking at the log I don't think the failure is related to this PR. Or am I missing something @travissluka ?

travissluka (approved):
I will insist, at some point, that the hardcoded variable names be removed and everything is done via yaml... at some point.

I'm assuming there's no need for `leftInverseMultiply` in the future with this?

src/soca/MLBalance/MLBalance.cc
```diff
@@ -0,0 +1,170 @@
...
+    for ( atlas::idx_t jlevel = 0; jlevel < fset["tocn"].shape(1); ++jlevel ) {
+        ds(jnode, jlevel) += dsdt(jnode, jlevel) * dt(jnode, jlevel);
+        dssh(jnode, 0) += dsshdt(jnode, jlevel) * dt(jnode, jlevel) +
+                          dsshds(jnode, jlevel) * dsi[jlevel];
+    }
+    dc(jnode, 0) += dcdsst(jnode, 0) * dt(jnode, 0) +
+                    dcdsss(jnode, 0) * dsi[0] +
+                    dcdhi(jnode, 0) * dhi(jnode, 0) +
+                    dcdhs(jnode, 0)* dhs(jnode, 0);
+  }
```
shlyaeva, line 112: Not sure how this will behave (this being using `+=` instead of `=` in TL code) if the variables gets passed all through the blocks. In the "preferred" (by me :)) flow, one would have different inner and outer variables in this block, and the relevant variables would get created and removed both here and in the adjoint, and then `ds`, `dssh`, `dc` here would be zero since they were just created, and `+=` vs `=` wouldn't matter. Not sure whether the behavior will be correct if the variables are carried through `multiplyAD`/`multiply`. 
Note: I am not requesting change for inner vs outer variables, OK with me to keep as is for now.

shlyaeva (commented):
Looks good to me, one question/comment below on where I am not sure whether this will behave correctly.

src/soca/MLBalance/MLBalance.cc
guillaumevernieres: No, I think you might have a valid point @shlyaeva ... I see that's how the vertical balance is done in saber. Hummmm ...

guillaumevernieres:
> I will insist, at some point, that the hardcoded variable names be removed and everything is done via yaml... at some point.
> 
> I'm assuming there's no need for `leftInverseMultiply` in the future with this?

No, there is a need for it ... at some point in the future.

src/soca/MLBalance/MLBalance.cc
guillaumevernieres: What's coded here is aligned with what we've been doing so far @shlyaeva . **K** in our case is square and the diagonal is 1 ... so ```+=``` is correct ... I hope!

shlyaeva (approved):
Thank you for confirming @guillaumevernieres !

PR #1025, opened 2024-03-08
fix sea_water_depth calculation

travissluka:
## Description

The `sea_water_depth` was being calculated from `geom%h` instead of the `hocn` variable in the state. This means the interpolated insitu obs (T/S) were slightly wrong. This mainly impacts the 4d and fgat tests where `hocn` changes over time.

@guillaumevernieres I'm assuming this doesn't impact your fgat since you're using the diag files on the fixed zstar grid?

I found this while ripping out `geom%h` in a separate branch. `h` doesn't belong in `geom`


kbhargava (approved):
Looks reasonable. Haven't tested

guillaumevernieres (approved):
Correct, it doesn't affect us. But still a bug. Nice catch.

Issue #886, opened 2023-05-15
Adapt to newer MOM6 used in UFS


travissluka:
## Description
Our current version of MOM6 is ancient. We need to make changes to use the newer version of MOM6 used by the UFS bundle

My initial look at this resulted in segfaults when trying to initialize SOCA, I need to figure out why.

symoore90:
Hey @travissluka , is there an epic you can attach this issue to?

travissluka:
fixed by https://github.com/JCSDA-internal/soca/pull/1043

PR #1026, opened 2024-03-19
Remove dependency on MOM6 and FMS forks (part 1)

### LINKED ISSUE #1023: Delete MOM6 model forecast from SOCA
I just want to make sure I'm not crazy before doing this, but the way I see it, the model class shouldn't live in soca:

1) We don't run in core right now, we do everything with the pseudo model, it'll stay that way for the foreseeable future.
2) In the distant future (i.e. not this year) when we want to do things in-core, we'll never run MOM6 by itself, an ocean model is not very useful without an atmospheric coupler to calculate the fluxes, so we would always be running some type of coupled model (MOM6SIS2, UFS S2S, UFS NGGODAS, GEOS, ..etc) each of which compile and interface differently. I like @shlyaeva 's idea of having the coupled model classes living outside of fv3/soca repos,

This will help us now with compiling soca with the ufs-bundle. And, if later when we go to do in-core it turns out that I'm wrong, we can just put the model forecast back into soca.

(I suppose I could make an exception and let @guillaumevernieres add an AI-based forecast model  for sea-ice that lives in soca)

thoughts? @guillaumevernieres @shlyaeva @kbhargava @climbfuji

travissluka:
- closes https://github.com/JCSDA-internal/soca/issues/1023
- partly addresses  https://github.com/JCSDA-internal/soca/issues/886

This PR removes our dependency on the JCSDA forks of MOM6 and FMS. The changes here allow SOCA to work with either the current old MOM6 fork or the latest version of the upstream MOM6 repo (after changing 1 or 2 lines of code in soca). This PR does NOT  yet switch to the new MOM6 repo (for CI and FMS version reasons) but this will be done in a later PR.

To accomplish this:
1. The soca `MOM6SOLO` model has been removed. All soca executables requiring a model should use `Identity` or `PseudoModel` now. If we want in-core capabilities in the future, we'll re-evaluate and create proper coupled model classes.
2. The 3D interpolation used by the convert state has been removed. It wasn't being used and was a bit of a mess. I hope we can adopt what @fmahebert is putting together for resolution changes, plus perhaps some soca-specific code to handle the vertical interpolations.

Neither of these is used in the real world so there is no point in maintaining them.

## Details
- A submodule pointing to the NOAA-EMC MOM6 repo has been added, as well as the cmake files needed to compile the mom6 library and `mom6solo` executable. See `./mom6` directory. This has been done in the same way as what the UFS code does. This is not being used with this PR, but will be turned on in part 2 (It does work though, trust me.)
- `soca_checkpoint_model.x` has been removed. It was relying on private mom6 internals. In its place, I have created a generic Python script at `./tools/soca_update_restart.py`. The script will update restart files and impose optional bounds on analysis values and/or increment magnitudes.
- everything in `./src/soca/Model/` has been deleted! The remaining grid decomposition code interfacing with mom6 has been moved to `./src/soca/Geometry/soca_geom_mod.F90`
- `h` has been removed from the geometry. It is not a static field, and so belongs solely in the state. As a consequence:
  - `h_zstar` is no longer calculated by geometry
  - `h` and `h_zstar` are no longer saved in the gridgen output (it's not needed)
  - `soca_geom_thickness2depth` removed from geometry
  - the 3D geometry iterator is temporarily broken since it relies on `h` being in geometry. The associated ctest is disabled. We don't use the 3d iterator right now, so it's not a problem, but I'll need to figure out later how to cleanly fix this.
- The `convert_state` code has been removed and is no longer being used by increment or state. This code was a bit of a mess and needs to be redone cleanly.
- The ctests now use the mom6 solo executable to run forecasts for the 4d/fgat tests. A python script `./test/mom6solo.py` is added to help with running the executable as part of the ctests. (the script is a bit over-engineered, but I did let chatgpt write most of it)
- all the tests relying on the forecast ctests have been updated, and answers changed

jedi-ci-build-cache=skip


travissluka:
@eap @climbfuji I'm trying to test this PR along with the matching one for jedi-bundle (https://github.com/JCSDA-internal/jedi-bundle/pull/85) but the CI seems to keep using the `develop` version of `jedi-bundle`. 

logs show it's not picking up the bundle from the PR
```
ecbuild_bundle( PROJECT mom6 GIT "https://github.com/jcsda-internal/MOM6.git" TAG 51ec489ad7d8a86762bef4c46eabd9af5fc41fa4 RECURSIVE )
ecbuild_bundle( PROJECT soca GIT "https://github.com/jcsda-internal/soca.git" TAG deb09180b6759c52f82ab4a194727a54ba7a094c )
```

should `build-group` work with jedi-bundle?

climbfuji:
> @eap @climbfuji I'm trying to test this PR along with the matching one for jedi-bundle ([JCSDA-internal/jedi-bundle#85](https://github.com/JCSDA-internal/jedi-bundle/pull/85)) but the CI seems to keep using the `develop` version of `jedi-bundle`.
> 
> logs show it's not picking up the bundle from the PR
> 
> ```
> ecbuild_bundle( PROJECT mom6 GIT "https://github.com/jcsda-internal/MOM6.git" TAG 51ec489ad7d8a86762bef4c46eabd9af5fc41fa4 RECURSIVE )
> ecbuild_bundle( PROJECT soca GIT "https://github.com/jcsda-internal/soca.git" TAG deb09180b6759c52f82ab4a194727a54ba7a094c )
> ```
> 
> should `build-group` work with jedi-bundle?

Unfortunately not yet

travissluka:
> > @eap @climbfuji I'm trying to test this PR along with the matching one for jedi-bundle ([JCSDA-internal/jedi-bundle#85](https://github.com/JCSDA-internal/jedi-bundle/pull/85)) but the CI seems to keep using the `develop` version of `jedi-bundle`.
> > logs show it's not picking up the bundle from the PR
> > ```
> > ecbuild_bundle( PROJECT mom6 GIT "https://github.com/jcsda-internal/MOM6.git" TAG 51ec489ad7d8a86762bef4c46eabd9af5fc41fa4 RECURSIVE )
> > ecbuild_bundle( PROJECT soca GIT "https://github.com/jcsda-internal/soca.git" TAG deb09180b6759c52f82ab4a194727a54ba7a094c )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > should `build-group` work with jedi-bundle?
> 
> Unfortunately not yet

ah, in that case, @climbfuji, how would you suggest we proceed with testing/merging?
I can try to see if this can be done in two stages.
1) merge most of this PR, but still using the old MOM6 repo as part of bundle then (I think it should still work with the old MOM)
2) remove mom6 from bundle and use submodule instead

travissluka:
image

hmm, now, what else can I delete from SOCA :thinking:

tools/soca_update_restart.py
```diff
@@ -0,0 +1,124 @@
...
+
+          ana_var = inputNcd.variables[var_name][:]
+          rst_var = restartNcd.variables[var_name][:]
+
+          # bounds check on increment
+          if 'increment max' in varConfig:
+            maxInc = varConfig['increment max']
+            print("    imposing max increment of ", maxInc)
+            inc = ana_var - rst_var
+            inc = np.clip(inc, -maxInc, maxInc)
```
kbhargava, line 106: Stupid question, do we ever have situations where we would want min and max increments to be different?

travissluka: I can't think of any. My rationale here was we might want to just limit the magnitude of the increment in some cases, such as U/V when using LETKF

kbhargava (approved):
Builds and tests fine on orion. 🎉

guillaumevernieres (approved):
Looks good.

shlyaeva:
@travissluka guess what! soca just broke something. 😆 

(Note: this was caught by integration CI)

shlyaeva:
coupling tests no longer pass, gridgen fails with:
>NOTE from PE     0: open_param_file: data_static/72x35x25/MOM_input has been opened successfully.
FATAL from PE     1: open_param_file: Input file MOM_override does not exist.

travissluka:
> coupling tests no longer pass, gridgen fails with:
> 
> > NOTE from PE     0: open_param_file: data_static/72x35x25/MOM_input has been opened successfully.
> > FATAL from PE     1: open_param_file: Input file MOM_override does not exist.

ah, oops, I'll fix it

travissluka:
> @travissluka guess what! soca just broke something. 😆
> 
> (Note: this was caught by integration CI)

we really have to get rid of those "acceptable number of failures = pass", I missed it!

shlyaeva:
(note: there's one test that is failing in coupling -- it's a known failure, I've just started working on fixing it, which is how I noticed this)

shlyaeva:
> we really have to get rid of those "acceptable number of failures = pass", I missed it!

it looks like we may be able to do that this week if all goes well.

PR #1030, opened 2024-03-26
Fix bug introduced by recent PR

travissluka:
## Description
- fixes problem created by https://github.com/JCSDA-internal/soca/pull/1026

ctests had been modified so that a 'MOM_override' file is dynamically created, which is needed by the mom6solo forecast. However this broke `coupling` because there was no `MOM_override` file.

I reverted the `input.nml` back to how it was, and instead create a new `input_mom6solo.nml` for use with the stand alone mom6solo forecast ctests. 
(not sure what I was thinking previously, it's cleaner this way anyway.)

build-group=https://github.com/JCSDA-internal/coupling/pull/40

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -274,7 +274,6 @@ function(soca_add_test)
       ${WORKDIR}/data_static/workdir/${FILENAME}
       ${WORKDIR}/${FILENAME} )
   endforeach()
-  file(WRITE ${WORKDIR}/MOM_override "")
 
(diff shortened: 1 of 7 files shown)


travissluka:
note, the integration tests that failed are fixed by https://github.com/JCSDA-internal/coupling/pull/40

shlyaeva (approved):
Thank you @travissluka! 🎉

kbhargava (approved):
Looks good.

PR #1033, opened 2024-04-17
prepare to switch from custom interp code (soca_remap_idw) to generic atlas interp

travissluka:
**_part of my inevitable march toward soca using more generic and atlas-ified C++ code, and less Fortran_**

The main part of this PR is
- the C++ implementation of an interpolator from an arbitrary set of lat/lon points to an atlas FieldSet. This is to replace (in the next PR) the Fortran `soca_remap_idw` interpolator completely. 
- I've started to use atlas fields within the geometry (`rossby_radius` and `distance_from_coast` now exist solely within the FieldSet, more variables will be moved over as they are needed by C++ code)

In support of the above:
- see `src/soca/Utils/readNcAndInterp.h`. (FYI @fmahebert, it might be nice to have a more polished up version in oops someday that can read arbitrary 1D or 2D files and interpolate to the model's FunctionSpace, but this is what I have for now for SOCA)
- The interpolator is tested with the `rossby_radius` field, which is now being read into an atlas field on the C++ side
- small change in test answers because the rossby radius interpolation is slightly different in some places. Nothing big (I'll post a plot if I'm ambitious)
- the `Geometry` class has been cleaned up a little bit because of the above changes. The writing and reading to/from soca_gridspec file now has a section of code that will read/write arbitrary fields to/from the atlas FieldSet in `Geometry`, for now it still uses the FMS IO code for backward compatibility, but someday I'll switch to using the generic atlas IO.
- 60 fewer lines of code, despite adding in an entire new interpolator in C++ !

>[!WARNING]
> @Dooruk , @guillaumevernieres the `rossrad.dat` file has been converted into a netCDF file. **This will require anyone running gridgen to have their yaml point to the nc file, not the dat file, this should be the only end-user change**

build-group=https://github.com/JCSDA-internal/coupling/pull/43


Dooruk:
Thanks for the heads up!

travissluka:
note: the integration tests pass with https://github.com/JCSDA-internal/coupling/pull/43

guillaumevernieres (approved):
looks good

Dooruk (approved):

PR #1035, opened 2024-04-25
Refactor C++ Increment/State to use FieldSets, and delete lots of Fortran!

travissluka:
## Description

This PR largely refactors `Increment` and `State` C++ classes so that the data are primarily stored in `atlas::FieldSet`. This allows us to use generic math operators on the C++ side, and eliminate a LOT of Fortran interface code (1,500 total fewer lines of code here!)

This change has the following impacts

- `Increment` and `State` C++ classes now have a shared base class `Fields`, where any common functions live
- The "authoritative" copy of fields live in the atlas fieldset (`Fields::fieldSet_`). The "fields" on the Fortan side is, for now, duplicated with what is in the atlas fieldset. The legacy sections of Fortran code therefore needs to call in their interface `sync_from_atlas()` and `sync_to_atlas()`. This is temporary. As I refactor those sections of Fortran, some of them will be converted into pure C++, and what remains will directly use the atlas fieldset on the fortran side.
- This generalization has let me remove ALL of the Fortran code associated with the `GeometryIterator`, yay! The code there is much cleaner and no longer has the hardcoded variable names which I have despised for many years now.
- Look at `soca_increment.interface.F90` and `soca_state.interface.F90` to get an idea of how much C++/Fortran interface code as been removed.

@fmahebert FYI, I think most of what is in the C++ `Fields`, `State`, and `Increment` is relatively generic and is probably a good starting point for the base classes you talked about. One thing I did think about: we eventually need a better way to share fieldsets other than `toFieldSet` and `fromFieldSet`. It seems those methods need to make a copy of the fieldsets, otherwise everything breaks.

## Follow up PRs
- Some of our other soca classes will be refactored into pure C++ where it makes sense
- anything left on the fortran side will use atlas directly, and the Fortran `Fields` / `Field` classes will be removed, along with the `sync_from_atlas()`/ `sync_to_atlas()` subroutines

> [!IMPORTANT] 
> Henceforth, code that is added to SOCA should be done in C++, unless it makes sense otherwise. I imagine we will eventually get rid of all Fortran except where I/O or  interfacing with the model is required.

## Testing
There is no change in answers

- [x] ctests pass
- [x] I still need to run a test with 1/4 deg global to make sure there is no significant impact on cpu/memory with this PR


travissluka:
the lastest graph showing how much Fortran has been removed!
image

src/soca/Fields/Fields.cc
```diff
@@ -0,0 +1,211 @@
...
+  // keep a copy of the metadata to copy back
+  std::map<std::string, atlas::util::Metadata> metadata;
+  for (const auto & f : fieldSet_) metadata[f.name()] = f.metadata();
+
+  util::copyFieldSet(fset, fieldSet_);
+
+  for (auto & f : fieldSet_) {
+    if (metadata[f.name()].has("name")) {  // that's weird
+      // THERE IS A BUG IN SOCA, i'm getting fieldsets with one or two fields
+      // (hocn) that have have no metadata even though the rest of the field
```
guillaumevernieres, line 202: Missing a few haves. More is better.

guillaumevernieres (commented):
Thanks @travissluka . I'll build and test this later this week, that's a lot of changes!

travissluka:
> Thanks @travissluka . I'll build and test this later this week, that's a lot of changes!

A "lot of changes" that doesn't actually change anything 🤔

travissluka:
@danholdaway FYI, if you're curious about what I did. Everything on the C++ side of `State`, `Increment`, and `Fields` _should_  be mostly generic and could be abstracted away eventually.

kbhargava (commented):
built fine on orion with intel. ctests are painfully slow and some fail because of tolerance issues.

kbhargava (approved):
Finally merged this with the cumulative bugfix PR and all the ctests pass on orion with intel. 🎉

src/soca/Fields/Fields.cc
```diff
@@ -0,0 +1,211 @@
...
+                          std::right << max
+        << "   mean=" <<  std::fixed << std::setw(12) <<
+                          std::right << sum;
+  }
+}
+
+// -----------------------------------------------------------------------------
+
+void Fields::toFieldSet(atlas::FieldSet & fset) const {
+  // copy seems like a waste, it would be nice if we could get away with "share"
```
fmahebert, line 186: And what goes wrong if you did share?

In `lfric-jedi`, if I'm not mistaken, the implementation follows this pattern (not this code):
```
fset.clear();
for (auto var : variables_.variables()) {
  fset.add(fields_[var]);
}
```
with `oops::Variables variables_` and `atlas::FieldSet fields_` being member variables for the State/Increment. Could also loop directly over `fields_`, but keeping a Variables lets you share only a subset of the model fields with oops, in case that's desired.

fmahebert (commented):
Overall this is very much in line with what I'm trying to do in the generic layer! Nice. I think we'll be able to combine efforts as soon as mine gets off the ground.

I did think sharing the Fields in `to/fromFieldSet` was possible. Would be curious to know if you ran into any problems with it. See snippet:

src/soca/Fields/Fields.cc
travissluka: I tried sharing and it did break, answers were drastically different... but there were several things going on at once so I'll give it a try again and track down the culprit if it's still breaking.

travissluka: the `soca_enspert.x` app is giving different answers if I share the fields. I'm not sure if this is a problem with the enseprt app, or with the randomization with my diffusion saber central block. :man_shrugging:

travissluka:
FYI, I ran a comparison with the 1/4 deg marine skylab, and as expected there is a slight performance penalty. The main culprit is the linear variable change, where we are going into and out of the Fortan code repeatedly, and resyncing between the fortran soca::fields and the atlas fieldset. Again, this is expected, and since the performance difference isn't that big, I'd like to go ahead and merge this PR. In the next round of PRs, those variable change classes will be refactored and the performance should go back to where it was.

**develop branch**

<pre><code>OOPS_STATS Name                                                :     min (ms)    max (ms)    avg (ms)     % total   imbal (%)
OOPS_STATS oops::Covariance::SABER::inverseMultiply            :     44333.99    44533.20    44391.99        8.60        0.45
OOPS_STATS oops::Covariance::SABER::multiply                   :    337049.10   362496.37   346742.92       67.19        7.34
OOPS_STATS oops::Increment::Increment                          :      7367.82     9381.67     8626.59        1.67       23.34
OOPS_STATS oops::Increment::operator=                          :      2280.99     3127.14     2859.12        0.55       29.59
OOPS_STATS oops::LinearVariableChange::changeVarAD             :     17412.78    24798.75    21151.26        4.10       34.92
OOPS_STATS oops::LinearVariableChange::changeVarTL             :     16553.97    24273.31    20867.89        4.04       36.99
OOPS_STATS util::Timers::Total                                 :    516045.91   516126.14   516047.18      100.00        0.02
OOPS_STATS util::Timers::measured                              :    475320.04   500873.69   486793.62       94.33        5.25


OOPS_STATS Run end                                  - Runtime:    519.70 sec,  Memory: total:   426.75 Gb, per task: min =     2.65 Gb, max =     2.83 Gb
</code></pre>

**this branch**
<pre><code>OOPS_STATS Name                                                :     min (ms)    max (ms)    avg (ms)     % total   imbal (%)
OOPS_STATS oops::Covariance::SABER::inverseMultiply            :     51618.27    51643.04    51627.11        8.27        0.05
OOPS_STATS oops::Covariance::SABER::multiply                   :    417350.61   418622.83   418054.42       66.95        0.30
OOPS_STATS oops::Increment::Increment                          :     18759.49    22306.41    21407.01        3.43       16.57
OOPS_STATS oops::Increment::operator=                          :     17884.01    19811.57    18618.22        2.98       10.35
OOPS_STATS oops::LinearVariableChange::changeVarAD             :     52585.81    68445.55    57595.87        9.22       27.54
OOPS_STATS oops::LinearVariableChange::changeVarTL             :     53254.06    55689.41    54249.85        8.69        4.49
OOPS_STATS util::Timers::Total                                 :    624440.19   624513.55   624441.78      100.00        0.01
OOPS_STATS util::Timers::measured                              :    593198.79   611025.18   599797.93       96.05        2.97

OOPS_STATS Run end                                  - Runtime:    627.09 sec,  Memory: total:   452.28 Gb, per task: min =     2.78 Gb, max =     3.02 Gb
</code></pre></div>

src/soca/Fields/Fields.cc
fmahebert: Ah bummer. Seems reasonable to me to merge this PR as-is and figure out this issue with sharing fields and enspert as a second step. Let me know if it's helpful to discuss.

fmahebert:
@travissluka ~do you understand why your timings show SABER multiply/inverse taking 15-20% longer? I would have expected that to be all on the atlas side and not affected by changes to the model interface. Am I misunderstanding that separation?~

I did misunderstand — that timer entry is, I think, the interface layer (`ModelSpaceCovarianceBase`), and wraps linear var changes, Increment copy constructors, all kinds of things.

fmahebert (approved):

src/soca/Increment/Increment.cc
```diff
@@ -38,242 +40,241 @@ namespace soca {
...
+    x1_at_geomres.toFieldSet(fs1); x2_at_geomres.toFieldSet(fs2);
+    util::copyFieldSet(fs1, fieldSet_);
+    util::subtractFieldSets(fieldSet_, fs2);
   }
+
   // -----------------------------------------------------------------------------
+
   Increment & Increment::operator=(const Increment & rhs) {
     time_ = rhs.time_;
+    vars_ = rhs.vars_;
```
guillaumevernieres: That's a change of behavior and it's not consistent with what's in the fortran below. We used to be able to make a copy of a subset of an increment, which was useful. Are we getting rid of that feature?
I can't say I care either way, except that this solution means I have to change things in the `gdasapp`! Ho nooo ... more work for g.

travissluka: @guillaumevernieres good point
@fmahebert do you know what the expected behavior of `Increment::operator=` is? should it adopt all the variables of the right hand side, or should it only copy the variables it was created with?

danholdaway: Not that it's necessarily correct but in fv3-jedi the code only allows a copy when the variables are [the same](https://github.com/JCSDA-internal/fv3-jedi/blob/787080f3c254748ad79ffdb6512bf232e46cf6d5/src/fv3jedi/Fields/fv3jedi_fields_mod.f90#L199). What's the use case where they would need to be different?

guillaumevernieres: Probably more for downstream applications @danholdaway , anytime you would need to extract a subset of the increment. 
It's used in the gdasapp for the ensemble re-centering, where we need to re-compute the steric height and separate the balanced/unbalanced terms.

fmahebert: I don't know that `Increment::operator=` has a required behavior from OOPS's point of view. Uh oh, do I need to start a new standardization mission?

OOPS does seem to be working with the fv3jedi implementation that, as Dan says, requires the variables to match. Besides that I'm not sure...

travissluka: I'll try to put the behavior back to how it was, because as we all know the last thing I would ever want to do is break Guillaume's code. @fmahebert I think that when you start work on the generic layer we should more explicitly define the expected behavior of the methods. For example, I couldn't use the built-in `util::addFieldSets()` for `Increment::operator+=` because the `util` function can't handle lhs and rhs having different variables.... things like that.

travissluka: @guillaumevernieres I changed ` vars_ = rhs.vars_;` to ` vars_ += rhs.vars_;`, and since the fortran side of the code is the same,  `Increment::operator=` should behave the same as before now. Let me know if you're still having issues

guillaumevernieres: No luck @travissluka , same error as before (`soca_fields::get():  cannot find field tocn`). But that solution also defeats the purpose of subsetting the increment? no?
What works is to just remove `L108` until the core team decides what should or should not be allowed? I vote for that solution obviously ...

travissluka: @guillaumevernieres yep, yep, you're right. I need more coffee.

guillaumevernieres (approved):
All good. Thanks @travissluka

PR #1038, opened 2024-05-03
Generic diffusion operator

travissluka:
## Description

Deletes the soca-specific diffusion saber central block and uses the generic diffusion operator added in saber. There are small differences in answers because 
 1) randomization is using different random numbers
 2) The grid properties are slightly different. Previously I used `dx`, `dy` directly from MOM6, but with the generic diffusion the properties of the dual mesh are calculated by atlas, and are slightly different.

I've also added a python tool to convert the output of the `saber::Diffusion` calibration into the same grid used by SOCA.

By generalizing the code and moving to saber, there goes another 1700 lines of soca code :bomb: !

> [!IMPORTANT]
> users (@guillaumevernieres, @Dooruk) will need to change their yaml file for the saber central block, see the changes in the `testinput/3dvar.yaml` in this PR. Also, unfortunately, the calibration will need to be redone, as the file format is now different. The precalculated horizontal and vertical diffusion parameter files will be available on the HPCs in `$EWOK_STATIC_DATA/skylab-9.0.0/soca/` once the following issue is closed:
> 
> - https://github.com/JCSDA-internal/r2d2-ingest/issues/127

## Dependencies
waiting for skylab to be updated
- https://github.com/JCSDA-internal/skylab/pull/538


hga007:
@travissluka: Thank you! It compiles, but it fails when running all my Unit Tests. For example, in the Geometry test, I get:

```
test_romsjedi_geometry: /opt/src/atlas-0.35.1/src/atlas/array/SVector.h:122: const T &atlas::array::SVector<T>
::operator[](int) const [with T = int]: Assertion `data_ && idx < size_' failed.
[pontus:3486125] *** Process received signal ***
[pontus:3486125] Signal: Aborted (6)
[pontus:3486125] Signal code:  (-6)
[pontus:3486125] [ 0] /lib64/libc.so.6(+0x54db0)[0x15372b654db0]
[pontus:3486125] [ 1] /lib64/libc.so.6(+0xa154c)[0x15372b6a154c]
[pontus:3486125] [ 2] /lib64/libc.so.6(raise+0x16)[0x15372b654d06]
[pontus:3486125] [ 3] /lib64/libc.so.6(abort+0xd3)[0x15372b6287f3]
[pontus:3486125] [ 4] /lib64/libc.so.6(+0x2871b)[0x15372b62871b]
[pontus:3486125] [ 5] /lib64/libc.so.6(+0x4dca6)[0x15372b64dca6]
[pontus:3486125] [ 6] test_romsjedi_geometry: /opt/src/atlas-0.35.1/src/atlas/array/SVector.h:122: const T &at
las::array::SVector<T>::operator[](int) const [with T = int]: Assertion `data_ && idx < size_' failed.
```

All my tests passed when I ran **`feature/remove_geometry_latlon`** today. Is your branch **`feature/diffusion`** up to date with OOPS and the SABER **`develop`** branches? 

I followed your example and my **parameters_diffusion.yaml** has 

```yaml
background error:
  covariance model: SABER
  saber central block:
    saber block name: DIFFUSION
    calibration:
      normalization:
        iterations: 1000                              # Ramdomization approach
      groups:
      - horizontal:
          fixed value: 160.0e3
        write:
          filepath: Data/diffusion
          files prefix: wc13_diffusion_hcor
      - vertical:
          fixes value: 150.0
          levels: 10
        write:
          filepath: Data/diffusion
          files prefix: wc13_diffusion_vcor
```

I need clarification on what the token **levels** means. In ROMS-JEDI, **levelsAreTopDown()=false**. Does this imply that the vertical correlation scale is the thickness of the top 10 vertical levels?

Also, how do you specify different correlation scales for each state variable? In our applications, for example, we have different scales for temperature and salinity.

travissluka:
@hga007 
- I just merged develop into the oops and saber branches.
- The vertical scales, as mentioned here (https://github.com/JCSDA-internal/saber/pull/847) are given as the number of levels, not meters. So with the yaml you gave that would be 150 levels, and would result in way too many iterations of vertical diffusion.
- the `levels` parameters is just the number of levels your model has, and is hopefully temporary. It's needed in order to read in 3D  files correctly, see comment https://github.com/JCSDA-internal/saber/blob/8a94f62a891e85f8109e72b5300abf82d7a51ced/src/saber/diffusion/Diffusion.cc#L282
- for different length scales for different variables, just have multiple groups in your calibration (e.g. see https://github.com/JCSDA-internal/soca/blob/59bd73cd0e20268f0a8b6d0b59f7c3e908de3980/test/testinput/parameters_diffusion.yml which has two `horizontal` groups) note that there is no need to specify the actual variables at this point, you just need to make sure your output files names are unique, 
then when using diffusion you specify which variables are used with which groups (e.g. see https://github.com/JCSDA-internal/soca/blob/59bd73cd0e20268f0a8b6d0b59f7c3e908de3980/test/testinput/3dhyb_diffusion.yml#L30)

hga007:
@travissluka:  Thank you. I'll try again then. I suspected that was implied by the **levels** token. For us, 150 meters is the vertical scale for temperature, and 30m is the vertical scale for salinity. Anyway, for testing, I will use many levels for the vertical scale. This can be fine-tuned later. The critical issue here is to evaluate the generic pseudo-diffusion operation as a method to spread covariance influence.

We want to compare horizontal convolutions for SSH between DIFFUSION and BUMP_NICAS operators. There is an issue in BUMP_NICAS for SSH, but I haven't been able to check in the debugger because of the Fortran Phyton Preprocessor, which is unfortunate. We get assembly code that it is impossible to understand.

guillaumevernieres (approved):

hga007:
@travissluka:  I tried again but still got the same error in **`atlas::array::SVector`** shown above.  Since it works for others, I wonder if I am missing an **ATLAS** update or ahead of the updates.

travissluka:
> @travissluka: I tried again but still got the same error in **`atlas::array::SVector`** shown above. Since it works for others, I wonder if I am missing an **ATLAS** update or ahead of the updates.

ah, I see, so nothing works even if you're not using `DIFFUSION` ?
Try to comment out this line in oops that I added https://github.com/JCSDA-internal/oops/blob/0368ea9ade2ee88d704b9be214576cb46ec69196/src/oops/base/GeometryData.cc#L334
It's the only change I made that isn't part of the `DIFFUSION` block, let me know if that works and we'll go from there

hga007:
@travissluka: Yes, the Geometry object fails. Thus, none of the Unit Tests work. If I comment on the line in GeometryData.cc that you suggest, I go a little further, and I get the following error:
```
Test "interface/Geometry/testConstructor" failed with unhandled eckit::Exception: Trying to access field `xyz' in Nodes, but no field with this name is present in Nodes. @  (/opt/src/atlas-0.35.1/src/atlas/mesh/Nodes.cc +77 field)
```
I suspect the issue is that ROMS has a different grid specification in ATLAS than the one used in SOCA.

travissluka:
> @travissluka: Yes, the Geometry object fails. Thus, none of the Unit Tests work. If I comment on the line in GeometryData.cc that you suggest, I go a little further, and I get the following error:
> 
> ```
> Test "interface/Geometry/testConstructor" failed with unhandled eckit::Exception: Trying to access field `xyz' in Nodes, but no field with this name is present in Nodes. @  (/opt/src/atlas-0.35.1/src/atlas/mesh/Nodes.cc +77 field)
> ```
> 
> I suspect the issue is that ROMS has a different grid specification in ATLAS than the one used in SOCA.

The only thing I added to oops was the `build_edges(mesh_);` line, so if your geometry constructor test is failing with that line commented then it is something unrelated to these diffusion PRs. Are you sure everything was working for you with the develop branches?

Can we move this discussion elsewhere, since this issue is not related to this SOCA PR.

hga007:
@travissluka: Sure, we can move the discussion somewhere else.  Everything worked when I tested the changes due to **feature/remove_geomety_latlon**.  I will test the **develop** branch to ensure it works.

hga007:
@travissluka: I can confirm that all the ROMS-JEDI Unit Tests run successfully with the **develop** branch.  The behavior I am getting is likely due to the changes in **feature/diffusion** in OOPS.  I am using atlas/0.35.1 for ifort.  Do I need to use a newer ATLAS version or a different OOPS and SABER branch?

byoung-joo:
Hi @travissluka , 
As you can see in the JEDI integration test, most (or all?) mpasjedi ctest failed.
For example of mpasjedi geometry unit test, https://cdash.jcsda.org/test/10860581
Does something need to be done in mpas-jedi for `atlas::mesh::actions::build_edges(mesh_);` (added in oops)?

maybe this should go to OOPS pr or jcsda discussion...

test/testinput/3dhyb.yml
```diff
@@ -25,22 +25,15 @@ cost function:
...
-          saber block name: EXPLICIT_DIFFUSION
-          active variables: [cicen, hicen, hsnon, tocn, socn, ssh, uocn, vocn]
-          geometry: *geom
-          group mapping:
-          - name: group1
-            variables: [tocn, socn, cicen, hicen, hsnon, uocn, vocn, ssh]
+          saber block name: diffusion
           read:
             groups:
-            - name: group1
```
kbhargava, line 36: One more yml file, [3dvar_diffmlb.yml](https://github.com/JCSDA-internal/soca/blob/develop/test/testinput/3dvar_diffmlb.yml), might need to be changed similarly. I have never paid attention to that test so don't know much.

tools/convert_atlas_file.py
```diff
@@ -0,0 +1,59 @@
...
+  latitudes[:] = np.mean(dstLat, axis=1)  # Assuming dstLat is 2D
+  longitudes[:] = np.mean(dstLon, axis=0)  # Assuming dstLon is 2D  
+
+  # Assign attributes to latitude and longitude variables
+  latitudes.units = 'degrees_north'
+  latitudes.standard_name = 'latitude'
+  longitudes.units = 'degrees_east'
+  longitudes.standard_name = 'longitude'
+
+  # # use KD tree to find mapping between src and dst points
```
kbhargava: Do you want to keep these comment section here for future purposes?

kbhargava (approved):
Looks good 🎉. Ctests pass on orion with GNU. I just had a couple comments that I put. Approving as I never use that one yaml file that wasn't changed.

test/testinput/3dhyb.yml
travissluka: I have no way of testing that file, so I'm not going to touch it

PR #1042, opened 2024-05-09
build mom6.x when building only soca

### LINKED ISSUE #1039: fix cmake soca dependency on mom6.x
## Description
the soca ctests need `mom6.x` but this is not built if building from inside `<build_dir>/soca`.

Add the CMakeLists.txt magic so that mom6.x is built as a dependency when soca-only is built.

travissluka:
## Description

- closes #1039

when building only soca (i.e. running `make` from within `build/soca`) the `mom6.x` executable was not being built automatically because it was not setup as a dependency.

This fixes that by adding `mom6.x` as a dependency of the tests.


Dooruk (approved):
Looks good 👍

kbhargava (approved):
Thanks @travissluka . Works fine on orion.

PR #1052, opened 2024-07-01
[ignore] Testing CI change

eap:
Testing CI due to change in https://github.com/JCSDA-internal/jedi-bundle/pull/76

run-ci-on-draft=true
debug-ci=false
jedi-ci-test-select=clang

Most of the tests are running in this pull request https://github.com/JCSDA-internal/oops/pull/2483, in order to not over-stress the CI system I am only testing one environment at a time here. If any tests pass here and all tests pass in oops, it is safe to assume the migration worked.


eap:
I cancelled the first run and re-started with debug disabled.

PR #1056, opened 2024-07-10
Add the capability to run 4dvar using HTLM

### LINKED ISSUE #1044: SOCA3: Ocean 4DVAR with Hybrid TLM
## Description
Demostrate HTLM 4Dvar  capabilities with regional 1/12 deg MOM6 (i.e. science). Using the HTLM capabilities that were recently added to JEDI, use an ensemble (vertical) and identity TL/AD (horizontal) to demonstrate feasibility of 4DVAR. 

## Sub tasks

- [ ]  Implement marine HTLM
     - [ ]  Run 3DGFAT for couple weeks with 1/4 degree model to establish a baseline
     - [ ]  Using Identity as simplified TLM, use HTLM and run for the same period as the above, debug, repeat
     - [ ]  Compare and contrast to check if HTLM is helpful
- [ ] Use generic advection/diffusion model in HTLM (low priority)

kbhargava:
## Description

This PR adds the capability to run a 4dVar using hybrid tangent linear model (HTLM). This includes generating coefficients for the HTLM from an ensemble of non-linear model runs, fine -tuning using different regularization evaluating the linearization errors, and reading these HTLM coefficients when running 4dVar. Currently, an identity is used at the simple linear model. 

**NOTES**

1. When calculating HTLM coefficients, a preconditioning step that uses RMS by Level [here](https://github.com/JCSDA-internal/oops/blob/db4a1d71f50ff4c262feefb4cc5588cbece68961/src/oops/generic/HtlmCalculator.h#L88). For the ocean, this calculation doesn't take into account the land mask. The land values would be 0 but this would impact the results. 
2. The test values generated in the generate coefficients test are perturbations propagated by HTLM and are unreasonable large for the test here. This is because the initial perturbation chosen is a matrix of ones, which is simple  but unrealistic. For example, the screen shot attached below. 

### Issue(s) addressed
fixes #1044 

#### Added ctests for: 
1. 4dvar_htlm
2. 4dvar with identity TLM
4. test for generating coefficientsfor HTLM
5. estimating linearization errors

#### Results 
Including the results to show that the  "plumbing" works. The actual "scientific" testing would require a larger ensemble size  and the rmsbylevel fixed for taking into account the land masks.

The difference between the linearization errors using a simple identity TLM and HTLM with the ctests can be seen below:
Linearization error

There also seems to be a reasonable difference between the increments at the beginning of the window with 4dvar using these 2 TLMs
Linearization error (1)

### What compilers / HPCs was it tested with?
orion GNU
### Are the changes covered by ctests? (If not, tests should be added first.)
Yes.


kbhargava:
Wohoo!! 🎉 I finally have something working for 4dvar_htlm. Tons of thanks to @xian22.  Below is a plot showing the difference of the 2 ctests 4dvar_htlm anad 4denvar at 12 Z. I am currently only seeing a difference at surface for temperature. Elsewhere both the algorithms give some result. So obviously "science" is wrong.
 
image

tom-j-h:
It's great to see the code getting used more widely!

Your tests seem sensible to me but I have a couple of minor comments:
- In the `LinearizationError` test, you're retraining coefficients - you could instead read in the ones generated in the `GenHybridLinearModelCoeffs` test as you've done in the `Variational` test.
- Also in the `LinearizationError` test, it makes sense to use the control member of the coefficient training/the background as `x1`, but it would be better to use something other than one of the ensemble members as `x2`. Ideally, you'd use the analysis increment from the assimilation. Is this something you can do?

And in response to your comment on the RMS calculation for preconditioning:
- Whether this will impact the result depends on the relative magnitudes of the variables in the matrix which the singular value decomposition is performed on. The preconditioner scales these so that they are of similar magnitude and the matrix is no longer ill-conditioned.
- If the inclusion of the masked values affects the RMS in such a way that the scaling leaves the values in the matrix with a wide variation in magnitudes, and the matrix is therefore ill-conditioned, the SVD could be unstable and produce bad results.
- If the magnitudes are left similar enough, then the results will be fine (ignoring small differences due to numerical rounding error).
- Which of the two cases is happening here, I don't know - but the safest thing to do is definitely to change the RMS-by-level function to ignore masked values. I'd be very happy to review an `oops` PR for this. If it cannot be done generically and is a `soca` specific thing then happy to chat about what we can do to rectify it (as this would almost certainly need some changes in `oops` anyway).

Sorry for the wordy comment!

kbhargava:
Thanks Tom for the detailed response. 

> It's great to see the code getting used more widely!
> 
> Your tests seem sensible to me but I have a couple of minor comments:
> 
> * In the `LinearizationError` test, you're retraining coefficients - you could instead read in the ones generated in the `GenHybridLinearModelCoeffs` test as you've done in the `Variational` test.
Thank makes sense I can edit that. 
> * Also in the `LinearizationError` test, it makes sense to use the control member of the coefficient training/the background as `x1`, but it would be better to use something other than one of the ensemble members as `x2`. Ideally, you'd use the analysis increment from the assimilation. Is this something you can do?

Yes, that sounds like a good idea. However, one challenge with using analysis increments is that we're relying on the pseudomodel to obtain the states. The system currently can’t save forecasts from the updated initial conditions when using the pseudomodel, so we won’t have a non-linear model run for the updated state.

At the moment, the test examines realistic initial perturbations between two states that the pseudomodel provides for non-linear runs. It then compares how the specified TLM propagates the perturbation versus the non-linear model. While this approach isn't ideal, it still offers some insight into linearization errors.

> 
> And in response to your comment on the RMS calculation for preconditioning:
> 
> * Whether this will impact the result depends on the relative magnitudes of the variables in the matrix which the singular value decomposition is performed on. The preconditioner scales these so that they are of similar magnitude and the matrix is no longer ill-conditioned.


> * If the inclusion of the masked values affects the RMS in such a way that the scaling leaves the values in the matrix with a wide variation in magnitudes, and the matrix is therefore ill-conditioned, the SVD could be unstable and produce bad results.

> * If the magnitudes are left similar enough, then the results will be fine (ignoring small differences due to numerical rounding error).
> * Which of the two cases is happening here, I don't know - but the safest thing to do is definitely to change the RMS-by-level function to ignore masked values. I'd be very happy to review an `oops` PR for this. If it cannot be done generically and is a `soca` specific thing then happy to chat about what we can do to rectify it (as this would almost certainly need some changes in `oops` anyway).
> 
The variables I am testing differ by an order of magnitude of 10. Even when the land mask is ignored, the linearization errors we observe are reasonably lower. This might suggest that the variation in magnitudes of 10 doesn’t have a significant impact, which could explain why we still see reasonable, though not entirely accurate, results. However, there is a possibility that we might be getting correct answers for the wrong reasons. I can't say with confidence until the rmsByLevel function properly accounts for the land mask.

> Sorry for the wordy comment!

Thank you for taking the time to look into this in detail. I really appreciate your thoroughness.

tom-j-h:
> Yes, that sounds like a good idea. However, one challenge with using analysis increments is that we're relying on the pseudomodel to obtain the states. The system currently can’t save forecasts from the updated initial conditions when using the pseudomodel, so we won’t have a non-linear model run for the updated state.
> 
> At the moment, the test examines realistic initial perturbations between two states that the pseudomodel provides for non-linear runs. It then compares how the specified TLM propagates the perturbation versus the non-linear model. While this approach isn't ideal, it still offers some insight into linearization errors.

Okay, that's fair enough. I think many groups are in a similar position, UKMO included. I agree that what you are doing currently is reasonable enough.

> The variables I am testing differ by an order of magnitude of 10. Even when the land mask is ignored, the linearization errors we observe are reasonably lower. This might suggest that the variation in magnitudes of 10 doesn’t have a significant impact, which could explain why we still see reasonable, though not entirely accurate, results. However, there is a possibility that we might be getting correct answers for the wrong reasons. I can't say with confidence until the rmsByLevel function properly accounts for the land mask.

This is the function:
https://github.com/JCSDA-internal/oops/blob/develop/src/oops/base/Increment.h#L186-L213
I don't know how masking works with the Atlas fields. If they're not included in the "owned" field included in the `Geometry` then you're in luck, but my suspicion is they are (since "owned" is referring to all points except those in the halo). Is the land mask accessible as an Atlas field in the `Geometry` like the "owned" field is? If it is, then it should be easy to adapt that function to ignore the masked values.

> Thank you for taking the time to look into this in detail. I really appreciate your thoroughness.

You're very welcome!

travissluka:
> I don't know how masking works with the Atlas fields. If they're not included in the "owned" field included in the `Geometry` then you're in luck, but my suspicion is they are (since "owned" is referring to all points except those in the halo). Is the land mask accessible as an Atlas field in the `Geometry` like the "owned" field is? If it is, then it should be easy to adapt that function to ignore the masked values.

@tom-j-h land points are included in `owned`, since `owned` just indicates what is in the halo and what is not. For the generic oops interpolation, which handles landmasks, the code looks at the `interp_source_point_mask` metadata for the atlas field, which points to the variable name from the Geometry atlas field. I'm tagging @fmahebert so he's aware. There are plenty of places in the generic code that could make use of land masks, so I want to make sure we have a common solution, I'm assuming `interp_source_point_mask` isn't quite a correctly named field for the many things it could be used for

tom-j-h:
> @tom-j-h land points are included in `owned`, since `owned` just indicates what is in the halo and what is not. For the generic oops interpolation, which handles landmasks, the code looks at the `interp_source_point_mask` metadata for the atlas field, which points to the variable name from the Geometry atlas field. I'm tagging @fmahebert so he's aware. There are plenty of places in the generic code that could make use of land masks, so I want to make sure we have a common solution, I'm assuming `interp_source_point_mask` isn't quite a correctly named field for the many things it could be used for

Thank you for clarifying. Will wait for further comment from Francois but am happy to help where changes are needed.

fmahebert:
@travissluka @tom-j-h @kbhargava — I think there's an open design question here that we can discuss: is it true that a Field has one mask relevant  _for all generic algorithms_ (i.e. norms (like in rmsByLevel) and interpolations and ...) and that this statement holds _for all models_?

Perhaps the answer is "yes", in which case we could adopt a one-mask-per-field design across the generic code (by which I mean something like the land mask, related to the validity of a particular field; this would be in addition to mpi-related masks like "owned" and "halo" related to the execution's partition). Then we'd rename the interpolation source-point mask to this convention.

But the answer may be "no" — different generic algorithms may wish to have individual definitions of the masks they use. In this case, one approach would be for algorithm X to request a mask with metadata named "x_mask". Having fine-grained metadata doesn't mean we're duplicating the actual masking fields: multiple algorithms may request the same underlying mask through different metadata, for example, a Field's `interp_source_point_mask` metadata may point to a `land_mask` geometry field and its `norm_mask` metadata may point to `land_mask` as well. This would give flexibility for multiple masks per field and individual mask control by each generic algorithm, just requires a bit more work keeping track of the different possible metadata...

What are your thoughts on all this?

tom-j-h:
@fmahebert it's hard for me to say without knowing the details of the other generic algorithms that already exist (or are yet to exist...). But my instinct would be more towards the flexibility of your "no" answer.

travissluka:
> @fmahebert it's hard for me to say without knowing the details of the other generic algorithms that already exist (or are yet to exist...). But my instinct would be more towards the flexibility of your "no" answer.

Funny, I was about to say the opposite. I have a hard time imagining a scenario where I would want different masks for different algorithms for the same field. The closest I can imagine might involve the sea-ice thermodynamic variables. When interpolating we don't want to include the ocean grid points that don't have sea ice (because they have meaningless sea-ice thermodynamic values), but then for some stats we might want to include all non-land points. (just guessing here, we don't even have sea-ice thermodynamic variables as part of our state yet!)

@fmahebert as a compromise, perhaps if we had both a generic `mask` metadata and an algorithm-specific `x_mask` metadata that took precedence if it is present. That way we could just set `mask` and assume all the generic algorithms would work, and only set `x_mask` if there was a specific yet-to-be-imagined need for it

tom-j-h:
@travissluka thanks for the much more informed answer!

test/testinput/4dvar_identity.yml
```diff
@@ -0,0 +1,141 @@
+_date_begin: &date_begin 2018-04-15T00:00:00Z
```
travissluka, line 1: do we need the `4dvar_identity` test? It should be the same as the `3dvarfgat_pseudo` test.
I think the only difference between `cost type: 4D-Var` and `cost type: 3D-FGAT` is that `3D-FGAT` doesn't require a `linear model` to be specified, it assumes identity.

src/mains/GenHybridLinearModelCoeffs.cc
```diff
@@ -0,0 +1,17 @@
+/*
+ * (C) Copyright 2017-2021 UCAR.
```
travissluka: fix dates

src/mains/LinearizationError.cc
```diff
@@ -0,0 +1,21 @@
+/*
+ * (C) Copyright 2017-2021 UCAR.
```
travissluka: fix dates

test/testinput/linearization_error.yml
```diff
@@ -0,0 +1,172 @@
...
+      name: Identity
+  update tstep: PT6H
+  variables: [socn, tocn]
+  coefficients:
+    influence region size: 3
+    update variables: [socn, tocn]
+    time window:
+      begin: 2018-04-15T00:00:00Z
+      length: PT12H
+    calculator:
```
travissluka: invalid yaml, I'm suprised it's not complaining. I don't think you should be able to leave `calculator:` empty. comment out or set to `calculator: {}`

test/CMakeLists.txt
```diff
@@ -705,6 +729,23 @@ soca_add_test( NAME 4dhybenvar
...
+               EXE soca_var.x
+               TEST_DEPENDS test_soca_gridgen
+                            test_soca_parameters_bump_loc
+                            test_socs_parameters_diffusion
+                            test_soca_forecast_mom6 )
+
+soca_add_test( NAME 4dvar_htlm
+               EXE soca_var.x
+               TEST_DEPENDS test_soca_gridgen
+                            test_soca_parameters_bump_loc
```
travissluka:
```suggestion
```

travissluka (commented):
small comments. I am by no means an expert in htlm.

test/testinput/linearization_error.yml
xian22: It can be left empty with the code as is. The only option one might specify is an regularization scheme. Which can either be location dependent or a one size fits all regularization value.  If that optional parameter is not there the regularization value defaults to zero.

tom-j-h: This is needed because of some poor coding on my part (https://github.com/JCSDA-internal/oops/blob/0a24dfffb46ed46c5636b9b641abaf6a9e66c279/src/oops/generic/HybridLinearModelCoeffs.h#L110-L117); you need to either specify _both_ `calculator` and `ensemble` (indicating you will generate coefficients now) or just `input` (indicating you will read them in from an earlier run). The code will fail (thinking you've provided neither option) if `calculator` isn't there, even if you've got nothing to configure under it (the only thing currently is the optional regularization configuration as Christian says). Using the empty curly braces is a good idea.

fmahebert:
> @fmahebert as a compromise, perhaps if we had both a generic mask metadata and an algorithm-specific x_mask metadata that took precedence if it is present. That way we could just set mask and assume all the generic algorithms would work, and only set x_mask if there was a specific yet-to-be-imagined need for it

@travissluka this seems like a good design to me, because it's simple by default but provides a sensible recommendation for handling potential special cases.

Travis and I discussed some implementation aspects offline, I'll share them here for completeness: I've tested this metadata change in jedi-bundle, and the only trouble I can see is within soca where we're currently using two metadata for masking: `interp_source_point_mask` for the generic interpolation and `mask` for soca-specific masking, and these masks can point to different fields in the u/v field case. Before we can adopt the proposed simplification in the generic code masking, we'll want to either adapt soca code to remove the soca-specific mask, or rename the soca-specific mask to something different from the generic "mask" string. Travis and I will coordinate...

test/testinput/4dvar_identity.yml
kbhargava: I want to keep a test for 4dvar which doesn't uses HTLM just for easier debugging/code upkeep. Since the only other option is identity, I'd like to keep it here if that's okay.

travissluka (approved):

xian22 (approved):

src/mains/GenHybridLinearModelCoeffs.cc
kbhargava: Thanks for checking these, I always forget to update those. Done now :)

src/mains/LinearizationError.cc
kbhargava: done

test/testinput/linearization_error.yml
kbhargava: updated to `calculator: {}`

PR #1059, opened 2024-07-25
refactor/speed-up BkgErrFilt

travissluka:
## Description

This PR is the first of several that convert our linear variable changes into proper saber outer blocks. (Once this is merged, a cleaned-up Saber-ified parametric bkgerr stddev will be next!)

- `BkgErrFilt` is converted into a saber outer block, with no change in its behavior
- This is done purely with C++, which speeds it up since it no longer has to deal with the C++/Fortran field synchronization

> [!NOTE]
>  Do we need `BkgErrFilt`?? I think its functionality can be handled by a combination of 1) the new diffusion 
> operator (which will mask out land and thin layers) and 2) tweaking the background error stddev. Regardless, I converted it as is and we can delete it later if we don't need it.

> [!IMPORTANT]
> Dear users, there is an associated change to the yamls now. See the `3dvar.yaml` test for an example of the change. (I'll update skylab soon-ish)

## Testing

The only real answer change is the diagnostic `Jb` calculated at the end. I'm not sure why. The actual results are all the same.

There is a small performance improvement. Testing on my machine (1/4 deg grid, 1 SST and 1 ADT source, 10 iterations of 3DVAR, 24 cores ):

before (current develop)
```
OOPS_STATS Name                                                :   total (ms)   count    time/call (ms)
OOPS_STATS soca::BkgErrFilt::BkgErrFilt                        :      2074.13       1         2074.1309
OOPS_STATS soca::BkgErrFilt::multiply                          :      7305.63      12          608.8025
OOPS_STATS soca::BkgErrFilt::multiplyAD                        :      6121.60      12          510.1333
```

with this PR
```
OOPS_STATS Name                                                :   total (ms)   count    time/call (ms)
OOPS_STATS soca::BkgErrFilt::BkgErrFilt                        :        31.05       1           31.0507
OOPS_STATS soca::BkgErrFilt::multiply                          :      1423.83      24           59.3262
```


kbhargava (commented):
Do we still need [this test](https://github.com/JCSDA-internal/soca/blob/develop/test/testinput/varchange_bkgerrfilt.yml)?

kbhargava:
Apart from my 1 comment, the code looks fine and ctests run fine on orion with GNU. 🎉 Thanks @travissluka 
PS I can't wait to rerun my experiments for quarter degree given so many "speed ups" PR

kbhargava (approved):

travissluka:
> Do we still need [this test](https://github.com/JCSDA-internal/soca/blob/develop/test/testinput/varchange_bkgerrfilt.yml)?

ha, nope. It turns out that test was not setup correctly for the past 3 years, the yaml was wrong. it wasn't running `BkgErrFilt`. Oops

guillaumevernieres (approved):

PR #1062, opened 2024-08-02
Add more seaice fields to the soca state

### LINKED ISSUE #1061: Add more sea ice variables
## Description
I'm planning to add the following to the `soca`  `State`. The fields will be read from a typical CICE6 history output:

```
2D atmos
========
	float Tair_h(time, nj, ni) ;

2D air ice interface
===============
       float Tsfc_h(time, nj, ni) ;

2D aggregated seaice variables
==========================
	float hi_h(time, nj, ni) ;
	float hs_h(time, nj, ni) ;
	float aice_h(time, nj, ni) ;
	float sice_h(time, nj, ni) ;

3D seaice dynamic
===============
	float aicen_h(time, nc, nj, ni) ;
	float vicen_h(time, nc, nj, ni) ;
	float vsnon_h(time, nc, nj, ni) ;

4D seaice thermo
===============
	float Tsnz_h(time, nc, nksnow, nj, ni) ;
	float Tinz_h(time, nc, nkice, nj, ni) ;
	float Sinz_h(time, nc, nkice, nj, ni) ;

2D cice mask (probably not needed?)
==============================
	float tmask(nj, ni) ;
```
The 2D variables can be read by simply changing the `field_metadata` yaml, that should be straight forward (this is what we currently do in the gdas implementation of soca).

The 3D seaice dynamic will require some coding on the io side since the categorical variables will have to be considered as separate variables (aice_1, aice_2, ...)

The 4D seaice thermo variables will be treated as the 3D ones by removing the category dimension.

This should help the cleanup and debugging of the `Soca2Cice` variable change as well as the AI based balance.

### LINKED ISSUE #1087: Clean up existing aggregated sea-ice variables
## Description

The variables 
- `sea_ice_category_area_fraction`
- `sea_ice_category_thickness`
- `sea_ice_category_snow_thickness`

don't contain category information anymore, they are all aggregated.

- update the variable names to reflect this (i.e. `sea_ice_area_fraction`)
- change places in the soca code that try to aggregate across the 3rd dimension to assume instead they are 2D
- deprecate the sea ice concentration operator in UFO (use Identity instead)
- update the sea ice thickness operator in UFO

I want to clean up the existing aggregated variables before adding the category variables in https://github.com/JCSDA-internal/soca/pull/1062

guillaumevernieres:
This is what it looks like so far @travissluka . This is based on the assumption that we are reading CICE history files.
I have only implemented the reading of the variables with a category dimension but no level dimension so far. I figured I'd ask for feed back before I start working on the rest.

The reading part works fine but I need to work on the write method, it currently outputs the new fields as separate variables instead of reorganizing them similarly to what a cice history file looks like.

I'll probably close the draft after I get some feedback, but this work will eventually close #1061 

- closes #1061
- closes #1087


travissluka:
no major complaints, 2 nitpicks 
1) it would be nice if the existing `levels` parameter could be used instead of adding one solely for sea-ice
2) for variables with categories, it would be nice if only 1 entry had to be added to the metadata yaml, somehow using a templated name to generate the names for each category.

However, I wouldn't worry about those issues at this stage. It would be better to get everything working first then clean up those things.

guillaumevernieres:
Thanks for the feed back @travissluka , I'll definitely address your first comment and I'll think about #2 , it's going to quickly get out of control when I start adding the rest of the sea ice variables ... 
Closing this for now.

kbhargava (approved):
Passes visual check and breaks nothing for me so all good.!!

test/testinput/state_moreice.yml
```diff
@@ -0,0 +1,29 @@
+geometry:
```
travissluka, line 1: this file isn't used in any active ctest

test/CMakeLists.txt
```diff
@@ -678,6 +680,11 @@ soca_add_test( NAME 3dvar
                TEST_DEPENDS test_soca_gridgen
                             test_soca_parameters_diffusion )
 
+soca_add_test( NAME 3dvar_moreice
```
travissluka, line 683: skip 3dvar test and just do the state_moreice test for now? 3dvar seems pointless since there is no aggregation/deaggregation variable change

test/Data/fields_metadata_moreice.yml
```diff
@@ -0,0 +1,572 @@
...
+- name: sicen
+  getval name: bulk_ice_salinity
+  io file: ice
+  io name: sice_h
+  property: positive_definite
+
+# --------------------------------------------------------------------------------------------------
+# ice state variables with category dimension
+# --------------------------------------------------------------------------------------------------
+- name: aice1_h
```
travissluka: I don't like the redundancy in this file, it's error prone. If you don't mind @guillaumevernieres, could I take a stab at reworking the fields metadata code so that we can have a simpler templated configuration?, e.g.
```
name: aice<CATEGORY>_h
categories: 5
io file: ice
io sup name: aicen_h
io name: aice1_h
property: positive_definite
```

test/Data/fields_metadata_moreice.yml
```diff
@@ -0,0 +1,572 @@
...
+  getval name: sea_ice_category_1_area_fraction
```
travissluka: do we need to define a `getval name`? Is any observation operator going to pull separate categories? If the control variables for the var are going to remain the aggregated variables then I don't think so

test/testinput/state_moreice.yml
guillaumevernieres: I removed the test and forgot to remove the yaml ...

test/CMakeLists.txt
guillaumevernieres: It reads and write, so not pointless.

test/Data/fields_metadata_moreice.yml
guillaumevernieres: Sure!
I don't care too much since we make use of jinja in our workflow. I suspect our yaml templates would have some logic to avoid the repetition.

test/Data/fields_metadata_moreice.yml
guillaumevernieres: Yes, these were the first operators that we put in ufo eons ago, freeboard and concentration requiring categories as input. I suspect they still exist but I haven't checked. I guess they could be moved as variable changes too.
The long term plan is to switch to having the categories as control variables.

travissluka:
@guillaumevernieres I'll merge in develop and merge this PR after https://github.com/JCSDA-internal/soca/pull/1082 is merged (there's no need adding in internal names that are just going to get deleted right away)

travissluka:
@guillaumevernieres  I'm in the process of merging in develop and cleaning up, but now the existing `sea_ice_category_area_fraction` variable bothers me. It should be renamed `sea_ice_area_fraction` since it only stores the aggregate. Do you mind if I take this opportunity to clean that up so that we have a clear separation of aggregated and category variables with this PR?
I'd like to :
- rename the existing `sea_ice_category_area_fraction`, `sea_ice_category_thickness`, and `sea_ice_category_snow_thickness` variables
- change the places in the code that use those variables to make them 2D variables, not 3D
- deal with the obs operators. ice concentration should switch to using `Identity`, ice thickness operator will have to be updated

EDIT: I think I'll go ahead and do that in a separate PR first before emerging in this one.... once again delaying merging this PR 🙄

travissluka (approved):
I think this is ready to merge. We all done @guillaumevernieres ?

@Dooruk for your awareness, the existing ice variables have been renamed in the yamls
- `sea_ice_category_area_fraction` -> `sea_ice_area_fraction`
- `sea_ice_category_thickness` -> `sea_ice_thickness`
- `sea_ice_category_snow_thickness` -> `sea_ice_snow_thickness`

since those variables have long since been used as the aggregated ice variables.

PR #1063, opened 2024-08-02
Add an ocean smoother class

### LINKED ISSUE #1060: create an ocean smoother wrapper
## Description

create a wrapper around `oops::Diffusion` that we can use to smooth arbitrary fields. The wrapper should initialize diffusion with either
1) an existing file with scales, such as the file created by the `saber::Diffusion` calibration,
2) or via a yaml with our usual parameters: rossby radius, grid size, mixed layer depth, etc. 

The smoothing will be used by 
1) The soon-to-be-refactored parametric background error block
2) the Jacobians in the soon-to-be-refactored balance block
3) whatever else @guillaumevernieres said he wanted to smooth, I forgot

(I actually have this almost done already, just putting an issue in now as a heads up)

travissluka:
## Description
- closes #1060

Creates a class that provides a quick setup for using `oops::Diffusion` to smooth any of our fields by our usual correlation lengths (Rossby radius and min grid multiplier in the horizontal, and constant / mixed-layer-depth-based vertical)

The default parameters have been set to provide something reasonable, you can use them with the following yaml:

```yaml
horizontal: {}
vertical: {}
```

If you only want horizontal smoothing, for example, remove `vertical: {}` from the yaml.

For a full list of the yaml parameters available, see `OceanSmoother::Parameters` in the header file.

## Testing

jedi-ci-test-select=intel

A ctest has been added that creates a random increment, applies the smoothing, and saves the result, checking that the values are reasonable (e.g. smoothing all ones results in all ones)

Also, I have tested with the 1/4 deg grid and it seems reasonable (note that the intensity is lower in the Southern hemisphere because the MLD is larger there)
image

And looking at a cross-section along the tropical Pacific, the increment (bottom) is spread within the mixed layer (top)  correctly
image

This will be tested more rigorously when I use it with the refactored background error stddev and balance Jacobians.


guillaumevernieres (approved):

kbhargava (approved):

travissluka:
Oops, looks like I have a divide by zero somewhere, thanks Intel compiler

PR #1065, opened 2024-08-19
Add horizontal resolution change

### LINKED ISSUE #1064: Populate the State and Increment resolution change stubs using the oops GlobalInterpolator
## Description

In SOCA, the State and Increment constructors that do a resolution change are stubs with no implementation:
- https://github.com/JCSDA-internal/soca/blob/f71c8466375e4045087ff11119004852bc61f3c4/src/soca/State/soca_state_mod.F90#L131
- https://github.com/JCSDA-internal/soca/blob/f71c8466375e4045087ff11119004852bc61f3c4/src/soca/Increment/soca_increment_mod.F90#L192

It should now be possible to fill in this code using the `oops::GlobalInterpolator`, as was done for fv3-jedi and mpas-jedi.

## Solution

From `State.cc` and `Increment.cc`, set up and use the `oops::GlobalInterpolator` to change the resolution of the State/Increment. Follow the example from fv3-jedi and mpas-jedi:
- https://github.com/JCSDA-internal/fv3-jedi/pull/1243
- https://github.com/JCSDA-internal/mpas-jedi/pull/1002

## Alternatives

Write SOCA-specific interpolator (and add many more lines of code to SOCA 😬 )

travissluka:
## Description
- closes https://github.com/JCSDA-internal/soca/issues/1064

Implement horizontal resolution change using the generic oops code.
- added an `==` operator for `Geometry` that uses `util::getGridUid()`. Note to @fmahebert , your assumption that uid not being a global operator should be okay will likely be broken soon! (I plan testing, soonish, with a resolution change that has the same number of grid points but only distorts the equator, meaning some PEs will have identical lat/lon. We'll deal with that later though.)
- Uses the `oops::GobalInterpolator` within the resolution change for `Increment` and `State` to handle horizontal resolution change (I still need to implement a solution to vertical resolution changes)
- as a small cleanup, the `Increment::diff()` and `State::operator+=()`, now check to see if the geometry is different before creating a new (interpolated) state/increment object. Previously new state/increment were made no matter what, this eliminates the needless copy of several state/increments (kudos to copilot for suggesting the clever use of a shared pointer using a NO-OP delete function to make the code nice and clean!). E.g. the 3dvar now, when resolution change is not needed:

develop
```
48: OOPS_STATS ----------------------------------------------------------------------------------
48: OOPS_STATS --------------------------- Object counts ----------------------------------------
48: OOPS_STATS ----------------------------------------------------------------------------------
48: OOPS_STATS                                      Total  Simult. Remain    Avg (Mb)    HWM (Mb)
48: OOPS_STATS soca::State                     :       23       7
```

this PR
```
48: OOPS_STATS ----------------------------------------------------------------------------------
48: OOPS_STATS --------------------------- Object counts ----------------------------------------
48: OOPS_STATS ----------------------------------------------------------------------------------
48: OOPS_STATS                                      Total  Simult. Remain    Avg (Mb)    HWM (Mb)
48: OOPS_STATS soca::State                     :       17       6
```

- The low-res geometry landmask was changed, Using the `GlobalInterpolator` revealed the fact that our low-res ctest grid had ocean points that did not map to any ocean points in the high-res grid. This resulted in missing values after the interpolation.
- An extra check is added to `State::operator+=` to handle missing values in the increment when converting back to high res. With this, our working assumption now is that converting from high res to low res should not result in missing values in the interpolated state/increment. However, It's okay if interpolating from low-res to high-res results in missing values, it just means that there is no increment added to the state at those locations.
- Added various `_lowres` tests for 
	- creating diffusion parameters (`parameters_diffusion_lowres`)
	- dirac test with those diffusion parameters (`dirac_diffusion_lowres`)
	- `3dvar_lowres`, which uses high res grid, but low res for the inner loops
	- turned back on the convert state test (`convertstate`)

> [!WARNING]
> There is one thing not working: Our SOCA specific saber outer blocks do not work because they rely on the background, which is provided at high resolution to the saber block constructor, and I don't think I have what I need to convert the background to the low res within that constructor.

## Testing

The 3dvar using a low-res resolution change for the inner iterations seems reasonable. 

- [x] I will do a test with real-world 1/4 and 1/2 deg grid to make sure everything is working, and post plots before merging this PR


travissluka:
testing with a realistic case, looking at surface T increment, they look nearly identical (small differences in the extratropics where the length scales are very small). But, a much nicer runtime, mainly because the diffusion operator doesn't have to work so hard in the tropics

**3dvar with no resolution change (1/4 deg grid)**
image
```
OOPS_STATS Run end     - Runtime:    765.62 sec,  Memory: total:   109.26 Gb, per task: min =     4.46 Gb, max =     6.06 Gb
```

**3dvar with resolution change to 1/2 deg grid**
image
```
OOPS_STATS Run end    - Runtime:    186.46 sec,  Memory: total:    78.66 Gb, per task: min =     3.03 Gb, max =     3.58 Gb
```

travissluka:
Question for @shlyaeva and maybe @fmahebert: first, the generic interpolator seems to be working just fine for soca, yay. However, I have a problem with the soca-specific saber outer block which relies on the background https://github.com/JCSDA-internal/soca/blob/a80e20f28441c725b23c30b24aa3beffc917b744/src/soca/SaberBlocks/BkgErrFilt/BkgErrFilt.cc#L23 used by the 3dvar test with a resolution change here https://github.com/JCSDA-internal/soca/blob/a80e20f28441c725b23c30b24aa3beffc917b744/test/testinput/3dvar_lowres.yml

  the background is passed into the constructor of the saber outer block at the high resolution, and the geometry data passed into the constructor is for the low resolution grid, I don't think I have the high resolution grid available to interpolate the background. Am I missing something or not doing something right? I was poking through saber code and saw that the background does get interpolated to the dual res geometry, but that is only for ensemble covariance (right?)  The `BkgErrFilt`  block here doesn't really matter, it's going to get deleted eventually, but I do have a parametric ocean stddev block and ocean balance operator block I was implementing that relies on the values in the background.

shlyaeva:
Aha, I think you're hitting an issue of https://github.com/JCSDA-internal/oops/issues/2274: background passed to error covariances is at high res.
I guess we never hit this issue with SABER before because no one has yet run anything with SABER covariance using dual resolution where a SABER block would use the background and no interpolation blocks were explicitly used before that block. There's code in SABER that changes resolution of the background when interpolation block is used (https://github.com/JCSDA-internal/saber/blob/7b53adda2c318e0d5ca37f256d4e4f2a15f20162/src/saber/blocks/SaberOuterBlockChain.cc#L198-L203) using a following trick: if geometry is changing after going through the block, then the block's `leftInverseMultiply` can be used to change geometry for the background.
I think we need a call to interpolate the background for the case of different inner/outer loop geometries too. Let me know if you'd like me to look into a fix or if you can take it on.
Tagging @ncrossette for awareness.

travissluka:
> Aha, I think you're hitting an issue of [JCSDA-internal/oops#2274](https://github.com/JCSDA-internal/oops/issues/2274): background passed to error covariances is at high res. I guess we never hit this issue with SABER before because no one has yet run anything with SABER covariance using dual resolution where a SABER block would use the background and no interpolation blocks were explicitly used before that block. There's code in SABER that changes resolution of the background when interpolation block is used (https://github.com/JCSDA-internal/saber/blob/7b53adda2c318e0d5ca37f256d4e4f2a15f20162/src/saber/blocks/SaberOuterBlockChain.cc#L198-L203) using a following trick: if geometry is changing after going through the block, then the block's `leftInverseMultiply` can be used to change geometry for the background. I think we need a call to interpolate the background for the case of different inner/outer loop geometries too. Let me know if you'd like me to look into a fix or if you can take it on. Tagging @ncrossette for awareness.

@shlyaeva @ncrossette I'm not very comfortable with that part of the saber code, so if either of you had time to look into it that would be great. No rush though.

shlyaeva:
@travissluka can do!

kbhargava:
Tested this on orion and so far everything builds and tests fine. Can test again when we have soca specific SABER outer blcoks working again.

travissluka:
> Tested this on orion and so far everything builds and tests fine. Can test again when we have soca specific SABER outer blcoks working again.

@kbhargava The only saber bock we have right now is the `BkgErrFilt`, I don't don't really think we need to keep it around. Everything should still work fine with that one disabled

kbhargava (approved):

test/testinput/3dvar_lowres.yml
```diff
@@ -8,9 +8,9 @@ _: &land_mask
 cost function:
   cost type: 3D-Var
   time window:
-    begin: &date_begin 2018-04-14T00:00:00Z
+    begin: 2018-04-14T00:00:00Z
     length: P2D
-  analysis variables: &a_vars [socn, tocn, ssh]
+  analysis variables: &soca_vars [cicen, hicen, hsnon, socn, tocn, ssh, hocn, sw, lhf, shf, lw, us, mld, layer_depth]
```
guillaumevernieres, line 13: hocn, mld and layer_depth are not analysis variables. They should be in `state variables` below but not here.

guillaumevernieres (approved):
Looks good. Thanks for doing this @travissluka . Minor comment about state vs analysis variables that probably should be addressed in the future.

test/testinput/3dvar_lowres.yml
travissluka: yeah, i was just copying what was in the existing 3dvar, not sure why its there. I'll add this to the list of things to do whenever I do another round of general ctest cleanup.

fmahebert:
Thanks for filling this in @travissluka.

> added an == operator for Geometry that uses util::getGridUid(). Note to @fmahebert , your assumption that uid not being a global operator should be okay will likely be broken soon! (I plan testing, soonish, with a resolution change that has the same number of grid points but only distorts the equator, meaning some PEs will have identical lat/lon. We'll deal with that later though.)

We spoke about this in-person, but for posterity: GridUid is really intended to compare grids (which in atlas are global with no MPI-partition awareness), so we'll want a more robust comparison system for JEDI geometries which depend on the grid and other things too. This is on my radar.

PR #1071, opened 2024-09-13
Update tolerances for ctests on mac

ytremolet:
## Description

Updated tolerances for ctests to pass on Mac OS Sonoma with spack stack 1.8.0
Somebody should check if they make sense.

### DIFF
--- a/test/testinput/enspert.yml
+++ b/test/testinput/enspert.yml
@@ -90,4 +90,4 @@ output:
 test:
   reference filename: testref/enspert.test
   test output filename: testoutput/enspert.test
-  float relative tolerance: 1e-7
+  float relative tolerance: 1e-6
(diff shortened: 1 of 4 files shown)


ytremolet:
@srherbener FYI

xian22 (approved):
The generate coeffs test output is essentially adding the HTLM coefficients to a zero increment and printing the new increment and then continues with the STLM and new coeffs printing at each time step.  It's interesting that that this test broke but none of the other HTLM tests did, but the prints are often nonsense increments as a result of starting with a zero increment to generate test output. Since none of the other tests broke this does not worry me too much. Makes me think we should perhaps instead print direct info about the coefficients that are calculated though.

kbhargava:
> The generate coeffs test output is essentially adding the HTLM coefficients to a zero increment and printing the new increment and then continues with the STLM and new coeffs printing at each time step. It's interesting that that this test broke but none of the other HTLM tests did, but the prints are often nonsense increments as a result of starting with a zero increment to generate test output. Since none of the other tests broke this does not worry me too much. Makes me think we should perhaps instead print direct info about the coefficients that are calculated though.

@xian22 I 100% agree with that last thought. I have similar tolerances issue with intel vs GNU on other SOCA PR.

travissluka (approved):
woo, `1e7` seems "reasonable", it's at least below `1e308`

PR #1072, opened 2024-09-17
fix soca::readNcAndInterp to only read netcdf file on root PE

### LINKED ISSUE #1067: soca::readNcAndInterp unnecessarily reading netCDF file on EVERY PE
## Description
The `soca::readNcAndInterp()` function is reading the same netcdf file on every PE, causing a slowdown with high PE counts. oops.

https://github.com/JCSDA-internal/soca/blob/24e635c636d803d90e5dab0b64ad337204be0515/src/soca/Utils/readNcAndInterp.cc#L18

The netCDF file should only be read on 1 PE, and scattered to everyone else who needs it before being interpolated

travissluka:
## Description
- fixes https://github.com/JCSDA-internal/soca/issues/1067

Reading the netcdf file for the Rossby radius file was done on every PE. This caused gridgen to go slow when many PEs were used on the HPC. Not a big deal, since gridgen is only run once, but this function is going to be used in the C++ refactored saber blocks, so it was noticeable there.

This fixes the problem by only reading the netcdf file on the root PE, and broadcasting to all other PEs.

## Testing

There is no change in behavior (other than running slightly faster when lots of PEs used)

### DIFF
--- a/src/soca/Utils/readNcAndInterp.h
+++ b/src/soca/Utils/readNcAndInterp.h
@@ -17,7 +17,7 @@ namespace soca
 /// @brief Read a netCDF file and interpolate the values to the destination atlas FunctionSpace.
 ///
 /// It is assumed that variables in the input file are all 1D, and that "latitude" and "longitude"
-/// variables are present in addition to the variables listed in `vars`
+/// variables are present in addition to the variables listed in `vars`. All  resulting fields are 2D
 ///
 /// @param filename The name of the netCDF file to read.
 /// @param vars A list of variable names to read in and process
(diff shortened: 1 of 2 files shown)


guillaumevernieres (approved):
:+1:

src/soca/Utils/readNcAndInterp.h
```diff
@@ -17,7 +17,7 @@ namespace soca
 /// @brief Read a netCDF file and interpolate the values to the destination atlas FunctionSpace.
 ///
 /// It is assumed that variables in the input file are all 1D, and that "latitude" and "longitude"
-/// variables are present in addition to the variables listed in `vars`
+/// variables are present in addition to the variables listed in `vars`. All  resulting fields are 2D
```
kbhargava: coding norm failure here `Lines should be <= 100 characters long`

kbhargava (commented):
some ctests are failing on orion with GNU let me make sure it's not an error on my part as they seemto pass CI except the coding norm ctest.

kbhargava (approved):
👍🏽

PR #1073, opened 2024-09-18
remove separate u/v mask used for prints

### LINKED ISSUE #1068: get rid of the separate "mask" atlas metadata being used by soca
I previously added a separate mask to u/v so that the answers would not change during one of the large C++/Fortran refactoring PRs. However, because of this there are two masks for the u/v fields, and the `mask` metadata is used explicitly for soca specific code.
Remove this by having u/v fields use either the normal cell center mask, or the u/v specific mask. (right now interp uses the cell center mask, and RMS calculation uses the u/v specific mask)

per conversation with @fmahebert 
> for the u/v fields there is today a interp_source_point_mask metadata pointing to (effectively) the array mask_h, and a mask metadata pointing to the array mask_u/v. If we want to consolidate the generic code to use just mask for the metadata, we need to do one of two things: (a) rename soca's mask metadata to something new (so the metadata name is made available for the generic code) and preserve today's functionality, or (b) change soca's behavior so there is just one mask in use for u/v, which if I understand your prior comment may depend on destaggering the u/v fields.

Ultimately, the correct thing to do is to make sure the U/V fields are destaggered, but until that day, I need to pick a single mask to use for all generic operations.

see discussion as to why this is needed, here https://github.com/JCSDA-internal/soca/pull/1056

travissluka:
## Description

- closes https://github.com/JCSDA-internal/soca/issues/1068

In order to keep answers from changing when I did a massive C++ refactoring of fields, I added a hack where the U/V fields had a second mask (the true u/v mask) added to the atlas metadata (this was in addition to the interpolation mask, which was the T mask). So that @fmahebert can make progress with regard to having a single "mask" metadata that the generic algorithms can use, I have removed this second mask.

This only impacts the print statements for U/V fields, not any real results.

This is, again, a bit of a hack, U/V fields should not be using the cell center mask, but this will all get sorted out when u/v de-staggering is re-implemented in the variable changes.

Due to these changes, the convert state application should not be used to do a resolution change with uocn and vocn (or you could still use it, and make sure your results don't have missing values in them, the problem at this point is illegal floating point operations where the u/v and t masks don't line up... fix that some other day)

## Testing

The reference values for u and v prints for all tests had to be updated

run-ci-on-draft=true
jedi-ci-test-select=clang


kbhargava (approved):
Looks good to me.

src/soca/Fields/Fields.cc
```diff
@@ -142,9 +147,9 @@ void Fields::print(std::ostream & os) const {
     const auto & vGhost = atlas::array::make_view<int, 1>(field.functionspace().ghost());
     const auto & view = atlas::array::make_view<double, 2>(field);
     std::unique_ptr<atlas::array::ArrayView<double, 2> > mask;
-    if (field.metadata().getBool("masked")) {
+    if (field.metadata().has(MASK_METADATA)) {
```
fmahebert, line 150: There is similar logic at line ~108 above, in the `norm` method. Did you want to make the same change there? Or is there a reason I'm missing to treat `norm` differently from `print`...

travissluka: thanks, you're right, nice catch! My `grep` skills have seemingly diminished lately.

fmahebert (approved):
Not a soca expert review, but this matches what I would expect.

PR #1075, opened 2024-09-19
template the category variables

travissluka:
A more elegant way to specify the per-category variables.
See `test/Data/fields_metadata_moreice.yml`


guillaumevernieres (approved):
Thanks @travissluka

PR #1078, opened 2024-09-30
Rename Field mask metadatum

svahl991:
## Description

Changes Field metadatum key due to https://github.com/JCSDA-internal/oops/pull/2759

## Dependencies

build-group=https://github.com/JCSDA-internal/oops/pull/2759
build-group=https://github.com/JCSDA-internal/fv3-jedi/pull/1273


travissluka (approved):

travissluka:
tagging @hga007 , not sure if you're using `interp_source_point_mask` anywhere in your code as well

hga007:
@travissluka: No, I am not using the **`interp_source_point_mask`**. Thank you! I am still determining what type of mask it is. ROMS has land/sea masks at every staggering grid cell point (center, corner, east-west sides, and north-south sides). Our experience with high-resolution coastal applications is that linear or nearest-point interpolation/extrapolation of the land/sea mask is not a good idea because it can alter the river outflow in estuaries and bays, which may change the coastal buoyancy-driven currents.

Now, @fmahebert mentioned that this mask is used in oops for unstructured interpolators, which I think is or will be exclusively done with ATLAS, which only allows the A-grid (grid cell center). We specify the appropriate mask from the application geometry in such FunctionSpace. Even if we are horizontally interpolating for fine-to-coarse and vice versa, why do we need to interpolate the mask between grids? We could specify the appropriate masks for fine and coarse grids to determine the unmapped points and perform a second-step extrapolation via different methodologies (nearest point, nearest inverse distance, creep fill, or other). This topic is critical in coupled atmosphere-ocean applications with incongruent grids and land/sea masks. It gets more complicated when the atmosphere model has partial masks.

kbhargava (approved):

HamidehGMAO (approved):

fmahebert:
> Even if we are horizontally interpolating for fine-to-coarse and vice versa, why do we need to interpolate the mask between grids? We could specify the appropriate masks for fine and coarse grids to determine the unmapped points and perform a second-step extrapolation via different methodologies (nearest point, nearest inverse distance, creep fill, or other).

Hi @hga007 — I agree with you: we should be defining a resolution-appropriate mask for each Geometry, and _not_ interpolating the mask across resolutions. I think we're doing this correctly in the code, but are you aware of a counter-example?

This field `interp_source_point_mask` (about to be renamed to just `mask`) comes _from_ the Geometry, so it is in principle constant with the Geometry, and is used as an interpolation masks when interpolating Fields that _start_ at this resolution.

Let me know if that's helpful?

hga007:
@fmahebert: Thank you for the explanation. No, I cannot come up with a sound counter-example, but there is the issue of the time-dependent masks due to wetting and drying applications, where the land-sea mask due to tides or storm surges can change at every time step of the model kernel. This implies we must add the new mask to the **ATLAS** FunctionSpace at every timestep. Another application is in atmosphere models with time-evolving nested grids for tracking hurricanes and, to a lesser extent, tornados. However, these are extreme cases where data assimilation may be optional. I wouldn't worry about it.

fmahebert:
@hga007 You're right, time-dependent masks would need some new code to be properly handled. We felt it wasn't an immediate priority when first putting this together, but we can certainly revisit this if/when it becomes important!

PR #1082, opened 2024-10-10
Remove SOCA internal variable names 

travissluka:
## Description

This PR removes the field variable names internal to SOCA, and instead adopts the proper model interface names throughout the code.

For example, the temperature field uses three different variable names depending on where it is in the code:
- file IO: `Temp`
- soca internal: `tocn`
- model interface: `sea_water_potential_temperature`

`tocn` is now removed. This was done for all variables. 

> [!NOTE] 
> The model interface variable names were not changed (with the exception of `mld`) because the official CCPP names we want to adopt do not exist yet for most of the ocean variables (with the exception of `mld`, which I've changed to the CCPP standard `ocean_mixed_layer_thickness`)

## Why?
this will make future developments in SABER and VADER easier. For example, this PR (https://github.com/JCSDA-internal/vader/pull/208) has been languishing but should be easier to adopt within soca after this PR (In short, it's hard to use vader with soca the way it is right now because vader recipes require the ingredients fields to have specific names, but within soca I would have to first copy all those variables over before calling vader. The linear variable change was even more annoying because I would have to then do it again in the other direction. It was a mess)

This might also make refactoring ice variable transforms easier, as well as future re-implementation of the u/v destaggering and rotating

## Impact
File IO will still be the same, so the only places you should need to make any changes:
- Any soca variable in a yaml will have to change its name. See the tests here and the new `test/Data/fields_metadata.yml` file.
- saber: diffusion is unaffected, there _might_ be impacts to variable names in bump files (though I'm not sure, I don't use bump anymore)
- saber: if you're using the stddev saber block you might have to rename the variables in the file

## Testing
I made sure that the ctest reference answers did not change, with the exception of the state test (the analytic state random values depend on the variable name, which of course changed).

- [x] ctests pass
- [ ] update skylab
- [ ] update soca-science
- [x] Tell @guillaumevernieres and @Dooruk that that'll need to update their config files.

build-group=https://github.com/JCSDA-internal/coupling/pull/48


kbhargava:
quick couple questions: 
1. for the purpose of this PR are we ignoring anything that goes into the mysterious MLBalance stuff?
2. Do we plan on keeping older variables like tocn when they are refrenced by other variables/routines for example, `tocn_b` and so on and `soca_bkgerrgodas_tocn` I mean otherwise it would just be so looong.

travissluka:
> quick couple questions:
> 
> 1. for the purpose of this PR are we ignoring anything that goes into the mysterious MLBalance stuff?

The CI tests do not run `MLBalance`, and I'm not setup on my machine to run those tests either, so I'm not touching that code for now.

> 2. Do we plan on keeping older variables like tocn when they are refrenced by other variables/routines for example, `tocn_b` and so on and `soca_bkgerrgodas_tocn` I mean otherwise it would just be so looong.

just the variable names of the atlas fields are changing.

test/Data/fields_metadata.yml
```diff
@@ -19,143 +18,123 @@
...
+- name: sea_ice_category_thickness
   io file: ice
   io name: hicen
   property: positive_definite
 
-- name: cicen
-  getval name: sea_ice_category_area_fraction
-  getval name surface: sea_ice_area_fraction     # note: not accurate, should be "sum" not "surface"
+- name: sea_ice_category_area_fraction
+  name surface: sea_ice_area_fraction     # note: not accurate, should be "sum" not "surface"
```
guillaumevernieres, line 67: outdated comment

test/testref/convertincrement.test
```diff
@@ -1,20 +1,18 @@
...
+Trajectory state:
   Valid time: 2018-04-15T00:00:00Z
     ssh   min=-1.9244847628277935   max=0.9272826517867588   mean=-0.2767903423591662
    tocn   min=-1.8883899372702533   max=31.7004645720658580   mean=6.0175649583533595
    socn   min=10.7210460395083924   max=40.4416591897031168   mean=34.5443897261524953
    hocn   min=0.0009999999999977   max=1345.6400000000003274   mean=128.6280642065023017
 layer_depth   min=2.2854716757984130   max=5658.3057467114012979   mean=1200.5229536158392420
     mld   min=2.2854716757984130   max=4593.1533423819937525   mean=192.4109073940401515
-Output increment: 
+Output increment:
```
guillaumevernieres, line 14: why no name change here?
Also, unrelated to his PR, but we've got a nan in the reference output ... Oops.

src/soca/Fields/soca_fields_metadata_mod.F90
```diff
@@ -19,12 +19,11 @@ module soca_fields_metadata_mod
 !!
 !! Instances of these types are to be held by soca_fields_metadata
 type, public :: soca_field_metadata
-  character(len=:),  allocatable :: name     !< internal name used only by soca code
+  character(len=:),  allocatable :: name     !< name used by soca and JEDI
+  character(len=:),  allocatable :: name_surface  ! name used by UFO for the surface (if this is a 3D field)
```
guillaumevernieres: doxygen formatting missing

guillaumevernieres (commented):
Looks good, see minor comments and a nan that sneaked in while back.

test/Data/fields_metadata.yml
travissluka: nope, still accurate. `sea_ice_area_fraction` is a summation of the category variable

test/testref/convertincrement.test
travissluka: I wanted to do a PR where none (or as few as possible) of the answers change so that I know the PR was correct. The test references only check numbers, not the text, so everything still works with the old variable names. At some point, I'll regenerate all the references (but that would mean all the numbers would slightly change, I didn't want to deal with that right now)

travissluka: also, that nan is a "feature", at least that's what I'm telling myself. It indicates a mismatch between the low and high-resolution grid, so there exist valid ocean points in one grid that don't exist in the other, so the interpolation produces a nan for the grid cell. In the real world, we would take the time to make sure the source and target grids line up better.

guillaumevernieres (approved):
:+1:

kbhargava (approved):
looks good. 🎉

Dooruk (approved):
Thanks, var names are much clearer now!

PR #1086, opened 2024-10-26
Move the torch based ice emulator to soca 

### LINKED ISSUE #1014: ML based balance

guillaumevernieres:
## Description
- Moved some of the AI code from the daml repo to soca. I'm only moving the sea-ice balance, the other balances are not implemented or don't work yet.
- Added a training ctest to generate the weights of the emulator instead of storing pre-computed weights
- switched to using `libtorch` from the spack-stack, but it's unfortunately not installed as a cmake package (can't use find_package)

### Issue(s) addressed
- fixes #1014 

## Testing
**How were these changes tested?**
**Are the changes covered by ctests? (If not, tests should be added first.)**
3 ctests: training, dirac and 3dvar
It's missing a ctest for the adjoint ... 

**What compilers / HPCs was it tested with?**
Only works with `gnu`, I have no idea how to build `torch` with intel, and it's also not in the intel spack-stack
Tested on hercules only

### Arctic
bkg-incr-mlb

### Antarctic
Animated distraction for the reviewer:
incr-sst-icec


guillaumevernieres:
I broke the `cmake` logic to skip the build of `MLBalance`, oops ... I'll look into it this morning.

shlyaeva:
I pushed a small change: removed linking daml and renamed variables (removed `_category` from the name); ml tests pass on orion for me when passing -DTorch_ROOT during ecbuild step 🎉

src/soca/MLBalance/KEmul/BaseEmul.h
```diff
@@ -0,0 +1,293 @@
...
+    int batchSize_;
+    size_t epochs_;
+    std::string modelOutputFileName_;
+    std::shared_ptr<Net> model_;
+    const eckit::mpi::Comm & comm_;
+    const eckit::Configuration & config_;
+
+   public:
+    // Constructor
+    explicit BaseEmul(const eckit::Configuration & config, const eckit::mpi::Comm & comm):
```
shlyaeva: nitpick: no need for explicit here and elsewhere for ctors with more than one argument.

src/soca/MLBalance/KEmul/BaseEmul.h
```diff
@@ -0,0 +1,293 @@
...
+
+  // -----------------------------------------------------------------------------
+  /// Utilities
+  // -----------------------------------------------------------------------------
+  // -----------------------------------------------------------------------------
+  /// Emulator base class
+  // -----------------------------------------------------------------------------
+  template <typename Net>
+  class BaseEmul {
+   public:
```
shlyaeva: nitpick: perhaps these could be private?

src/soca/MLBalance/KEmul/BaseEmul.h
```diff
@@ -0,0 +1,293 @@
...
+      model_->initWeights();
+
+      // Load model if asked in the config
+      if (config_.has("ffnn.load model")) {
+        std::string modelFileName;
+        config_.get("ffnn.load model", modelFileName);
+        torch::load(model_, modelFileName);
+        model_->loadNorm(modelFileName);
+        /*
+        TODO(G): Move that to oops log debug?
```
shlyaeva: would be nice to move to debug or info or test log here and elsewhere.

src/soca/MLBalance/KEmul/BaseEmul.h
```diff
@@ -0,0 +1,293 @@
...
+      // Track all operations on the tensor to enable automatic differentiation
+      input.set_requires_grad(true);
+
+      // Train the model
+      float finalLoss(0.0);
+      oops::Log::trace() << "Train ..." << std::endl;
+      for (size_t epoch = 0; epoch < epochs_; ++epoch) {
+        // Setup the model for training
+        // TODO(G): figure out why I commented the following line
+        // model_->train();
```
shlyaeva: Perhaps because it has no effect in this case? https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.train
(I don't know, just guessing).

src/soca/MLBalance/KEmul/BaseEmul.h
```diff
@@ -0,0 +1,293 @@
...
+        for (int i = 0; i < barWidth; ++i) {
+          if (i < barLength) {
+            std::cout << "=";
+          } else {
+            std::cout << " ";
+          }
+        }
+        std::cout << "] " << std::setw(3) << static_cast<int>(percentage * 100)
+                  << "% "<< "Loss: " << loss << "\r";
+        std::cout.flush();
```
shlyaeva: would be good to move this to an oops info log and use `std::endl` (doesn't render correctly for me on orion otherwise).

src/soca/MLBalance/KEmul/IceEmul.h
```diff
@@ -0,0 +1,232 @@
...
+    prepData(const std::string& fileName, bool geoloc = false, int n = 400000) override {
+      // TODO(G): Move this elsewhere and leverage soca and/or atlas io
+      // Read additional config
+      std::string pole;
+      config_.get("domain.pole", pole);
+      bool cleanData;
+      config_.get("domain.clean data", cleanData);
+
+      // Read the patterns/targets
+      std::vector<float> lat = readCice(fileName, "ULAT");
```
shlyaeva: May be nice to move file open/close outside of this routine so it's only opening file once.

shlyaeva (commented):
Looks good to me, though I haven't played with changing parameters etc yet.
I have a few minor technical suggestions below and I don't mind making those changes myself if you'd like @guillaumevernieres.

I have one scientific question. Application of just this balance operator (K I K^T) to a dirac on ice concentration leads to reasonably looking increments in other variables, but also an increased ice concentration increment (it's no longer 1, but a value larger than 1). With a simple 2x2 matrix math I see why this happens, but it still goes against my naive intuition of what a balance operator should do. Does anyone understand whether that's desired behavior (and why if so)?

src/soca/MLBalance/KEmul/BaseEmul.h
guillaumevernieres: Ha! yes, you're right. I vaguely remember something about normalizing the input/output using a torch generic method  but I couldn't figure out how to make it work. The normalization is done separately, so that line can probably be deleted.

src/soca/MLBalance/KEmul/IceEmul.h
guillaumevernieres: good catch ... yes. 
Keep in mind that section needs to be re-written. I had access to oops but not soca when I coded this in the `daml` repo. We should leverage the io from soca and eventually add the possibility to read multiple backgrounds.

src/soca/MLBalance/KEmul/BaseEmul.h
guillaumevernieres: Damn ... OK. I'll add the get/set functions and switch the data members to private.

src/soca/MLBalance/KEmul/BaseEmul.h
guillaumevernieres: I switched to oops::Log , it seems to render properly for me ...

CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ include( soca_compiler_flags )
 ################################################################################
 # Dependencies
 ################################################################################
-find_package( NetCDF REQUIRED COMPONENTS Fortran C)
+find_package( NetCDF REQUIRED COMPONENTS Fortran C CXX)
```
travissluka: @guillaumevernieres The infra team was previously nagging us to get rid of dependency on the C++ bindings of NetCDF (The only remaining repo was soca-science), as it is not well supported by the netcdf people. Please switch to the C bindings instead, I don't want to add the C++ bindings to soca.

travissluka (changes requested):

CMakeLists.txt
guillaumevernieres: Done (pending ci status), but it's ugly. It took me a while to realize that the c netcdf binding doesn't deal with degenerate dimension implicitly, ouch.
@travissluka , @shlyaeva , don't get too worked up with the `prepData` function. I'll re-write this from scratch next time I work on this project.

src/soca/MLBalance/KEmul/BaseEmul.h
```diff
@@ -51,11 +51,39 @@ namespace soca {
...
+    void setInputSize(int inputSize) { inputSize_ = inputSize; }
+    void setOutputSize(int outputSize) { outputSize_ = outputSize; }
+    void setHiddenSize(int hiddenSize) { hiddenSize_ = hiddenSize; }
+    void setKernelSize(int kernelSize) { kernelSize_ = kernelSize; }
+    void setStride(int stride) { stride_ = stride; }
+    void setBatchSize(int batchSize) { batchSize_ = batchSize; }
+    void setEpochs(size_t epochs) { epochs_ = epochs; }
+    void setModelOutputFileName(const std::string & modelOutputFileName)
+                               { modelOutputFileName_ = modelOutputFileName; }
+    void setModel(const std::shared_ptr<Net> & model) { model_ = model; }
```
shlyaeva: I don't think you need any of the setters (which is what I was hoping for!).

shlyaeva (approved):
Thank you for addressing my comments! I agree, some code can be further improved, but OK with me to do that later.

src/soca/MLBalance/KEmul/BaseEmul.h
guillaumevernieres: Oops. Good catch. Removed.

travissluka (approved):

shlyaeva:
> I have one scientific question. Application of just this balance operator (K I K^T) to a dirac on ice concentration leads to reasonably looking increments in other variables, but also an increased ice concentration increment (it's no longer 1, but a value larger than 1). With a simple 2x2 matrix math I see why this happens, but it still goes against my naive intuition of what a balance operator should do. Does anyone understand whether that's desired behavior (and why if so)?

Following up on my question to close this thread. Here, balance operator transforms ice concentration into the balanced and unbalanced parts. The Identity covariance used in the simplified dirac test corresponds to a diagonal matrix with variance of the temperature error and variance of the _unbalanced_ ice concentration error on the diagonal (both equal to 1). Variance of the total ice concentration error (total here == balanced + unbalanced) is higher and not equal to one, so this result is expected. Good explanation of balance operators [for ocean] in [Weaver et al's ECMWF tech memo 491](https://www.ecmwf.int/sites/default/files/elibrary/2006/13007-multivariate-balance-operator-variational-ocean-data-assimilation.pdf).

PR #1088, opened 2024-10-29
Fixes for "Feature/more seaice fields ts"

travissluka:
- the logic in `does_variable_exist()` was wrong and was causing an MPI deadlock. Switched to using the built-in fms `file_exist()` and `field_exist()` instead
- The file read logic was wrong if category ice variables were present but no aggregate ice variables were present (it resulted in trying to read  an ice restart file with no registered fields, which FMS does not like apparently)
- the aggregate ice variables are renamed (`sea_ice_area_fraction`, `sea_ice_thickness` and `sea_ice_snow_thickness`) and some places in the code were changed to always assume 2D shape instead of 3D
- hofx3d test no longer uses freeboard (until we care enough to update UFO )
- the two fields_metadata files were merged (I don't want to maintain two). There was a discrepancy in the aggregate variable io names, so the old aggregated cice names in the restart files were changed to match what is in the new gdas file (`aice_h` instead of `aicen`, etc.)

This will address https://github.com/JCSDA-internal/soca/issues/1087


guillaumevernieres (approved):
Thanks for doing this @travissluka . No clue why all the tests are failing.

PR #1098, opened 2024-11-18
remove BkgErrSOCA

### LINKED ISSUE #1089: remove BkgErrSOCA
## Description

@guillaumevernieres , @Dooruk , do either of you still use the `BkgErrSOCA` linear variable change?

I would like to delete it, as people should instead be using the `StdDev` saber outer block now.

travissluka:
## Description

- closes #1089 

Removes the `BkgErrSOCA` code, which no one was using (and if they were, they should be using the saber `StdDev` block instead). That code, and the associated `util/soca_bkgerrutil_mod.F90` that it was using, was full of hardcoded variable names that I've wanted to eliminate for a while now.


shlyaeva (approved):
my favorite kind of PR 😆 
approving assuming no one uses it. Also, from quickly reviewing the removed code I agree that saber StdDev can do all of this (maybe with an exception of taking an abs value of the field on the fly which I think shouldn't be a blocker).

guillaumevernieres (approved):
RIP BkgErrSOCA.

PR #1099, opened 2024-11-19
Refactor Soca2Cice variable change to allow to process arctic and antarctic simultaneously

shlyaeva:
## Description

Refactors Soca2Cice (variable change adding soca ice concentration increment to cice restart) to allow processing for arctic and antarctic with different parameters in the same run. No effect on numerical results, just changing yamls.

Yaml changes:
- now has to specify `arctic` and `antarctic` sections
- `seaice edge`, `shuffle` and `rescale prior` yaml sections are moved into `arctic` and `antarctic`
- `region` yaml option is removed

Additionally, I moved yaml options to the Parameters, mostly to simplify the specification/assignment of default values and so all options can be documented in one place.

## Testing

Tested on orion with icepack built in the bundle (soca2cice test is only run when icepack is built), test passes.
Also tested with gdasapp low res tests: processing in one run with this branch reproduces cice restart compared to develop using two apps to process arctic and antarctic.


guillaumevernieres (approved):
Thanks @shlyaeva

Dooruk (approved):
Thanks, I won't be able to test it until we update our build to the post naming spring. Huh, so `tstep` was not functional after all..

travissluka:
@shlyaeva one suggestion:
I like the parameter class (I know certain people don't), mainly because it can keep users from doing wrong things with the yaml files. However, it has lost the ability to globally check for users putting in incorrect or unused parameters (thanks to the parameter class removal efforts going on in the other jedi repos). If you put random unused parameters in the yaml for the ctest you'll see that it runs without complaints.

I have figured out a workaround to get full yaml checking working again, at least for the soca-controlled components. An example of this is in the soca  `ParametricOceanStdDev` saber block. It involves wrapping the parameter class (see https://github.com/JCSDA-internal/soca/blob/fe20d587baf5e6e7a2f4cb2216d5f337eb76f049/src/soca/SaberBlocks/ParametricOceanStdDev/ParametricOceanStdDev.h#L138) so that it captures the full yaml config, then an explicit validation call (see https://github.com/JCSDA-internal/soca/blob/fe20d587baf5e6e7a2f4cb2216d5f337eb76f049/src/soca/SaberBlocks/ParametricOceanStdDev/ParametricOceanStdDev.cc#L37) in the class constructor.

It's the least hacky way would could come up with, short of a complete redesign of the Parameters class. But, I do want to make as much of the soca code as possible give you an error if you're doing something wrong with the yamls. Could you implement that wrapper "hack" here for this class?

travissluka (changes requested):
@shlyaeva actually, ignore everything I just wrote above! What I said doesn't exactly apply in this case since an eckit configuration is being passed into the constructor (instead of the case of the saber block where a parameter class is directly passed in). So, no hack needed, just change the deserialize in the constructor to a validate and deserialize!

guillaumevernieres:
> Thanks, I won't be able to test it until we update our build to the post naming spring. Huh, so `tstep` was not functional after all..

Made me feel good but doesn't do anything ... yaml key placebo?

shlyaeva:
@travissluka turning on validation turns out to be non-trivial (or I'm just doing something silly); I'm happy to give details over a short call if you'd like to (may be quicker to discuss). I'd still like to turn the validation on, but is it OK with you if I do it in a follow-up PR?

shlyaeva:
@travissluka: a couple of assorted side notes on Parameters and validation:

- Nice trick to get validation working for saber blocks! I like Parameters for validating yamls, since I don’t trust myself with setting up yaml correctly, especially in the fast paced development like jedi. I think it may actually be feasible to do the validation of saber block yamls in saber. My understanding is that one of the main concerns with using Parameters is the complication of the code that self-configures itself (i.e. where Configuration is used not just to get yaml options, but also to pass options that are set up in the code), and it’s acceptable to use it where Configuration/Parameters are used just for yaml options and where validation may be quite beneficial (e.g. ufo). I think saber may be a good place for a better use of parameters. I looked into this before and there were some places in saber where it was a little tricky to just switch to Parameters, but I think it should be feasible to move as much as possible to validating at saber::ErrorCovariance level. Given the complexity of saber yamls, it may be good to catch yaml errors there. If you ever are ~~ambitious enough~~ ~~bored~~ have time you can commit to the user-friendliness :) you and I could do a 2-day code sprint to improve yaml validation in saber.

- Because of the move from validating full yamls and some unresolved efficiency issues with validation, sometimes it’s not a good idea to call validateAndDeserialize instead of deserialize (longer explanation [here](https://github.com/JCSDA-internal/oops/pull/2760#issuecomment-2399808753)). However this shouldn’t be the case for validating VariableChange yamls in soca because the yaml schema for the options here is trivial and I only expect it to be created a handful of times during a run.

travissluka:
@shlyaeva ah, ok. Yeah I think I see the problem with other parameters being added in by default in the convert state executable. Oh well.

travissluka (approved):

PR #1100, opened 2024-12-02
Remove hardcoding for snow and ice levels in Soca2Cice

### LINKED ISSUE #969: Ice level is hardcoded in soca ciceutils
While using `convertstate_soca2cice` task, it is possible to change `ice_lev` within the yaml config. However, this is hard coded in the following module, so yaml `ice_lev` is not registered, and causes an issue when the `ice_level` in `iced.nc` is not 7:

https://github.com/JCSDA-internal/soca/blob/ef327de7d8f12bbc911554164a7bbdc31bf518c7/src/soca/VariableChange/Soca2Cice/soca_ciceutils_mod.F90#L27

7 might be the standard for CICE6 but I converted some of Eric Hackert's CICE4 ice restarts and they have 4 levels.

### Output:

Sad Doruk with lots of print statements

shlyaeva:
## Description

Number of snow and ice levels is now initialized based on values in yaml and not hardcoded.

### Issue(s) addressed

- fixes https://github.com/JCSDA-internal/soca/issues/969

## Testing

Orion, gnu, with icepack installed. The test doesn't cover non-default number of ice levels and snow levels.

### DIFF
--- a/src/soca/VariableChange/Soca2Cice/soca_ciceutils_mod.F90
+++ b/src/soca/VariableChange/Soca2Cice/soca_ciceutils_mod.F90
@@ -26,7 +26,7 @@ type, public :: agg_cice_state
 end type agg_cice_state
 
 type, public :: cice_state
-   integer :: ncat, ni, nj, ice_lev=7, sno_lev=1
+   integer :: ncat, ni, nj, ice_lev, sno_lev
    integer :: isc, iec, jsc, jec   ! compute domain
    integer :: isd, ied, jsd, jed   ! data domain
    logical :: initialized = .false.
@@ -60,11 +60,12 @@ contains
 
 
 ! ------------------------------------------------------------------------------
-subroutine soca_ciceutils_init(self, geom, rst_filename, rst_out_filename, global)
+subroutine soca_ciceutils_init(self, geom, rst_filename, rst_out_filename, ice_lev, sno_lev, global)
   class(cice_state),    intent(inout) :: self
   type(soca_geom), target, intent(in) :: geom
   character(len=*),        intent(in) :: rst_filename
   character(len=*),        intent(in) :: rst_out_filename
+  integer,                 intent(in) :: ice_lev, sno_lev
   logical,    optional, intent(inout) :: global
   integer(kind=4) :: ncid, dimid, varid
   logical :: isglobal
@@ -99,6 +100,9 @@ subroutine soca_ciceutils_init(self, geom, rst_filename, rst_out_filename, globa
   call nc_check(nf90_inquire_dimension(ncid, dimid, len = self%nj))
   call nc_check(nf90_close(ncid))
 
+  self%ice_lev = ice_lev
+  self%sno_lev = sno_lev
+
   ! Allocate seaice fields
   if (isglobal) then
      call self%alloc(1, self%ni, 1, self%nj)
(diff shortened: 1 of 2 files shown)


shlyaeva:
@Dooruk I noticed this issue today and decided to fix since I'm looking into Soca2Cice code in general. Do you still have a use case to test this with non-default number of ice levels?

Dooruk:
> @Dooruk I noticed this issue today and decided to fix since I'm looking into Soca2Cice code in general. Do you still have a use case to test this with non-default number of ice levels?

We also switched to 7 layers with CICE6. Previously I was converting CICE4 outputs that had 4 layers to CICE6 using a script Bin Zhao wrote so I can't test this unfortunately.

I have a somewhat related question regarding CICE6 outputs and `fields_metadata`. Are you using CICE6 history outputs in GDASapp? That's what we started using and these are the variables we use for aggregated category variables:

Screenshot 2024-12-03 at 2 32 17 PM

shlyaeva:
> I have a somewhat related question regarding CICE6 outputs and fields_metadata. Are you using CICE6 history outputs in GDASapp? 

I _think_ we're using history files for backgrounds during assimilation. Then we use restart files as background and analysis during the step that adds the increment using Soca2Cice. I'm still vague on the differences between the files so pinging @guillaumevernieres to check if it is the correct answer.
The field names in the history files from the run that I did look slightly different from yours (with `_h`, perhaps it's instantaneous vs averaged?):
```
        float hi_h(time, nj, ni) ;
                hi_h:units = "m" ;
                hi_h:long_name = "grid cell mean ice thickness" ;
                hi_h:coordinates = "TLON TLAT time" ;
                hi_h:cell_measures = "area: tarea" ;
                hi_h:comment = "ice volume per unit grid cell area" ;
                hi_h:missing_value = 1.e+30f ;
                hi_h:_FillValue = 1.e+30f ;
                hi_h:time_rep = "instantaneous" ;
        float hs_h(time, nj, ni) ;
                hs_h:units = "m" ;
                hs_h:long_name = "grid cell mean snow thickness" ;
                hs_h:coordinates = "TLON TLAT time" ;
                hs_h:cell_measures = "area: tarea" ;
                hs_h:comment = "snow volume per unit grid cell area" ;
                hs_h:missing_value = 1.e+30f ;
                hs_h:_FillValue = 1.e+30f ;
                hs_h:time_rep = "instantaneous" ;
        float aice_h(time, nj, ni) ;
                aice_h:units = "1" ;
                aice_h:long_name = "ice area  (aggregate)" ;
                aice_h:coordinates = "TLON TLAT time" ;
                aice_h:cell_measures = "area: tarea" ;
                aice_h:missing_value = 1.e+30f ;
                aice_h:_FillValue = 1.e+30f ;
                aice_h:time_rep = "instantaneous" ;
```

guillaumevernieres:
We do use history files from CICE6 and we keep our own version of fields_metadata [here](https://github.com/NOAA-EMC/GDASApp/blob/develop/parm/soca/fields_metadata.yaml) ... I suspect it's almost the same as what's in soca with some minor changes to reflect the io names of some of these variables.

guillaumevernieres (approved):
Ha! I remember @Dooruk complaining about this a while back! Thanks for doing this @shlyaeva

Dooruk:
Ah, thanks @shlyaeva. The difference does seem to be the instantaneous vs. average.

> Ha! I remember @Dooruk complaining about this a while back! Thanks for doing this @shlyaeva

I enjoy complaining 😎

Dooruk (approved):

travissluka (approved):

PR #1104, opened 2024-12-05
Bugfix for Soca2Cice (don't search when there's nothing to be found)

### LINKED ISSUE #1050: SOCA2CICE conversion segfault issue (caused by shuffle)
## UPDATE ##

Per a tip by @guillaumevernieres, it is the `shuffle` option related. Don't shuffle, folks.

## Description
Not quite sure if this should be defined as a bug or some compiler issue regarding Discover. Essentially `soca2cice` conversion is acting unpredictable for our 0.25-deg setup. @guillaumevernieres told me they haven't had any issues and Skylab uses high res setup with ocean+sea-ice (not sure if that's tested on Discover?). I can run the 5 degree SOCA tests no problem.

### To Reproduce:
What compilers/machines are you seeing this with?

**Spack-stack 1.7, NCCS Discover SLES12**

The following command, using a single node, has a success rate of about %60, runs successfully most of the time, segfault error other times. I just kept executing the same code over and over again:

`mpirun -np 24 $jedibuild/soca_convertstate.x soca2cice.yaml` 

The following, when I use multiple nodes, fails all the time (I just repeated the same command 8-10 times):

`mpirun -np 720 $jedibuild/soca_convertstate.x soca2cice.yaml`

shlyaeva:
## Description

Soca2Cice shuffle option is failing on develop with high res (https://github.com/JCSDA-internal/soca/issues/1050). This happens when KD tree search is performed on a stencil that has no background points with ice. In this case there is nothing to search from, and the accessor to the first element of an empty array fails:
https://github.com/JCSDA-internal/soca/blob/116c20eb94015ee31a480765343731ad4af77de0/src/soca/VariableChange/Soca2Cice/soca_soca2cice_mod.F90#L237
~~When this happens on develop, it often (always for my particular run) happens when analysis has veeeery small ice concentration (e.g. 4.72E-174, 9.97E-052…). I don’t think we care!~~

This PR fixes this problem by:
- skipping the search if there are no ice points in the background.
- ~~making ice concentration analysis below arbitrary 1e-4 look like no ice. I’m happy to change to something else or make yaml configurable.~~

### Issue(s) addressed

- fixes https://github.com/JCSDA-internal/soca/issues/1050

## Testing

- Tested with gdasapp high-res tests by turning on "shuffle" option (orion + intel), run fails on develop and succeeds with this fix.
-  Ran soca tests with icepack (orion + gnu). The reference slightly changed because now the few previously very low ice concentration points are zeros.


guillaumevernieres (commented):
Nice! Thanks @shlyaeva . I think icepack has a minimum threshold value for ice concentration, we should probably check against that number. I'll go do some grepping in icepack ...

shlyaeva:
@guillaumevernieres good idea! I didn't find anything in icepack (min thickness is 0.01m for thermodynamics); cice seems to have 0.001 min ice concentration for computing velocities: https://github.com/NOAA-EMC/CICE/blob/513f8d153924e18ae0f91f5440195d15210a489e/doc/source/cice_index.rst#L30

src/soca/VariableChange/Soca2Cice/soca_soca2cice_mod.F90
```diff
@@ -169,7 +169,7 @@ subroutine check_ice_bounds(self, geom, xm)
   call xm%get("sea_ice_snow_thickness",hsno_ana)
 
   ! check seaice fraction bounds
-  where (aice_ana%val<0_kind_real)
+  where (aice_ana%val<1e-4_kind_real)
```
guillaumevernieres: I couldn't find a threshold for concentration. This should probably be set to the non-zero minimum value from the background, so we don't interfere with possible ice growth from the model?

shlyaeva: We could just roll back this change. Shuffle will take care of it: it will keep concentration at zero if there's no ice around (with the bugfix in this PR), and take whatever is closest if there is ice around (which for very low concentrations is likely to be zero in the background anyway). What do you think?

guillaumevernieres: OK. Rolling back sound fine to me.

shlyaeva:
I've pushed an update and verified that soca test is passing. I'll rerun the high-res experiment for sanity check (I'm pretty sure it works with just the remaining change), and report back.

Dooruk:
I can test highres on Discover if necessary

What is the purpose of the shuffle option?

guillaumevernieres:
> I can test highres on Discover if necessary
> 
> What is the purpose of the shuffle option?

add seaice where the analysis has ice but the background has none.

shlyaeva:
> add seaice where the analysis has ice but the background has none.

more precisely, shuffle in* seaice where analysis has ice, and background's ice is smaller than `seaice edge` in yaml.

*replace by sea ice background state from one of the nearby gridpoints that has total ice concentration closest to what analysis should be.

> I can test highres on Discover if necessary

that would be great, thank you! I just lost my setup.

shlyaeva:
Confirming that gdasapp high res case runs with the shuffle option on and the updated fix.

travissluka (approved):

Dooruk (approved):
It works on my end too 👍 

This is not related to this PR (more towards Anna's the previous PR) but could there be applications where `arctic` and `antarctic` domains for soca2cice used separately?

guillaumevernieres (approved):
:+1:

shlyaeva:
> This is not related to this PR (more towards Anna's the previous PR) but could there be applications where arctic and antarctic domains for soca2cice used separately?

Maybe regional?

Dooruk:
> Maybe regional?

Right but when I accidentally didn't have the `antarctic` section as well as `arctic` in the yaml it failed. Maybe I had other issues so I can test again.

PR #1105, opened 2024-12-11
Add option to skip nonlinear variable transform

danholdaway:
## Description

Companion PR for https://github.com/JCSDA-internal/oops/pull/2789 where the nonlinear variable transform is part of the algorithm when a linear variable change is needed around the background error model.

For SOCA this is not really needed at the moment. Though if full use of Vader were to be implemented it likely would require more thought. For now we adapt to the changes by adding an option to simply add new fields to the state, set them to zero and skip the variable transform. No impact on results.

### Issue(s) addressed

N/A

## Dependencies

build-group=https://github.com/JCSDA-internal/oops/pull/2789


ncrossette:
is there a change needed in this yaml?

https://github.com/JCSDA-internal/soca/blob/feature/lvc_refactor/test/testinput/parametric_stddev.yml

danholdaway:
> is there a change needed in this yaml?
> 
> https://github.com/JCSDA-internal/soca/blob/feature/lvc_refactor/test/testinput/parametric_stddev.yml

I don't think so as this code does not appear to be invoking the background error model.

PR #1111, opened 2024-12-27
Improve Soca2Cice (use icepack to cleanup after rescaling)

### LINKED ISSUE #993: More engineering fixes required for the CICE initialization
## Description
There are a few more cases for which the CICE model forecast becomes unstable.

## Solution
Something needs to be done about snow enthalpy and small sea-ice concentration values. 
Solution from Philip Pegion:

1. Removal of small sea-ice

```python
import xarray as xr
import numpy as np
import sys
import shutil

infile=sys.argv[1]
ds_in=xr.open_dataset(infile)

var_list=['vicen','vsnon','Tsfcn','iage','alvl','vlvl', 'apnd', 'hpnd', 'ipnd',\
 'dhs', 'ffrac', 'sice001', 'qice001', 'sice002', 'qice002', 'sice003', 'qice003',\
 'sice004', 'qice004', 'sice005', 'qice005', 'sice006', 'qice006', 'sice007', 'qice007','qsno001','aicen']
for var in var_list:
    tmparr=np.where(ds_in['aicen'].values >0.15,ds_in[var],0.0)
    ds_in[var][:]=tmparr


new_frac=tmparr.sum(axis=0)
new_mask=np.where(new_frac > 0, 1.0,0.0)
ds_in['iceumask'][:]=new_mask
ds_in.to_netcdf('trimmed_'+infile)
ds_in.close()
shutil.move(infile,'old_'+infile)
shutil.move('trimmed_'+infile,infile)
```
2. Limit snow enthalpy
```python
import xarray as xr
import numpy as np
import sys

infile=sys.argv[1]
anl=xr.open_dataset('anl_'+infile)
rhos      = 330.0
cp_ice    = 2106.
c1        = 1.0
Lsub      = 2.835e6
Lvap      = 2.501e6
Lfresh=Lsub - Lvap
rnslyr=1.0
puny=1.0E-012

# icepack formulate for snow temperature
A = c1 / (rhos * cp_ice)
B = Lfresh / cp_ice
zTsn = A * anl['qsno001'][:].values + B
# icepack formula for max snow tempature
Tmax = -anl['qsno001'][:].values*puny*rnslyr /(rhos*cp_ice*anl['vsnon'][:].values)

# enthlap at max now tempetarure
Qmax=rhos*cp_ice*(Tmax-Lfresh/cp_ice)

# fill in new enthalpy where snow temperature is too high
newq=np.where(zTsn <= Tmax,anl['qsno001'][:].values,Qmax)
newf=np.where(anl['vicen'] > 0.00001,anl['aicen'][:].values,0.0)

# fill in snow enthalpy (0) where there is no snow
newq2=np.where(anl['vsnon'][:]==0.0,anl['qsno001'][:].values,newq)
anl['qsno001'][:]=newq2
anl['aicen'][:]=newf
# write out file
anl.to_netcdf(infile)
```
The above features need to be added to the `soca2cice` var change.

shlyaeva:
## Description

There are several improvements to Soca2Cice here:

- **use icepack's `cleanup_itd` after rescaling** to rebin thickness categories if necessary, remove very small ice areas and remove snow if snow enthalpy indicates out-of-bounds temperatures (see details in `cleanup_itd`).
Notes:
  - Ice thickness categories are set up automatically by `icepack_init_itd`. I am not sure this is always the correct way to set up ice categories; does anyone know? It's important for rebinning.
  - I have removed the existing adjustment of snow enthalpy from the code since I expect it's now covered (in a different way) by `cleanup_itd`.
  - The rebinning uses the model's time step. I added a new yaml option `icepack time step` with pseudorandom default 300.
  - If `cleanup_itd` indicates that icepack aborted, the code will abort. This can be replaced (or optionally replaced) by shuffling if the `cleanup_itd` aborts.
  - Only ice/snow temperature, ice and snow enthalpies and ice salinity are currently updated in the `cleanup_itd`, but I can add more tracers (e.g. melt ponds?)
- **exclude backgrounds that are equal to `sea ice edge` value from rescaling**. Previously the code both shuffled and rescaled where background == `sea ice edge`. This fix is mostly to cover an edge case when `sea ice edge` is set to zero. In this case, when the background is zero, we definitely don't want to rescale (i.e. divide by zero)
- **explicitly cover zero analysis ice concentration case**. I think this was not covered before (`cleanup_ice` on develop covers the case where the restart fields updated after shuffling and rescaling are zero, but doesn't cover the case where they should have been updated to zero).
Note: I kept Guillaume's update to the ice/snow temperature as icepack's `liquidus_temperature_mush`. 
- **add `shuffle stencil size` as a yaml option (default is 9)** to add some flexibility for testing
- removed some unused variables.

### Issue(s) addressed

fixes https://github.com/JCSDA-internal/soca/issues/993 in a roundabout way (instead of adding the fixes suggested here, `cleanup_itd` is used to do that and more)

## Testing

Tested on orion with gnu and unit tests, the current test mostly rebins categories up and down, and zaps small areas on a few occasions.

Tested on hera with intel with high res gdas 3DVar experiment, setting ice edge to 0.15 and running a few cycles. `cleanup_itd` mostly rebins categories and zaps snow with low snow volumes and zero snow enthalpies.

### DIFF
--- a/src/soca/VariableChange/Soca2Cice/soca_soca2cice_mod.F90
+++ b/src/soca/VariableChange/Soca2Cice/soca_soca2cice_mod.F90
@@ -11,12 +11,12 @@ use fckit_exception_module, only: fckit_exception
 use fckit_mpi_module, only: fckit_mpi_comm
 use kinds, only: kind_real
 
-use icepack_itd
-use icepack_mushy_physics, only: liquidus_temperature_mush
-use icepack_therm_shared, only: l_brine, icepack_ice_temperature
+use icepack_itd, only: icepack_init_itd, cleanup_itd
+use icepack_warnings, only: icepack_warnings_flush
+use icepack_tracers, only: icepack_init_tracer_indices, nt_tsfc, nt_qice, nt_qsno, nt_sice
 use icepack_parameters, only: icepack_init_parameters, icepack_recompute_constants
 use icepack_parameters, only: ktherm, heat_capacity
-use icepack_parameters, only: rhos, Lfresh, cp_ice
+use icepack_therm_shared, only: l_brine
 
 use soca_geom_mod, only: soca_geom
 use soca_state_mod, only: soca_state
@@ -269,12 +269,22 @@ subroutine cleanup_ice(self, geom, xm)
   type(soca_geom), target, intent(in)  :: geom
   type(soca_state),      intent(inout) :: xm
 
-  integer :: i, j, k, n, n_src
+  integer :: i, j, k, n, n_src, ntracers
+  integer :: nt_tsfc_in, nt_qice_in, nt_qsno_in, nt_sice_in
+  real(kind=kind_real) :: aice, aice0
   type(soca_field), pointer :: t_ana, s_ana, aice_ana, hice_ana, hsno_ana
-
+  real(kind=kind_real) :: dt
   real(kind=kind_real), allocatable :: h_bounds(:)
-  real(kind=kind_real), allocatable :: zTin(:), zTsn(:), temp_sno_test
+  real(kind=kind_real), allocatable :: tracers(:,:)   ! (ntracers, ncat)
+  logical, allocatable :: first_ice(:)                ! (ncat) ! For bgc and S tracers. set to true if zapping ice.
+  integer, allocatable :: trcr_depend(:)              ! (ntracers), = 0 for aicen tracers, 1 for vicen, 2 for vsnon
+  real(kind=kind_real), allocatable :: trcr_base(:,:) ! (ntracers, 3);  = 0 or 1 depending on tracer dependency
+                                                      ! argument 2:  (1) aice, (2) vice, (3) vsno
+  integer, allocatable :: n_trcr_strata(:)            ! number of underlying tracer layers
+  integer, allocatable :: nt_strata(:,:)              ! indices of underlying tracer layers
+  real(kind=kind_real), allocatable :: fiso_ocn(:)    ! isotope flux to ocean     (kg/m^2/s)
 
+  dt = 900.0 ! randomly chosen
   ! pointers to soca fields (most likely an analysis)
   call xm%get("sea_water_potential_temperature",t_ana)
   call xm%get("sea_water_salinity",s_ana)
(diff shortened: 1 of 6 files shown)


src/soca/VariableChange/Soca2Cice/soca_soca2cice_mod.F90
```diff
@@ -285,69 +293,75 @@ subroutine cleanup_ice(self, geom, xm)
...
+                         ! number of aerosol tracers, bio tracers, bio layers
+                         0, 0, 0, &
+                         ! aerosol flag, topo pond flag, heat capacity flag, flag for zapping ice for bgc and s tracers
+                         .false., .false., .true., first_ice, &
+                         ! tracer indices and sizes used in rebinning
+                         trcr_depend, trcr_base, n_trcr_strata, nt_strata, &
+                         fiso_ocn=fiso_ocn)
+        call icepack_warnings_flush(6)
+        if (icepack_warnings_aborted()) then
+           call abor1_ftn("Soca2Cice: icepack aborted during cleanup_itd")
```
guillaumevernieres, line 354: I suspect this conditional is going to be a "rare  event"? Or have you seen failures already at that stage? Is this where you are planning to shuffle in the future?

guillaumevernieres (approved):
Looks good. Thanks @shlyaeva .

src/soca/VariableChange/Soca2Cice/soca_soca2cice_mod.F90
shlyaeva: I haven't seen this fail on this condition yet, so hoping it is a rare event. Yes, I think that's where it would make sense to do shuffle!

shlyaeva:
Per offline discussion with @guillaumevernieres, I'll add the melt ponds tracers to cleanup_itd too.

travissluka (approved):

guillaumevernieres:
> Per offline discussion with @guillaumevernieres, I'll add the melt ponds tracers to cleanup_itd too.

That, or a `TODO`/issue would be fine as well, up to you.

shlyaeva:
@travissluka thank you! I've opened an issue: https://github.com/JCSDA-internal/soca/issues/1115; I'll address it in a follow-up PR if that's good with everyone.

PR #1114, opened 2025-01-07
Add optional increment output in Soca2Cice

shlyaeva:
## Description

Adds an option to output the postprocessed sea ice increment from Soca2Cice (useful for comparing the increment that DA/soca generated to the increment that we actually could add to the model). I'm not super satisfied with where/how it's implemented in the code, but the good news it's optional 😆 . 

## Testing

Tested on orion with gnu (ctests) and on hera with intel within gdasapp workflow.

An example of output:
soca increment (no postprocessing):
aice_h_North

postprocessed soca2cice increment (shuffle on, sea ice edge = 0.4) -- used in cp4 experiment
aice_h_North

out of interest, postprocessed soca2cice increment (shuffle off, sea ice edge = 0.8) -- used before cp4 experiments
aice_h_North

### DIFF
--- a/src/soca/VariableChange/Soca2Cice/Soca2Cice.cc
+++ b/src/soca/VariableChange/Soca2Cice/Soca2Cice.cc
@@ -20,16 +20,15 @@ static VariableChangeMaker<Soca2Cice>
 // -----------------------------------------------------------------------------
 
 Soca2Cice::Soca2Cice(const Geometry & geom,
-                             const eckit::Configuration & conf)
+                     const eckit::Configuration & conf)
   : geom_(geom)
 {
   util::Timer timer("soca::Soca2Cice", "Soca2Cice");
 
-  Parameters params;
-  params.deserialize(conf);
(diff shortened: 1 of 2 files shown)


guillaumevernieres (approved):
:+1:

src/soca/VariableChange/Soca2Cice/Soca2Cice.cc
```diff
@@ -46,6 +45,22 @@ void Soca2Cice::changeVar(const State & xin, State & xout) const
   xout = xin;
   soca_soca2cice_changevar_f90(keySoca2Cice_, geom_.toFortran(),
                                xin.toFortran(), xout.toFortran());
+  if (params_.incOutput.value() != boost::none) {
```
travissluka, line 48: fyi, for future reference, you can just do `if(params_incOutput.value())`, no need for the explicit comparison

travissluka (approved):

PR #1116, opened 2025-01-08
move icepack to a submodule, and enable icepack in CI

### LINKED ISSUE #968: CI is not using icepack
icepack is optional, but is enabled in the TravisCI tests ( or at least it was supposed to have been enabled, see https://github.com/JCSDA-internal/soca/pull/967).

It seems icepack is not being built with the new CI. Turn it on.

travissluka:
## Description

Icepack is now a submodule of soca, instead of using a custom fork of the repo in the bundle, the same as what was done for `mom6`. I used version `1.2.5` of icepack, which let me compile without changing the soca code. Updating to a newer version (1.5.0 is the latest) is outside the scope of this PR

Using git submodules for this is ok assuming the following conditions remain true:
- SOCA is the only repo in JEDI using icepack. This will probably remain true, unless we want to move some functionality into vader.
- no changes are needed to icepack other than cmake wrappers (now true since we are not using private interfaces anymore)
- we don't intend to update icepack often (every now and then when there is a new release, but we can't automatically keep up with develop)

If any of those are ever false, we should revert back to a fork of the repo.

---

Other things that were changed:
- tests using icepack are now enabled in CI!
- mom6 and icepack were both moved to the `external/` directory, to tidy things up
- all of the `if(icepack_FOUND)` cmake statements were removed. ice is important, so we now assume that icepack is always available! 🧊 🌊 🐧 ❄️ 
- the `icepack` library built here is named `icepack_limited` because only the files needed by soca are compiled. Keep this in mind if you ever need to use other icepack routines, you might have to enable other files in `icepack_files.cmake`

---

~After this PR is merged, separate PRs will be issued to remove icepack from the CI and the jedi-bundle, and our icepack fork can be archived.~ <- nevermind, I forgot icepack was not being used by the CI until now!

- closes https://github.com/JCSDA-internal/soca/issues/968


travissluka:
@eap there is something weird going on with the CI here. There are 4 instances now of the unit test "running", but the integration test is skipped because "prior step failed". The first CI unit test says it has been "running" for 17.5 hours now.

I'm not sure what's going on since I cant look at any of the build output

shlyaeva (commented):
@travissluka 🎉 you rock! looks good to my non-expert eye. I'm going to build and also check what needs to be done to adapt in gdasapp, I'll be back to approve once ready.

shlyaeva (approved):
Looks good to me, and I've prepared our workflow for this change. Thank you @travissluka 🎉

travissluka:
pinging @Dooruk in case anything needs to be changed in your build scripts

Dooruk:
I gave this a try and had some issues as we use our own `jedi_bundle` wrapper.  I'm attending AMS this week so won't be able to troubleshoot until next week but I don't want to hold this back.

Dooruk:
~Hmm I fixed the `jedi_bundle` bug but now I'm encountering saber related errors. I may have missed a PR or change related to SABER or SOCA but not sure what might be the reason. Is it obvious to anyone?~

solution -> fresh clone

shlyaeva:
no rush from my side to get this merged, definitely can wait till next week or even the week after. 
@Dooruk not sure about the saber error; but sometimes in the past I needed to wipe out saber build directory for things to build.

Dooruk (approved):
Well, a fresh clone worked. All good 👍

PR #1118, opened 2025-01-13
Bugfixes for icepack cleanup in Soca2Cice

### LINKED ISSUE #1117: bugs in soca2cice call of icepack cleanup
## Description

There's a bug in how tracers are passed to cleanup_itd, resulting in messed up tracers after shifting ice between categories.
The good news is that there's another bug in how cleanup_itd is used in soca2cice: the tracers are not put back into the analysis fields after cleanup, so it doesn't matter that they are messed up. 🤦‍♀

shlyaeva:
## Description

Bugfixes for icepack cleanup (#1117) for tracers (ice/snow temperature, ice and snow enthalpies, ice salinity):
- fixes problems with tracer index data passed to cleanup_itd (no need to compare to the previous code here, it's all wrong; the new code follows roughly what's done in icepack [here](https://github.com/JCSDA-internal/Icepack/blob/d1a42fb142033ca8c82a3f440ed38c63d992a314/configuration/driver/icedrv_init.F90#L1140-L1209) (you'll also see why I don't want to deal with melt ponds right now :))
- also fixes the cancelling bug: tracers updated after cleanup_itd are now copied back to analysis.

This time, besides running one high res forecast (it didn't fail), I also looked at how ice/snow temperature, ice and snow enthalpies and ice salinity changed with cleanup (spoiler alert: the code no longer generate 90C ice temperatures just to ignore them later). Cleanup redistributes ice between the categories if needed (and updates tracers accordingly), and zaps ice with small areas and snow when snow enthalpies are inconsistent.

All of the below plots are showing update to different ice categories from just cleanup (after the analysis increment was added); so the ice concentration difference is just redistribution of ice between the categories. Note that it's the first cycle and background is quite far from analysis, so the increment was quite large in a lot of areas, so there's a lot to redistribute.

cice_inc_aice
cice_inc_qsnow
cice_inc_tmp

### Issue(s) addressed

Fixes #1117 

## Testing

hera/intel with soca tests (I needed an update to tolerance similar to #1116 for the test to pass on develop) and one analysis/forecast for high-res gdas experiments.


travissluka (approved):

guillaumevernieres (approved):
:+1:

PR #1122, opened 2025-01-16
Bugfix to allow read/write of "many" files

### LINKED ISSUE #1121: bug in fms_io_init causing issues when writing multiple files
Our LETKF runs with 80 members fail with fms io error
```
FATAL from PE    14: fms_io(setup_one_field), 1: max_domains exceeded, needs increasing
```
It's exposing a bug in `fms_io_init`, and is fixable by making sure we only call `fms_io_init`/`fms_io_exit` once per geometry (instead of on each read/write like we do now). I'll issue a PR soon to fix.

shlyaeva:
## Description

LETKF with 80 members was choking on fms_io with an error:
```
FATAL from PE    14: fms_io(setup_one_field), 1: max_domains exceeded, needs increasing
```
This is because FMS has a bug in `fms_io_init` that gets exposed if you run `fms_io_init`/`fms_io_exit` one too many times (one over 100 to be precise). Due to the bug `fms_io` increases the domain counter every read/write call even though it's the same domain every time. 
This PR changes soca to call `fms_io_init`/`fms_io_exit` once per `Geometry` to avoid hitting this bug (I think it's the way to use these subroutines intended by developers anyway). 

### Issue(s) addressed

- fixes #1121 

## Testing

hera/intel:
- soca CI: change enspert test to generate 120 members; fails with the above error on develop, runs fine with this branch
- gdasapp high-res LETKF with 80 members: fails with the above error on develop, runs fine with this branch

### DIFF
--- a/src/soca/Fields/soca_fields_mod.F90
+++ b/src/soca/Fields/soca_fields_mod.F90
@@ -25,7 +25,7 @@ use oops_variables_mod, only: oops_variables
 use tools_const, only: deg2rad
 
 ! MOM6 / FMS modules
-use fms_io_mod, only: fms_io_init, fms_io_exit, register_restart_field, &
+use fms_io_mod, only: register_restart_field, &
                       restart_file_type, restore_state, free_restart_type, save_restart, &
                       file_exist, field_exist
 use fms_mod,    only: write_data, set_domain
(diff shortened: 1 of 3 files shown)


guillaumevernieres (approved):
:tada: ... Hopping I did not jinx the ci!

travissluka (approved):
oops, @kbhargava and I saw this bug several months ago and just ignored it.

PR #1127, opened 2025-01-21
Bugfix for the bugfix for fms io

### LINKED ISSUE #1123: Bug in bugfix for fms i/o
## Description

@RussTreadon-NOAA discovered that gdasapp marine var test started failing after merging #1122 (I conveniently relied on soca tests and high res LETKF to test, and didn't run those tests).

While I'm just starting to look at the issue, I suspect the problem is in how `fms_io_init`/`fms_io_exit` are called: in #1122 I left only one pair in `Geometry` ctor and dtor. This works fine as long as there's one geometry, but is problematic when there are several geometries with overlapping lifetimes. E.g. in Var, the first geometry will get created at the beginning, `fms_io_init` gets called; then the inner loop geometry gets created, another `fms_io_init` gets called (already not great), then inner loop geometry gets destroyed, `fms_io_exit` is called, and essentially no fms output calls will work after that, although the outer loop geometry is still in scope.

I need advice on solution.
A few ideas:
- somehow call `fms_io_init`/`fms_io_exit` (and other model init?) only once before and after JEDI application run
- have a global counter of geometries that keeps track of how many times init/exit were called from geometries, and only calls it on the first geometry ctor and last dtor.

@travissluka, @guillaumevernieres what do you think?

In the meantime, if we don't have a fix soon, it may be wise to rollback #1122.

shlyaeva:
## Description

Bugfix for a bugfix.. You can see where it's going next if you let me continue contributing to soca!
See details in #1123 

### Issue(s) addressed

fixes #1123 

## Testing

orion/gnu (only soca tests)

### DIFF
--- a/src/soca/Geometry/soca_geom_mod.F90
+++ b/src/soca/Geometry/soca_geom_mod.F90
@@ -53,6 +53,9 @@ private
 ! oops!
 real(kind=kind_real), parameter :: INVALID_HALO = -999_kind_real
 
+!> counter for geometries, used to call fms_io
+integer, save, private :: global_soca_geom_counter = 0
+
 ! ------------------------------------------------------------------------------
 !> Geometry data structure
 type, public :: soca_geom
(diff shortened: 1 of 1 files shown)


guillaumevernieres (approved):
Thanks @shlyaeva for the quick turn around! I have not tested the changes, but lgtm.

travissluka (approved):

Issue #1126, opened 2025-01-21
Remove Fortran copy of field values.


travissluka:
Previously, the authoritative copy of the fields was moved into an Atlas Fieldset on the C++ side. However, the fortran `soca_field` contains it's own copy within `val`, so each State/Increment uses 2x the memory it should be using. Some Fortran code has been refactored into C++, but the remaining code that uses the Fortran copy of the fields needs to be refactored.

1. Do a search for `sync_to_atlas()` and `sync_from_atlas()`, all fortran interface functions that use those calls use the fortran internal copy of the fields and needs to be refactored (there are 24 calls to `sync_to_atlas()`, so this is going to be fun)
2. remove those sync functions, and `soca_field::val(:,:,:)`
3. admire the drop in memory usage in celebrate.

Things that need refactoring

- [x] `Model2GeoVaLs`
- [x] `Balance`
- [x] `Soca2Cice`
- [x] `Increment`, `State`, `Fields`

PR #1128, opened 2025-01-22
Remove usage of Fortran copy of field data, part 1

travissluka:
## Description
 Part 1 of 3 to address
- https://github.com/JCSDA-internal/soca/issues/1126

This refactors `Model2GeoVaLs` and `Balance` so that:
- the Atlas fieldset is used on the Fortran side for accessing field data, no longer using the Fortran-only internal `field%val` structure
- The `sync_to_atlas()` and `sync_from_atlas()` calls in the interface for these classes were removed, which should make these classes go faster
- Also, I fixed the loop order in `Balance`, it was wrong, and there were switch statements inside the loops, eeek. Should be faster now

2 more PRs after this before we can remove the Fortran-only internal field data and free up some memory!

## Testing

There are no change in any answers.

I also tested with the 1/4 deg, 1/2 deg minimization on my computer, 24 cores, there are no changes in answers and runtime for these components is faster (not much, and would be even less with more PEs, but every little bit counts)

develop branch
```
OOPS_STATS Name                                                :   total (ms)   count    time/call (ms)
OOPS_STATS soca::Balance::Balance                              :      2835.24       1         2835.2442
OOPS_STATS soca::Balance::multiply                             :      5452.57      12          454.3808
OOPS_STATS soca::Balance::multiplyAD                           :      4556.61      12          379.7176
OOPS_STATS soca::Model2GeoVaLs::changeVar                      :      1482.66       3          494.2216
```

this branch
```
OOPS_STATS Name                                                :   total (ms)   count    time/call (ms)
OOPS_STATS soca::Balance::Balance                              :      2231.90       1         2231.8992
OOPS_STATS soca::Balance::multiply                             :       448.99      12           37.4157
OOPS_STATS soca::Balance::multiplyAD                           :       467.96      12           38.9964
OOPS_STATS soca::Model2GeoVaLs::changeVar                      :         3.82       3            1.2744
```

### DIFF
--- a/src/soca/LinearVariableChange/Balance/soca_balance_mod.F90
+++ b/src/soca/LinearVariableChange/Balance/soca_balance_mod.F90
@@ -472,45 +472,74 @@ subroutine soca_balance_multinvad(self, dxa, dxm)
   type(soca_increment), target, intent(inout) :: dxm !< output increment
   type(soca_increment), target, intent(in)    :: dxa !< input increment
 
-  integer :: i, j, n
-  type(soca_field), pointer :: fld_a, fld_m
-  type(soca_field), pointer :: socn_a, ssh_a, cice_a
+  integer :: i, j, k, n, idx
 
-  cice_a => null()
+  type(atlas_field) :: fld_a, fld_m, socn_a, ssh_a, cice_a
+  real(kind=kind_real), pointer :: data_a(:,:), data_m(:,:), data_socn(:,:)
+  real(kind=kind_real), pointer :: data_ssh(:,:), data_cice(:,:) => null()
 
-  call dxa%get("sea_water_salinity", socn_a)
-  call dxa%get("sea_surface_height_above_geoid",  ssh_a)
-  if (dxa%has("sea_ice_area_fraction")) call dxa%get("sea_ice_area_fraction",cice_a)
+  socn_a = dxa%afieldset%field("sea_water_salinity")
+  ssh_a = dxa%afieldset%field("sea_surface_height_above_geoid")
+  call socn_a%data(data_socn)
+  call ssh_a%data(data_ssh)
+  if (dxa%afieldset%has("sea_ice_area_fraction")) then
+    cice_a = dxa%afieldset%field("sea_ice_area_fraction")
+    call cice_a%data(data_cice)
+  end if
 
-  do n = 1, size(dxm%fields)
-    fld_m => dxm%fields(n)
-    fld_a => dxa%fields(n)
+  do n = 1, dxm%afieldset%size()
+    fld_m = dxm%afieldset%field(n)
+    fld_a = dxa%afieldset%field(n)
+    call fld_m%data(data_m)
+    call fld_a%data(data_a)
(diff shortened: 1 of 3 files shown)


src/soca/LinearVariableChange/Balance/soca_balance_mod.F90
```diff
@@ -227,49 +252,79 @@ subroutine soca_balance_mult(self, dxa, dxm)
...
-        case ("sea_ice_area_fraction") ! Ice fraction
-            fld_m%val(i,j,1) = fld_a%val(i,j,1) + &
-              & self%kct(i,j) * tocn_a%val(i,j,1)
+    case ("sea_surface_height_above_geoid")
+      do j = self%geom%jsc, self%geom%jec
+        do i = self%geom%isc, self%geom%iec
+          idx = self%geom%atlas_ij2idx(i,j)
+          data_m(1, idx) = data_a(1, idx)
+          do k = 1, tocn_a%shape(1)
+            data_m(1, idx) = data_m(1, idx) + &
```
shlyaeva, line 301: Double-checking since I don't know this code (and don't feel like thinking, I guess): both in develop and here is it intended to have `data_m = data_m + ...`? (I see all other variables updated as `data_m = data_a + ...`.

src/soca/LinearVariableChange/Balance/soca_balance_mod.F90
```diff
@@ -282,44 +337,73 @@ subroutine soca_balance_multad(self, dxa, dxm)
...
+          idx = self%geom%atlas_ij2idx(i,j)
+          do k = 1, fld_a%shape(1)
+            data_a(k, idx) = data_m(k, idx) + &
+              self%kst%jacobian(i,j,k) * data_socn(k, idx) + &
+              self%ksshts%kssht(i,j,k) * data_ssh(1, idx)
+          end do
+          if (associated(data_cice)) then
+            data_a(1, idx) = data_a(1, idx) + &
+              self%kct(i,j) * data_cice(1, idx)
+          end if
```
shlyaeva, line 393: similar question here: intended to have `data_a = data_a + ...` as opposed to `data_a = data_m + ...`

guillaumevernieres (approved):
nice :+1:

shlyaeva (approved):
🎉 looks good to me!
a couple of unrelated naive questions for my understanding of this code in the comments below.

src/soca/LinearVariableChange/Balance/soca_balance_mod.F90
travissluka: yes, it's different here because 3D variables are combined into a 2D variable. There is an initial `data_m` = `data_a` on line 299  (2D) then the loop is a summation over 3D data putting into the 2D `data_m`

travissluka: ... in fact, when I refactored this i did have `data_m = data_a +` at first and couldn't figure out  why it was giving completely wrong answers !

PR #1134, opened 2025-02-11
Update icepack version

### LINKED ISSUE #1119: Update icepack version
It would be good to update icepack version used with soca to a more recent one (e.g. one that's currently used in UFS).

I looked into what changes are made in this fork in the tag that [EMC is currently using](https://github.com/JCSDA-internal/Icepack/commits/73136ee8dcdbe378821e540488a5980a03d8abe6/), and the only needed changes are for ecbuild, so updating technically should be fairly easy (pick a commit off the CICE-Consortium/Icepack and cherry-pick ecbuild changes on top). 

First, is that something others are OK with, and do we need to discuss which CICE-Consortium/Icepack tag to update to?

Second, how do we do this better? Manually update more often? (I don't mind doing this myself every once in a while). Forget about this repo, the ecbuild wrappers and somehow build without them? (I don't know how).

@guillaumevernieres, @travissluka, @Dooruk.

shlyaeva:
## Description

Updates icepack hash to [3792520](https://github.com/cice-consortium/icepack/tree/3792520561cf9419082ef41f9f0dffd03edf2e43) and changes the calling code in Soca2Cice accordingly.

For the low res soca2cice test here, the only changes are in the ice/snow surface temperature. That's not surprising, given that `icepack_itd` now additionally takes freezing temperature (computed now in Soca2Cice as `icepack_liquidus_temperature`, similar to how @guillaumevernieres coded it for zero ice case).

### Issue(s) addressed

Fixes https://github.com/JCSDA-internal/soca/issues/1119

## Testing

orion/gnu; ctest in soca.
TBD: intel/hera; high res case

### DIFF
--- a/src/soca/VariableChange/Soca2Cice/soca_ciceutils_mod.F90
+++ b/src/soca/VariableChange/Soca2Cice/soca_ciceutils_mod.F90
@@ -8,7 +8,7 @@ module soca_ciceutils_mod
 use kinds, only: kind_real
 use netcdf
 use fckit_mpi_module, only: fckit_mpi_comm
-use icepack_itd
+use icepack_itd, only: aggregate_area
 use soca_utils, only: nc_check
 use soca_geom_mod, only: soca_geom
 use mpp_domains_mod, only : mpp_update_domains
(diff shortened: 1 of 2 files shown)


guillaumevernieres (approved):
Thanks @shlyaeva

travissluka (approved):

PR #1143, opened 2025-03-27
Make soca2cice more robust

shlyaeva:
## Description

Coupled model and/or CICE doesn't love it when there's absurdly thin ice in some categories. CICE fails in this case, usually with picard nonconvergence failure.

This PR makes the initial conditions more robust wrt this feature by removing ice when ice volume is small (fix proposed and used by Phil Pegion previously). Ideally we'd move this ice to a different category instead of just zapping it, but this solution is good enough for now.

More details in https://github.com/NOAA-EMC/GDASApp/issues/1575

## Testing

Tested on hera with intel:
- soca2cice test in soca.
- high-res coupled experiments. More than 20/30 ensemble forecasts were failing with picard nonconvergence when using initial conditions post-processed with soca2cice on develop. All of them are running fine with this fix. The ice concentration differences are minimal compared between the two branches.

### DIFF
--- a/src/soca/VariableChange/Soca2Cice/soca_soca2cice_mod.F90
+++ b/src/soca/VariableChange/Soca2Cice/soca_soca2cice_mod.F90
@@ -272,6 +272,7 @@ end subroutine shuffle_ice
 !> clean-up the CICE state
 !!
 subroutine cleanup_ice(self, geom, xm)
+  use fckit_log_module,   only: fckit_log
   class(soca_soca2cice), intent(inout) :: self
   type(soca_geom), target, intent(in)  :: geom
   type(soca_state),      intent(inout) :: xm
@@ -290,6 +291,8 @@ subroutine cleanup_ice(self, geom, xm)
 
   type(atlas_field) :: tocn, socn, aice, hice, hsno
   real(kind=kind_real), pointer :: data_tocn(:,:), data_socn(:,:), data_aice(:,:), data_hice(:,:), data_hsno(:,:)
+  character(255) :: msg
+  integer :: count_thinice
 
   ! get fields from atlas
   tocn = xm%afieldset%field("sea_water_potential_temperature")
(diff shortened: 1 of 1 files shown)


src/soca/VariableChange/Soca2Cice/soca_soca2cice_mod.F90
```diff
@@ -391,12 +395,34 @@ subroutine cleanup_ice(self, geom, xm)
         if (icepack_warnings_aborted()) then
            call abor1_ftn("Soca2Cice: icepack aborted during cleanup_itd")
         endif
+        ! remove ice if ice volume is less than 0.00001: empirical hack
```
guillaumevernieres, line 398: HOOOoooo!!!! I'm shocked! :scream:

guillaumevernieres (approved):
Thanks for the quick fix @shlyaeva !

src/soca/VariableChange/Soca2Cice/soca_soca2cice_mod.F90
```diff
@@ -391,12 +395,34 @@ subroutine cleanup_ice(self, geom, xm)
...
+            self%cice%tsfcn(i,j,k) = Tf
+          endif
+        enddo
         ! re-compute aggregates = analysis that is effectively inserted in the restart
         data_aice(1, idx) = sum(self%cice%aicen(i,j,:))
         data_hice(1, idx) = sum(self%cice%vicen(i,j,:))
         data_hsno(1, idx) = sum(self%cice%vsnon(i,j,:))
      end do
   end do
+  if (count_thinice > 0) then
```
travissluka, line 421: how often do we expect this to happen? If a lot, would this flood the log and would it be better to add a flag somewhere to turn off the warning?

shlyaeva: That's why I'm only printing it once [per mpi task] with total stats, and not every time it happens. So unless you're running with thousands of MPI tasks, shouldn't be too bad. I can also switch to another log that would only print on task 0 if it's a concern.

shlyaeva: I also don't mind just removing the log altogether, let me know.

travissluka: ah yep. (just realized i forgot my coffee this morning)

travissluka (approved):

PR #1161, opened 2025-05-20
Bugfix for Soca2Cice: use correct mask to mask out points on land

### LINKED ISSUE #1160: Soca2Cice is using incorrect mask to remove ice over land
## Description

Soca2Cice is effectively removing ice close to the coast, especially evident in the Canadian Archipelago in winter. This is because it's using umask from the restart files to remove ice "over land" (except it turns out it's not really over land).

I'm working on a fix.

shlyaeva:
## Description

Soca2Cice on develop is effectively removing ice close to the coast, especially evident in the Canadian Archipelago in winter. This is because it's using umask from the restart files to remove ice "over land" (except it turns out it's not really over land).

This is fixing it by using the mask from the geometry to mask out points on land. Some code that is no longer used is removed.

### Issue(s) addressed

Fixes #1160 

Test from a winter high-res experiment:
CICE restart background:
bg
Raw SOCA increment:
soca_inc
CICE restart analysis (result of ConvertState using Soca2Cice and increment above) - CICE restart background, on develop (note coastlines pattern):
restan_restbg_diff_dev
CICE restart analysis - CICE restart background, with this branch:
restan_restbg_diff_fix

### DIFF
--- a/src/soca/VariableChange/Soca2Cice/soca_ciceutils_mod.F90
+++ b/src/soca/VariableChange/Soca2Cice/soca_ciceutils_mod.F90
@@ -17,14 +17,6 @@ use mpp_mod, only : mpp_gather, mpp_root_pe, mpp_pe
 implicit none
 private
 
-type, public :: agg_cice_state
-   real(kind=kind_real),  allocatable :: aice(:)
-   real(kind=kind_real),  allocatable :: lon(:)
-   real(kind=kind_real),  allocatable :: lat(:)
-   integer,  allocatable :: ij(:,:)
-   integer :: n_src
-end type agg_cice_state
-
 type, public :: cice_state
    integer :: ncat, ni, nj, ice_lev, sno_lev
    integer :: isc, iec, jsc, jec   ! compute domain
@@ -43,9 +35,7 @@ type, public :: cice_state
    real(kind=kind_real),  allocatable :: qice(:,:,:,:)   ! volume-weighted ice enthalpy
    real(kind=kind_real),  allocatable :: qsno(:,:,:,:)   ! volume-weighted snow enthalpy
    real(kind=kind_real),  allocatable :: sice(:,:,:,:)   ! volume-weighted ice bulk salinity
-   real(kind=kind_real),  allocatable :: iceumask(:,:)
    real(kind=kind_real),  allocatable :: aice(:,:)       ! total ice concentration
-   type(agg_cice_state):: agg
 contains
   procedure :: init => soca_ciceutils_init
   procedure :: alloc => soca_ciceutils_alloc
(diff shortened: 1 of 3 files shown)


shlyaeva:
Note: the `convertstate_soca2cice` test output has changed slightly, but the output files show no difference. I don't understand why the test output in the log has changed.

guillaumevernieres (approved):
I'm a bugvloper.
Thanks @shlyaeva !

src/soca/VariableChange/Soca2Cice/soca_ciceutils_mod.F90
```diff
@@ -301,25 +277,9 @@ subroutine soca_ciceutils_read(self, geom)
...
-           call aggregate_area (self%aicen(i,j,:), &
-                                self%aice(i,j), aice0)
-           self%agg%lon(cnt) = geom%lon(i,j)
-           self%agg%lat(cnt) = geom%lat(i,j)
-           self%agg%aice(cnt) = self%aice(i,j)
-           self%agg%ij(1,cnt) = i
-           self%agg%ij(2,cnt) = j
-           cnt = cnt + 1
-        end if
+       call aggregate_area(self%aicen(i,j,:), self%aice(i,j), aice0)
```
shlyaeva, line 282: I understand why the test reference changed. This computation (total ice concentration) used to happen only for the values that were not masked by the incorrect mask, and now it's happening for all gridpoints. The result of this computation is used in shuffling. On develop, incorrectly, some points that were eligible for shuffling did not have correct values. That's now resolved.

Tested this hypothesis by using develop and moving this call outside of the if statement; the test reference changed to what is used in this PR.

Dooruk (approved):

Dooruk:
So, I tested my sandbox 3dvar diffusion setup with 5 different JEDI builds, with the same adt, insitu, sst, and sea-ice (north & south) observations. Essentially everything is identical except the JEDI builds from:

February 
March
April
May x2 (latest build today)

My `norm reduction`s or `residual norms` in OOPS logs are identical between these builds and I see @shlyaeva made quite some changes to seaice part and I see test references are changing. Are my results expected?

**Edit:** After thinking for more than 10 seconds, I guess most if not all changes are in the soca2cice part?

shlyaeva:
@Dooruk since soca2cice is only used in postprocessing after Var, I wouldn't expect the norm reduction in var to be affected by this, which is consistent with your runs.

Dooruk:
> @Dooruk since soca2cice is only used in postprocessing after Var, I wouldn't expect the norm reduction in var to be affected by this, which is consistent with your runs.

Thanks, that occurred to me as I pressed send. I guess all of your changes have been pertaining the postprocessing part then, right? Over the past few months*

shlyaeva:
@Dooruk yes

PR #1162, opened 2025-05-29
Change halo sizes to match localization lengthscales

shlyaeva:
## Description

https://github.com/JCSDA-internal/ufo/pull/3710 adds a check on halo size and localization lengthscale to avoid user errors of specifying halo sizes that are smaller than localization lengthscale.

This fixes soca tests to specify halo size that is large enough. Test references have changed since now LETKF assimilates all the obs that are supposed to be assimilated, unlike on develop.


bhuang95 (approved):
Looks good. Thanks!

guillaumevernieres (approved):

kbhargava (approved):
looks good to me 🎉

PR #1169, opened 2025-06-06
Adjust SST in Soca2Cice

shlyaeva:
## Description

Adjusts SST when adding ice increment if `update SST` is set to true (it is by default now):
- if ice is removed, average current SST with SST from the nearby gridpoints without ice, and limit the increment to `max positive SST update` (1C by default)
- if ice is added, make SST a weighted average of current SST with freezing temperature (weights are based on ice concentration).

A new exciting way to do coupled covariances 🩹 

## Testing

I'll post results from the high-res runs once ready.

### DIFF
--- a/src/soca/VariableChange/Soca2Cice/Soca2Cice.h
+++ b/src/soca/VariableChange/Soca2Cice/Soca2Cice.h
@@ -51,6 +51,11 @@ class Soca2Cice: public VariableChangeBase {
     oops::Parameter<RescaleParameters> rescale{"rescale prior",
       "Option to rescale sea ice in the ice pack zone (where ice concentration"
       " > seaice edge)", {}, this};
+    oops::Parameter<bool> adjustSST{"update SST",
+      "Option to update sea surface temperature", true, this};
+    oops::Parameter<double> sstDiffMax{"max positive SST update",
+      "Maximum update to sea surface temperature (K)", 1.0, this,
+      {oops::minConstraint(0.0)}};
   };
 
   // ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


src/soca/VariableChange/Soca2Cice/Soca2Cice.h
```diff
@@ -51,6 +51,11 @@ class Soca2Cice: public VariableChangeBase {
     oops::Parameter<RescaleParameters> rescale{"rescale prior",
       "Option to rescale sea ice in the ice pack zone (where ice concentration"
       " > seaice edge)", {}, this};
+    oops::Parameter<bool> adjustSST{"update SST",
```
guillaumevernieres, line 54: :fire_extinguisher:

guillaumevernieres (approved):
:+1:

shlyaeva:
I ran a 3 day experiment using the updated Soca2Cice both for 3dVar increment, and for the recentering increments. Here's a representative example of changes to SST:
postprocessed ice increment:
Screenshot 2025-06-11 at 6 36 20 AM
postprocessed SST increment (includes changes here and other SST postprocessing changes that @guillaumevernieres added recently):
Screenshot 2025-06-11 at 6 36 27 AM

At least for this period it's mostly cooling temperatures where ice is added in the analysis. The max absolute difference is 0.4 degrees, the scale on the plot is: [-0.1, 0]

kbhargava (approved):
Thanks @shlyaeva ! Haven't tested the said feature but the code looks good and builds fine without breaking anything :)

shlyaeva:
Thank you so much for merging @kbhargava !

PR #1171, opened 2025-06-21
Update LETKF test reference

shlyaeva:
## Description

Updates LETKF test reference.

The results changed due to https://github.com/JCSDA-internal/oops/pull/2874 where the order of H(x) computations has changed. Prior to https://github.com/JCSDA-internal/oops/pull/2874 H(x) for ensemble members was computed before H(x) for ensemble mean, and after https://github.com/JCSDA-internal/oops/pull/2874 it's the opposite.  `letkf_split` tests in soca use the `Thinning` obs filter. The behavior of randomization depends on the order of the `Thinning` filter calls, so after https://github.com/JCSDA-internal/oops/pull/2874 the thinning for the ensemble mean (that's used for LETKF QC) has changed. I have verified that that's the reason for changed results by hacking `ufo::Thinning` to reset the seed at every call, in this case results prior and post https://github.com/JCSDA-internal/oops/pull/2874 are unchanged.

This may require https://github.com/JCSDA-internal/oops/pull/2891 (to be updated).

### DIFF
--- a/test/testref/letkf_split_solver.test
+++ b/test/testref/letkf_split_solver.test
@@ -86,24 +86,24 @@ Background mean :
 molar_concentration_of_biomass_in_sea_water_in_p_units   min=0.0000000000000000   max=0.0000001872794279   mean=0.0000000065402047
 Analysis mean :
   Valid time: 2018-04-15T00:00:00Z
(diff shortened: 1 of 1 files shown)


shlyaeva:
This fixes a failure of soca test on develop.

guillaumevernieres (approved):
:+1:

ytremolet (approved):

PR #1174, opened 2025-07-02
Fix bug in ML Balance when B resolution differs from background

### LINKED ISSUE #1124: MLBalance returns NaN's when not using the model's resolution for B
## Description
There is something wrong in the MLBalance saber block. It returns NaN's in the first inner iteration. I suspect something related to the trajectory. 

### To Reproduce:
Run a variational application using the MLBalance block at a different resolution than the deterministic.

guillaumevernieres:
## Description
I was not checking for ghost and mask, which resulted in NaN's when running with backgrounds and **B** at different resolutions.

### Issue(s) addressed
- fixes #1124 

## Testing

How were these changes tested? **ctests and HR on HPC**
What compilers / HPCs was it tested with? **GNU**
Are the changes covered by ctests? **YES**

### DIFF
--- a/src/soca/MLBalance/MLJac.h
+++ b/src/soca/MLBalance/MLJac.h
@@ -60,7 +60,7 @@ namespace soca {
       const int nnodes = xb["sea_water_potential_temperature"].shape(0);
       for (atlas::idx_t jnode = 0; jnode < nnodes; ++jnode) {
         if ((ghost(jnode)) | (mask(jnode, 0) == 0) |
-            ( abs(lonlat(jnode, 1)) <= 40.0 )) continue;
+            (abs(lonlat(jnode, 1)) <= 40.0)) continue;
         pattern[0] = tair(jnode, 0);
         pattern[1] = tsfc(jnode, 0);
         pattern[2] = sst(jnode, 0);
--- a/src/soca/MLBalance/MLJac.h
+++ b/src/soca/MLBalance/MLJac.h
@@ -10,6 +10,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+        #include <iostream>
 
 #include "torch/torch.h"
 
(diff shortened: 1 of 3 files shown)


src/soca/MLBalance/MLJac.h
```diff
@@ -10,6 +10,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+        #include <iostream>
```
shlyaeva:
```suggestion
#include <iostream>
```

src/soca/MLBalance/MLJac.h
```diff
@@ -57,6 +59,8 @@ namespace soca {
       torch::Tensor pattern = torch::zeros({iceEmulArctic_.getInputSize()});
       const int nnodes = xb["sea_water_potential_temperature"].shape(0);
       for (atlas::idx_t jnode = 0; jnode < nnodes; ++jnode) {
+        if ((ghost(jnode)) | (mask(jnode, 0) == 0) |
+            (abs(lonlat(jnode, 1)) <= 40.0)) continue;
```
shlyaeva, line 63: I think the check should be both for arctic and antarctic, or could possibly be removed. I see code below that checks for arctic vs antarctic, but I'm not sure what happens when the node is in [-40, 40].

guillaumevernieres: Resurrecting an old branch and I clearly didn't pay attention. Good catch.

guillaumevernieres: It's skipping latitudes where we shouldn't find seaice and the mask/ghost nodes. Seems right to me but I'll re-run this at high res. I'm not convinced I resurrected the correct branch.

shlyaeva: oh, I missed abs in this code, sorry.

shlyaeva (approved):
Looks good to me, I misread the code earlier.

Dooruk (approved):
I trust you that this works seamlessly (famous last words).

src/soca/MLBalance/MLJac.h
guillaumevernieres: > oh, I missed abs in this code, sorry.

:scream:

PR #1179, opened 2025-07-18
Add 3D geometry iterator option

shlyaeva:
## Description

Added at this point for profiling LETKF with vertical localization, may be used in the future.

### DIFF
--- a/src/soca/Geometry/Geometry.cc
+++ b/src/soca/Geometry/Geometry.cc
@@ -150,23 +150,17 @@ namespace soca {
   // -----------------------------------------------------------------------------
 
   GeometryIterator Geometry::begin() const {
-    ASSERT(IteratorDimension() == 2);  // Modification will be needed for 3D.
-                                       // We don't use 3D right now
-
     // find the first non ghost point
     const auto & ghost = atlas::array::make_view<int, 1>(functionSpace_.ghost());
     size_t idx = 0;
     while (idx < ghost.size() && ghost(idx)) idx++;
-    return GeometryIterator(*this, idx, -1 );
+    return GeometryIterator(*this, idx, 0);
   }
 
   // -----------------------------------------------------------------------------
 
   GeometryIterator Geometry::end() const {
-    ASSERT(IteratorDimension() == 2);  // Modification will be needed for 3D.
-                                       // We don't use 3D right now
-
-    return GeometryIterator(*this, functionSpace_.size(), -1);
+    return GeometryIterator(*this, functionSpace_.size(), 0);
   }
 
   // -----------------------------------------------------------------------------
--- a/src/soca/Increment/Increment.cc
+++ b/src/soca/Increment/Increment.cc
@@ -325,7 +325,6 @@ namespace soca {
       } else if (geom_.IteratorDimension() == 3) {
         if (view.shape(1) > iter.k()) {
           // 3D case, only add if this variable has this level
-          // (e.g. don't add surface variables for below-surface iterators))
           values.push_back(view(iter.i(), iter.k()));
         }
       }
--- a/src/soca/Increment/Increment.cc
+++ b/src/soca/Increment/Increment.cc
@@ -290,24 +290,44 @@ namespace soca {
   // -----------------------------------------------------------------------------
 
   oops::LocalIncrement Increment::getLocal(const GeometryIterator & iter) const {
-    ASSERT(geom_.IteratorDimension() == 2);  // Change will be needed for 3D.
-                                             // We don't use 3D right now.
     std::vector<int> varlens(vars_.size());
-
     // count space needed
-    size_t idx = 0;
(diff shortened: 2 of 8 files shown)


frolovsa:
ha, i forgot that we never implemented 3D iterator for ocean letkf. 

At the time i was thinking about implementing a GETKF vertical localization for the ocean. we could still do that as well. The eigen vecotrs will just have to change in horizontal space. This is now supported bcause VerticalLocalization is actually a state set of 3D fields. Practically, we would have to lift the vertical localization code from ocean var and use it to compute vertical localization matrix for each vertical column, then decompose it into eigenvectors and save it in vertical localization. This could either be done on the fly or saved on disk.

guillaumevernieres (approved):
:+1: 

Could that PR be a good opportunity to ask our favorite `llm` to properly document `GeometryIterator.h` ?

src/soca/Geometry/Geometry.cc
```diff
@@ -150,23 +150,17 @@ namespace soca {
...
   GeometryIterator Geometry::begin() const {
-    ASSERT(IteratorDimension() == 2);  // Modification will be needed for 3D.
-                                       // We don't use 3D right now
-
     // find the first non ghost point
     const auto & ghost = atlas::array::make_view<int, 1>(functionSpace_.ghost());
     size_t idx = 0;
     while (idx < ghost.size() && ghost(idx)) idx++;
-    return GeometryIterator(*this, idx, -1 );
+    return GeometryIterator(*this, idx, 0);
```
frolovsa, line 157: could you comment why 0 makes sense now?

src/soca/Increment/Increment.cc
```diff
@@ -290,24 +290,44 @@ namespace soca {
...
-        values.push_back(view(iter.i(), lvl));
+      if (geom_.IteratorDimension() == 2) {
+        // 2D case, iterate over levels
+        for (size_t lvl = 0; lvl < view.shape(1); lvl++) {
+          values.push_back(view(iter.i(), lvl));
+        }
+      } else if (geom_.IteratorDimension() == 3) {
+        if (view.shape(1) > iter.k()) {
+          // 3D case, only add if this variable has this level
+          // (e.g. don't add surface variables for below-surface iterators))
```
frolovsa: "below surface" might not be the right word here?

do we need any special logic for levels that are below bathymetry?

test/testinput/obslocalization.yml
```diff
@@ -24,7 +24,7 @@ observations:
     reference gridpoints:
       lons: [-172.5, -192.5, -172.5, -17.5]
       lats: [  42.5,    2.5,    2.5,  22.5]
-      depths: [-99999.0, -99999.0, -99999.0, -99999.0]
+      depths: [1, 1, 1, 1]
```
frolovsa, line 27: are we testing obsloc for 2d and 3d? or just 3d?

frolovsa (commented):
Thank you Anna,

I have a couple of questions. 

Nothing really that requires the code change that i can see

src/soca/Geometry/Geometry.cc
shlyaeva: Previously, index in vertical was always set to -1, and its value was never used. Now for 3D iterator the vertical index goes between 0 and nlev-1, and for 2D iterator it can still be any value (0 works just as well as -1).
3D iterator starts iterating vertically from zero (up to nlev-1, and then to the next point, restarting from zero). 
So for the `begin` iterator, we use the first valid gridpoint and 0-th level. The `end` iterator needs to be just past the last gridpoint's last level, so the horizontal index is the same as before (gridpoint past last), and the vertical index is 0.

src/soca/Increment/Increment.cc
shlyaeva: I'll change the wording, thanks for catching.

> do we need any special logic for levels that are below bathymetry?

I don't know, maybe? @guillaumevernieres? Either way, this would be a subject for some future PR, when we use this functionality. This is added just for profiling tests at this point.

test/testinput/obslocalization.yml
shlyaeva: Just for 2D at this point.

src/soca/Increment/Increment.cc
guillaumevernieres: @frolovsa , @shlyaeva , We don't have levels below bathymetry in any of our configurations (so far). If I remember correctly a z type coordinate would have the "unused" layers/levels set to 0 thickness (or a tiny number).

shlyaeva:
I've added documentation to GeometryIterator class, and some other suggestions from Claude.

src/soca/Geometry/Geometry.cc
frolovsa: sounds good. I just wanted to know what was your logic

frolovsa (approved):
Looks good to me. thank you.

PR #1182, opened 2025-07-28
Updates to iterator for local solvers

shlyaeva:
## Description

Updates iterator to allow for GETKF, and adds some optimization.

- changes `getLocal`/`setLocal` to always fill `nlevs` values for each variable (for 2D variables everything below level 1 is filled with zero);
- implements `Geometry::verticalCoord` that is used in GETKF for "levels" only;
- moves some computations from `getLocal` to the `Increment` ctor, this speeds up `getLocal`/`setLocal` (noticeable for LETKF with vertical localization);
- adds tests for vertical obs localization and GETKF.

This allows to run LETKF with vert localization (still slow) and GETKF; both with simple vertical coordinate of "levels".

## Testing

Tested on hera with intel: 1/4 degree tests with LETKF with and without vertical localization and GETKF.
New tests added for CI as well.


src/soca/Geometry/Geometry.cc
```diff
@@ -170,6 +171,15 @@ namespace soca {
     soca_geo_get_num_levels_f90(toFortran(), vars, lvls.size(), lvls.data());
     return lvls;
   }
+
+  // -----------------------------------------------------------------------------
+  std::vector<double> Geometry::verticalCoord(std::string &) const {
```
frolovsa: do you want to put a check if the vertical coordinate is in levels then it produces this input. if it is not in levels that it produces an error?

test/CMakeLists.txt
```diff
@@ -62,6 +62,7 @@ set( soca_test_input
   testinput/geometry_iterator_2d.yml
   testinput/geometry_iterator_3d.yml
   testinput/geometry.yml
+  testinput/getkf.yml
```
frolovsa, line 66: should we also specify a 3d letkf test case? users over start their applications from these ctests examples.

frolovsa (dismissed):
Thank you as always for your excellent contributions. 

I learned about the view object! 

I made two suggestions. 

Side note. i always thought it would be cool to specify localization distance in terms of water density. but this would require for geometry to know about the state at that iterator. But in GETKF verticalEV is a set of 3D fields. I guess we could implement this in the VerticalEV class somehow.

src/soca/Geometry/Geometry.cc
shlyaeva: Good idea, done!

test/CMakeLists.txt
shlyaeva: I've added that as well.

shlyaeva:
I've updated this PR to address Sergey's comments and all tests are passing now.
@bhuang95 would you be comfortable re-reviewing instead of Sergey and checking that his comments are addressed while he's unavailable?

bhuang95 (approved):
Updated PR addresses Sergey's comments and looks good according to the PR description. Thanks for adding this feature. Approved it.

shlyaeva:
Thank you @bhuang95! @guillaumevernieres or @Dooruk do you mind providing a second review when you have time?

shlyaeva:
@guillaumevernieres or @Dooruk could you please provide a second review?

Dooruk (approved):
I can approve this not to hold it back but I'm not super familiar with the code here.  @guillaumevernieres would provide a substantive review.

PR #1187, opened 2025-08-27
Bugfix for ice thickness and snow thickness adjustment

shlyaeva:
## Description

Update for soca2cice code to fix an issue with ice and snow thickness adjustment.

This was noticed by Nick Szapiro in https://github.com/NOAA-EMC/GDASApp/issues/1868#issuecomment-3222018558.

Plots below are for ice concentration with: background restart, soca analysis, result of soca2cice on develop, result of soca2cice with the fix (plots are from the results of soca2cice test):
background_arctic_ice_thickness_m outputan_arctic_hi_h analysis_arctic_ice_thickness_m analysisnew3_arctic_ice_thickness_m

With the changes in this branch, as expected, in the ice pack where rescaling is happening, ice thickness now matches what analysis is.

I also ran the test replacing soca analysis with what's in the background to mimic what we do in GFSv17 (no ice thickness updates).
background_arctic_ice_thickness_m outputnew_arctic_hi_h analysisnew5_arctic_ice_thickness_m analysisnew_arctic_ice_thickness_m

Again, with changes in this branch ice thickness is correct.

### DIFF
--- a/src/soca/VariableChange/Soca2Cice/README.md
+++ b/src/soca/VariableChange/Soca2Cice/README.md
@@ -32,7 +32,7 @@ Sea ice in CICE is represented using multiple thickness categories (`ncat`), whi
 - Operates in the ice pack (where concentration > `seaice_edge`) 
 - Rescales the background ice distribution to match the analyzed aggregated values
 - Preserves the shape of the distribution across categories
-- Separately adjusts ice volume and snow volume when they exceed minimum thresholds
+- Separately adjusts ice thickness and snow thickness to match the analysis when they exceed minimum thresholds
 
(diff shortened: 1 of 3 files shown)


shlyaeva:
@Dooruk I was wrong last week: there is code in soca2cice that accounts for ice thickness and snow thickness increments, but it is unfortunately incorrect on develop (and leads to ice thickness increments in restarts even when there are no ice thickness increments in the analysis). This PR fixes it.

guillaumevernieres (approved):
:+1:

Dooruk (approved):

Dooruk:
FYI @yongfeizhang

PR #1190, opened 2025-09-05
Bugfix for Fortran fieldset finalizations

shlyaeva:
## Description

TLDR: this should fix large memory leaks observed with spack-stack/1.9+ and with earlier spack-stack versions on some intel compilers.

Starting stack-stack/1.9 with all compilers (and before that with some intel compilers), fckit is built with `-DENABLE_FINAL=OFF`. Practically this means that one has to call `%final()` methods explicitly in Fortran code to ensure that objects are destructed correctly.
On develop we are missing `afieldset%final()` on increment creation and destruction. This would especially affect applications that create many increments or states. Remember we had unexpectedly huge memory gain on the problematic platform when switching from oops to saber covariances? (https://github.com/NOAA-EMC/GDASApp/issues/1322). oops ensemble covariance uses Increments in multiply, while saber ensemble covariance uses fieldsets. We were just leaking a lot less memory when using saber covariances.

Note: `%final()` should be called for all relevant Fortran fckit/atlas objects, and we may be leaking memory in other places too. I went through the `atlas` objects in Fortran code and checked which ones are not finalized, but it would be good for someone else to double-check if I missed something.

For those interested, investigation happened in https://github.com/JCSDA/spack-stack/issues/1588.

## Testing

Tested on my mac/orbstack/spack-stack1.9 with a simple memory debugging app (https://github.com/JCSDA/spack-stack/issues/1588#issuecomment-3254585996) that creates an increment in a loop, and calls toFieldset on it.

On develop:
```
Iteration 0 before create inc RSS = 78.7656 MB
Iteration 0 after create inc RSS = 79.3906 MB
Iteration 0 after create fs = 79.3906 MB
...
Iteration 99 before create inc RSS = 378.656 MB
Iteration 99 after create inc RSS = 381.656 MB
Iteration 99 after create fs = 381.656 MB
```
On this branch:
```
Iteration 0 before create inc RSS = 79.0273 MB
Iteration 0 after create inc RSS = 79.6523 MB
Iteration 0 after create fs = 79.6523 MB
...
Iteration 99 before create inc RSS = 80.6523 MB
Iteration 99 after create inc RSS = 80.6523 MB
Iteration 99 after create fs = 80.6523 MB
```

### DIFF
--- a/src/soca/AnalyticInit/soca_analytic_mod.F90
+++ b/src/soca/AnalyticInit/soca_analytic_mod.F90
@@ -80,8 +80,8 @@ subroutine soca_analytic_state(state)
       end do
     end do
     call field%set_dirty()
-    call field%final()
   end do
+  call field%final()
 end subroutine
 
 
--- a/src/soca/Fields/soca_fields_mod.F90
+++ b/src/soca/Fields/soca_fields_mod.F90
@@ -338,11 +338,10 @@ end subroutine
 subroutine soca_fields_create(self, geom, vars, aFieldset)
   class(soca_fields),        intent(inout) :: self
   type(soca_geom),  pointer, intent(inout) :: geom !< geometry to associate with the fields
-  type(oops_variables),      intent(in)    :: vars !< list of field names to create
-  type(atlas_fieldset),      intent(inout) :: aFieldset
+  type(oops_variables),      intent(in) :: vars !< list of field names to create
+  type(atlas_fieldset),      intent(in) :: aFieldset
 
   self%afieldset = aFieldset
-  call aFieldset%final()
   self%geom => geom
 
   ! make sure current object has not already been allocated
@@ -433,8 +432,8 @@ subroutine soca_fields_ones(self)
     call field%data(fdata)
     fdata(:,:) = 1.0_kind_real
     call field%set_dirty(.false.)
-    call field%final()
   end do
+  call field%final()
 
 end subroutine soca_fields_ones
 
@@ -523,7 +522,6 @@ subroutine soca_fields_read(self, f_conf, vdate)
         afield1  = self%afieldset%field(self%fields(f)%name)
         call afield1%data(adata1)
         adata1(:,:) = self%fields(f)%metadata%constant_value
-        call afield1%final()
       end if
     end do
 
@@ -628,8 +626,6 @@ subroutine soca_fields_read(self, f_conf, vdate)
         end do
       end do
       call afield1%set_dirty()
-      call afield1%final()
-      call afield2%final()
     end if
 
     ! Compute mixed layer depth TODO: Move somewhere else ...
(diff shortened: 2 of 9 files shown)


guillaumevernieres (approved):
On a roll again! Thanks @shlyaeva

l90lpa:
@shlyaeva I've had a look though at the use of final and just recorded everything (I think you already got those in Utils and Field):
- Utils/soca_utils.F90:
  - soca_stencil_interp: `ageometry` needs final
- Fields/soca_fields_mod.F90:
  - soca_fields_create: remove unused variable `afield`, `afieldset` maybe doesn't need final
  - soca_fields_ones: move `field` final to the end
  - soca_fields_read: `afield1/2/3/4` final calls could be consolidated and moved to the end
  - soca_fields_read_seaice: `afield` final calls could be consolidated and moved to the end
  - soca_fields_write_rst: remove unused variable `afield`
  - soca_fields_tohpoints: move `afield` final to the end
  - soca_fields_update_fields: move `afield` and `ameta` final to the end
- State/soca_state.interface.F90:
  - soca_state_create_c: `afieldset` needs final, also some unused atlas imports could be removed
- Increment/soca_increment.interface.F90:
  - soca_increment_create_c: `afieldset` needs final
- Geometry/soca_geom_mod.F90:
  - soca_geom_end: `fieldset` needs final ???
  - soca_geom_init_fieldset: final calls could be moved to end
  - soca_geom_distance_from_coast: `ageometry` needs final
  - soca_geom_write: `afield` final could be moved to end
- LinearVariableChange/Balance/soca_balance_mod.F90:
  - soca_balance_mult: `fld_m, fld_a` final could be moved to end
  - soca_balance_multad: `fld_m, fld_a` final could be moved to end
  - soca_balance_multinv: `fld_m, fld_a` final could be moved to end
  - soca_balance_multinvad: `fld_m, fld_a` final could be moved to end
- AnalyticInit/soca_analytic_mod.F90:
  - soca_analytic_state: `field` final could be moved to end
- VariableChange/Model2GeoVaLs/Model2GeoVaLs.F90:
  - soca_model2geovals_changevar_f90: `field_out, field2` finals could be consolidated and moved to end
- Increment/soca_increment_mod.F90:
  - soca_increment_dirca: `field` final could be moved to end
  - soca_horiz_scales: `afield` final could be moved to end
  - soca_vert_scales: `field` final could be moved to end

What I noticed with this is that we will want to look at how all of the derived types are destructed and make sure that they finalize any atlas objects they hold, for example soca_geom appears to not finalize the fieldset that it holds when it is destructed. This pattern might also be present in mpas-jedi and fv3-jedi?

shlyaeva:
@l90lpa done here now too, I've gone through the code independently and then verified with your list.

eap:
I'm relaunching the tests to see if it fails again on fv3 linearmodel, vader, or not at all. The interesting and but frustrating thing about the failures in https://github.com/JCSDA-internal/ufo/pull/3794 was that vader was __consistently__ the source of this credential failure. In this PR I can't tell for sure that the issues are related (although it appears that way).

l90lpa (approved):
Looks great, thanks @shlyaeva!

PR #1191, opened 2025-09-09
Fix the value of CO2 to be in mol/mol instead of ppmv

rajichidamb:
## Description

The file had ppmv value for variable mole_fraction_of_carbon_dioxide_in_air, changing that to mol/mol value by dividing the values with 1e6

### Issue(s) addressed

Resolves #(https://github.com/JCSDA-internal/ufo/issues/3734)

## Testing

No yaml or ctest was associated with this file, so no specific test was run. But since the value in the file is wrong it is updated.


fmahebert (approved):

BenjaminRuston (approved):
thanks @rajichidamb for the update though we suspect file is not used

PR #1195, opened 2025-09-18
Cache horizontal localization results when iterating up a column

shlyaeva:
## Description

A way to speed up LETKF with a 3D iterator (using vertical localization in addition to horizontal localization), following up on @ctgh's idea.

build-group=https://github.com/JCSDA-internal/ufo/pull/3804


src/soca/ObsLocalization/ObsLocRossby.h
```diff
@@ -39,15 +39,17 @@ class ObsLocRossby: public ufo::ObsHorLocGC99<MODEL> {
...
 
 // -----------------------------------------------------------------------------
 
 template<typename MODEL>
 ObsLocRossby<MODEL>::ObsLocRossby(const eckit::Configuration & config,
-                                  const ioda::ObsSpace & obsspace):
+                                  ioda::ObsSpace & obsspace):
     ufo::ObsHorLocGC99<MODEL>::ObsHorLocGC99(config, obsspace),
-    options_() {
+    options_(), cacheVector_(obsspace), cachePoint_(0, 0) {
```
ctgh: In the ufo PR I set the initial `cachePoint_` to `(-999, -999)` in order to avoid situations in which the geometry iterator starts at `(0, 0)`.

src/soca/ObsLocalization/ObsLocRossby.h
```diff
@@ -57,6 +59,12 @@ template<typename MODEL>
 void ObsLocRossby<MODEL>::computeLocalization(
     const GeometryIterator_ & i,
     ioda::ObsVector & locvector) const {
+  eckit::geometry::Point3 refPoint = *i;
```
ctgh: `refPoint` and `refPoint2` can be `const`.

ctgh (commented):
Many thanks for adding this! Just two minor suggestions.

src/soca/ObsLocalization/ObsLocRossby.h
shlyaeva: done!

src/soca/ObsLocalization/ObsLocRossby.h
shlyaeva: good idea, I've changed it too!

ctgh (approved):
Thanks, looks great!

Dooruk (approved):

shlyaeva:
Note: I think merge needs to be coordinated with ufo PR.

PR #1196, opened 2025-09-26
Update HTLM test reference

ytremolet:
## Description

See JCSDA-internal/oops#2734

## Dependencies

build-group=JCSDA-internal/oops#2734

PR #1197, opened 2025-10-01
[draft] Test CI

eap:
Testing "--output-junit" flag in ctest.

Check results using `aws s3 ls s3://jedi-ci-build-public-logs/junit_output/`

run-ci-on-draft=true
jedi-ci-test-select=gcc


eap:
Prior run had 2945 tests, 2335 passing, 610 failing.

I've added ioda-data and ufo-data and I'll see if this changes.

eap:
I've merged these changes into https://github.com/JCSDA-internal/soca/pull/1196 and I'm closing this change.

PR #1200, opened 2025-10-16
Fix interface of eckit exception

fmahebert:
## Description

With newer eckit, this current line of code fails to compile; this fix is tested with eckit `1.31.4`.


shlyaeva (approved):
Thank you @fmahebert !

Dooruk (approved):

PR #1206, opened 2025-11-12
Allow to use cice6 history files (ice and snow volumes instead of thicknesses)

shlyaeva:
## Description

**TLDR:** This allows to fix the issue of ice/snow volumes in history files instead of ice/snow thicknesses. To fix, one would need to update their `fields_metadata.yaml`. Without changes to the `fields_metadata.yaml` the behavior stays as before.

**Long story:**

In CICE6 history files (at least for GFSv17) `hi_h` and `hs_h` mean not ice thicknesses, but rather ice volumes (values "per unit grid cell area"). To compute the thicknesses, they need to be divided by ice fraction (`aice_h`).

The `fields_metadata.yaml` that we use at EMC assume that `hi_h` and `hs_h` are ice thicknesses.

This inconsistency can lead to several issues:
- passing incorrect geovals for ice and snow thickness (e.g. for ice thickness or ice freeboard observations) -- not currently used in our setup
- incorrect analysis passed to soca2cice for direct insertion of analysis into cice restarts (the analysis comes from the history files, and so we are inadvertently passing ice and snow volumes as thicknesses and using those to update volumes by-category in the restart files) -- currently used in our setup

To fix this, I decided to add a capability in soca fields read to read in ice/snow volumes and compute thicknesses from them if thickness is requested by the user, but only volume is available in the file.

To use history files like in CICE6 (`hi_h` and `hs_h` volumes, not thicknesses), one would need to update their `fields_metadata.yaml` to associate `hi_h` with `sea_ice_volume` and `hs_h` with `sea_ice_snow_volume`, and give other io names to `sea_ice_thickness` and `sea_ice_snow_thickness` (see an example in a new `fields_metadata_cice6.yaml` in this PR. In this case, if the variables used in the JEDI application are thicknesses, they would be read from the files either using the new io names, or would be computed from `hi_h`, `hs_h` and ice concentration. The output files produced by JEDI application would output ice and snow thicknesses using the new io names.

To continue using history files like in CICE5 (`hi_h` and `hs_h` thicknesses) no change is needed. The currently used `fields_metadata.yaml` likely already associates `hi_h` and `hs_h` with thicknesses.

### Issue(s) addressed

Addresses part of https://github.com/NOAA-EMC/GDASApp/issues/1975

## Testing

1. Tested that without change to `fields_metadata` there's no change in results between this branch and develop:
  - on soca tests: all tests pass
  - on gdas variational and soca2cice applications, both for low and high res: bit-identical results with develop when using the default `fields_metadata.yaml`

2. Tested that if the same data is in cice5-like and cice6-like files, passing correct `fields_metadata.yaml` results in the same results. I added a new `state_cice6` test here, which is a copy of `state` test, except that new ice history file was generated from the existing one to have `hi_h` == ice volume (in the existing file `hi_h` == ice thickness). By passing the correct `fields_metadata_cice6.yaml` to the new test the test passes.

3. Tested that using the update in this branch + `fields_metadata_cice6.yaml` in high-res GDAS tests results in consistent ice thickness and snow thickness updates in soca2cice.
Since the only place where ice and snow thicknesses from analysis are used in soca2cice is in the rescaling, for the first test I configured soca2cice to do rescaling everywhere where possible by setting `ice edge`, `min hice`, `min hsno` to 0.0. The plots below are for the case where there is no ice or snow thickness increment coming from the analysis. The expected result is to only have changes in sea ice thickness where ice was removed.
On develop, or on this branch without changing `fields_metadata.yaml` (plots from left to right: data from the background restart; soca2cice restart; difference between the two):
hr_bg_vs_soca2cice_dev_upd_north
On this branch, using `fields_metadata_cice6.yaml`:
hr_bg_vs_soca2cice_branch_upd_north

Note large-ish blue areas in the ice thickness difference on develop. They correspond to ice concentrations below 100%; with inconsistency between ice thickness and volume, we are effectively multiplying the analysis ice thickness by ice concentration and putting that into the restart files.

Some saving grace (and probably the reason why we didn't spot it earlier) is that in our setup we only do rescaling when analysis ice concentration is above 40% (`ice edge` = 0.4) and where analysis ice thickness (or with incorrect setup, ice volume) is above 50cm (`min_hice=0.5`) and analysis snow depth is above 10cm (`min_hsno=0.1`). With those settings, it's harder (though still possible) to spot the difference between incorrect and correct setup.
On develop, or on this branch without changing `fields_metadata.yaml`:
hr_bg_vs_soca2cice_dev_default_north
On this branch, using `fields_metadata_cice6.yaml`:
hr_bg_vs_soca2cice_branch_default_north


guillaumevernieres:
Thanks for the detailed detective/debugging work @shlyaeva !

test/Data/fields_metadata_cice6.yml
```diff
@@ -0,0 +1,311 @@
...
+# ice state variables with no categories
+# --------------------------------------------------------------------------------------------------
+- name: sea_ice_volume
+  io file: ice
+  io name: hi_h #note, was hicen
+  property: positive_definite
+
+- name: sea_ice_thickness
+  io file: ice
+  io name: hi_div_aice_h
```
shlyaeva: @guillaumevernieres, @Dooruk taking suggestions on better name for this variable

test/Data/fields_metadata_cice6.yml
```diff
@@ -0,0 +1,311 @@
...
+  vert interp: false
+
+- name: sea_surface_height_above_geoid
+  io file: ocn
+  io name: ave_ssh
+
+# --------------------------------------------------------------------------------------------------
+# ice state variables with no categories
+# --------------------------------------------------------------------------------------------------
+- name: sea_ice_volume
```
guillaumevernieres: These names are trying to be COARDS/CF compliant, so in this case, `hi_h` is in fact `sea_ice_thickness` defined as the average thickness in the grid cell.

test/Data/fields_metadata_cice6.yml
guillaumevernieres: It's not sea_ice_thickness as defined by COARDS/CF. `sea_ice_thickness` by default, is the cell average thickness, so hi_h. COAARDS/CF has a `cell_method` attribute (default is mean over grid cell) that could be redefined, but that wouldn't work in this context without modifying the code by more than we want ... 
How about `sea_ice_thickness_where_sea_ice_present`?

test/Data/fields_metadata_cice6.yml
```diff
@@ -0,0 +1,311 @@
...
+- name: sea_ice_thickness
+  io file: ice
+  io name: hi_div_aice_h
+  property: positive_definite
+
+- name: sea_ice_area_fraction
+  io file: ice
+  io name: aice_h #note, was aicen
+
+- name: sea_ice_snow_volume
```
guillaumevernieres: same comments as above

guillaumevernieres (commented):
The original name was correct and following the COARDS/CF definition, it's just not intuitive how `sea_ice_thickness` should be defined.

test/Data/fields_metadata_cice6.yml
shlyaeva: OK, great point.

test/Data/fields_metadata_cice6.yml
shlyaeva: OK, so we'll need to change our yamls to use `sea_ice_thickness_where_sea_ice_present` (no big deal, we can do that).

shlyaeva:
@guillaumevernieres, @Dooruk: given that this affects both noaa and nasa, maybe we can do this more systematically.

1. Soca tests. How about I change all the test files in soca to have `hi_h`, `hi_s` use the definition as we now understand it, and change the default fields_metadata accordingly?

2. Names. I'm not sure now that I agree with you @guillaumevernieres. I did some googling, and here's what I see.
https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html: has `sea_ice_thickness` (definition: _"Thickness" means the vertical extent of a layer. "Sea ice" means all ice floating in the sea which has formed from freezing sea water, rather than by other processes such as calving of land ice to form icebergs._) and also `sea_ice_volume`. 
https://cfconventions.org/Data/cf-conventions/cf-conventions-1.4/build/cf-conventions.html defines `sea_ice_thickness` as _sea_ice_thickness:cell_methods="area: mean where sea_ice over sea"_
https://pcmdi.llnl.gov/mips/cmip3/variableList.html defines `sea_ice_thickness` as _this thickness, when multiplied by the average area of the grid cell covered by sea ice, yields the time-mean sea ice volume_.
Are you both OK with sticking to `sea_ice_thickness` defined as a quantity over ice area, and `sea_ice_volume` defined as a quantity over grid cell? I'm OK with changing if we find that those names are not what's commonly used in the field.

3. Science. Should we be updating ice thickness or ice volume? I think it should be ice thickness (as in hi_h/aice_h), but raising the question nonetheless.

guillaumevernieres:
Thanks for digging further into this. All I could find was coming from what looked like an old forum discussion with the exact same confusion we are talking about here. 

Your proposed solution in this PR sounds good to me @shlyaeva , perhaps with proper comments in the metadata yaml defining these variables precisely? If not we'll probably have the same discussion in 6 months!

shlyaeva:
I made the following updates:
- changed all relevant cice history files in the tests to have hi_h and hs_h be volumes by replacing the existing values with hi_h * aice_h and hs_h * aice_h. Files that were changed:
  - `test/Data/36x17x25/restarts/cice.res.nc`, 
  - `test/Data/72x35x25/history/cice_history.nc`
  - `test/Data/72x35x25/restarts/cice.res.nc`
  - `test/Data/72x35x25/restarts_ens/cice.res.ens.*.nc`
- changed `fields_metadata.yml` to associate ice and snow volumes with `hi_h` and `hs_h`, and new io names with ice and snow thicknesses
- updated references for the ensemble-related tests. This had to be done because the ensembles in `test/Data/72x35x25/restarts_ens/cice.res.ens.*.nc` had non-zero values for hi_h and hs_h where aice_h was zero, and so the min/max/avg values in the prints changed after changing to use volumes (those non-zero values became zero when switching to volumes). To make sure that there's no error introduced in this PR, I generated new references from develop, by updating the ensemble files to remove hi_h and hs_h where aice_h is zero. The tests on this branch pass with those references.
- updated yaml and reference for `3dvar_moreice` test. This test is using a combined restart-history file `gdas.ice.t12z.inst.f006.nc`. In that file `hi_h` and `hs_h` already meant ice and snow volumes, so I changed the yaml to use volumes instead of thicknesses and changed the strings in the reference accordingly.

This is ready for re-review. The merge will not affect existing experiments in any way until changes are made to `fields_metadata.yaml` used in the experiments.

shlyaeva:
@guillaumevernieres, @Dooruk would you have time to review this soon?

guillaumevernieres (approved):
Looks right. Thanks for the fix @shlyaeva.

Dooruk (approved):
Looks good to me. Thank you @shlyaeva for making the change as easy as changing the `fields_metadata.yaml`!

shlyaeva:
I have run some more tests with forecasts, and with the default soca2cice settings (`min hice: 0.5`, `min hsno:0.1`) CICE throws its hands up in the air and says it can't handle this (those settings were used for ice/snow volume before; now that it's thickness, apparently that's not enough). In the meantime, Dmitry Dukhovskoy from EMC has been working on direct insertion of ice/snow thickness, and has an approach that may work better than what we do now. I'll look into adopting it for soca2cice.
I am proposing not to update `fields_metadata.yaml` in the meantime.

Dooruk:
Doesn't clarify our sea-ice case but here are the ROMS fields:

https://github.com/myroms/roms/blob/develop/ROMS/External/varinfo.yaml

PR #1057, opened 2024-07-11
Feature/add structured grid and cmakelist edit

jrocha22:
## Description

PR to add a ConvertToStructuredGrid executable to alter files to be in structured lat/lon grid for easier post processing.
CMakeLists.txt file edited to run executable

### Issue(s) addressed

No associated issue number but resolves grid difficulties that might arise from further processing of files

## Testing

New executable tested with twelve different high resolution (1440x1080x75) files and two low resolution (72x35x25) files from MOM6. Variables pulled were sea surface temperature and sea water potential temperature. All files ran through with no issues.  
Executable tested with Hercules.
Are the changes covered by ctests? No (If not, tests should be added first.)

PR #1212, opened 2025-12-12
Interpolate State to Gaussian Grid

guillaumevernieres:
Duplicate of https://github.com/JCSDA-internal/soca/pull/1057

jedi-ci-test-select=intel


guillaumevernieres:
The `ci` failures aren't related to this PR, the added test (_converttogauss_) is fine:
converttogauss

test/testinput/converttogauss.yml
```diff
@@ -0,0 +1,40 @@
...
+    interpolation method:
+      type: finite-element
+    number of latitude gridpoints: 12
+    variables to output:
+    - sea_surface_height_above_geoid
+    - sea_water_potential_temperature
+    - sea_water_salinity
+    - sea_ice_area_fraction
+    all model levels: true
+    bottom model level: false
```
Dooruk, line 34: Just curious why is bottom level false but all levels true?

Dooruk:
Looks good to me, I will wait for tests to pass before I approve it.

Does this handle land mask interpolation well?

guillaumevernieres:
> Looks good to me, I will wait for tests to pass before I approve it.
> 
> Does this handle land mask interpolation well?

no

test/testinput/converttogauss.yml
guillaumevernieres: @Dooruk , It's a copy paste from the `fv3jedi` ctest yaml. I'm not sure we need the `bottom model level` key here, you can't have it to true when `all model levels` is true.
But, I don't mind having the extra key. I usually use the ctest yamls to template our application yamls, and not having to dive into the code to figure out what the key is for the  `bottom model level` option sounds good to me!

Dooruk: Ah ok, I didn't realize its either/or.

shlyaeva (approved):
I'll look into the test failures, I agree they are not related, so must be happening on develop as well.

shlyaeva:
> Does this handle land mask interpolation well?

> no

@guillaumevernieres what do you mean by that?

guillaumevernieres:
> > Does this handle land mask interpolation well?
> 
> > no
> 
> @guillaumevernieres what do you mean by that?

@shlyaeva , I guess it depends on our definition of "well". I'd say it works for most cases.
But it doesn't work for all cases: This is related to that flooding business I was trying to push to oops a while back. What we have here will work well enough for most cases but will be off when the coastline geometry is "complex" (island, peninsula, ...) and the interpolation ends up using the wrong neighbors.

Dooruk:
> @shlyaeva , I guess it depends on our definition of "well". I'd say it works for most cases. But it doesn't work for all cases: This is related to that flooding business I was trying to push to oops a while back. What we have here will work well enough for most cases but will be off when the coastline geometry is "complex" (island, peninsula, ...) and the interpolation ends up using the wrong neighbors.

To me that sounds unwell 😃 it's tough nut to crack though..

Dooruk (approved):
Very well 👍🏼

PR #1217, opened 2025-12-18
Add TLMToolbox executable and test for SOCA

### LINKED ISSUE #1216: Add TLMTool box executable and ctest for SOCA
This issue is to add the TLMTool box executable to soca and add a ctest which will eventually replace the current linearization error ctest when the linearization error application is removed from oops.

xian22:
## Description

This PR adds the an executable for `TLMToolbox` as well as a ctest to replace the current `LinearizationError` test as the executable for that will be removed soon.  The new test passes with the reference of the old test which is expected.  

### Issue(s) addressed
- fixes #1216 

## Testing

The output using the new app should pass with the existing `LinearizationError` reference file and it does. 

Tested with GNU spack stack 1.9

## Dependencies
none


shlyaeva:
Thank you so much for doing this @xian22 ! Similar to the comments in the fv3-jedi PR, can we perhaps replace the current linearizationerror test with this one, and remove the linearizationerror application? I am happy to work on that too if you're busy with other things, let me know!

xian22:
Sort of forgot about this over the holiday, I have removed the old linearization error test and it is replaced with the tool box equivalent.

shlyaeva (approved):
Great, thank you so much @xian22 !

Dooruk (approved):

tom-j-h:
Thanks all!

PR #1224, opened 2026-03-09
Add sequential kalman filter framework, and EAKF

travissluka:
## Description

This adds a test for the sequential EnKF (EAKF in this case)

Also added the `computeLocalization(Point3, Point3)` interface to the obs localization so that things would compile and warn us if we tried to use it. We can't use the rossby based localization for sequential EnKF at the moment, though. The rossby based localization needs to call `GeometryIterator::getFieldValue()` so it will not work with obs/obs localization. The fix to this requires more thought, and Travis has hit his thought quota and can't afford any more credits.

## Testing

An EAKF test is added, with minimal scientific or technical evaluation (If you run LETKF with GC99 localization and compare it to EAKF, the increments look similar if you squint. And no, I did not think to save those plots)

## Dependencies

- https://github.com/JCSDA-internal/oops/pull/3193
- https://github.com/JCSDA-internal/ioda/pull/1674
- https://github.com/JCSDA-internal/ufo/pull/4027


shlyaeva (approved):
Nice, thank you!

ncrossette (approved):

kbhargava (approved):
🎉

frolovsa:
wohoo!

fmahebert:
@shlyaeva I believe you can merge this one now

PR #1226, opened 2026-03-17
Remove specific github templates

ytremolet:
## Description

Remove specific github templates so organization defaults are used.


gibbsp (approved):
I'm not sure how many times templates are actually used but I think I've seen different ones in different repo's. Perhaps we should consolidate all under .github, for example, does "feature_request" exist anywhere else? But maybe I'm stating the obvious here, and that is the plan.

eap (approved):

shlyaeva (approved):

PR #1233, opened 2026-03-25
Remove the commented out tests

### LINKED ISSUE #1228: Commented tests
**test/CMakeLists.txt**

```
## vertical localization is not working right now. h was removed from the
## geometry (it is part of the state), so we need to rethink how the
## geometryiterator works in the vertical
# soca_add_test( NAME obslocalization_vertical
#                SRC  TestObsLocalization.cc )

# # Test mom6_bgc forecast (optional)
# if ( ENABLE_OCEAN_BGC )
# soca_add_test( NAME forecast_mom6_bgc
#                EXE  soca_forecast.x
#                NOTRAPFPE
#                TEST_DEPENDS test_soca_gridgen )
# endif()

# NOTE, enshofx is disabled because MOM6 always writes out some diagnostic
# netcdf files (which can't be disabled). This is normally not a problem, except
# with enshofx, because multiple instances of MOM6 are started under separate
# MPI_COMMs, causing problems when they all try to write to the same file at the
# same time. We'll need to take a closer look at MOM6 output when we start
# dealing with in-core MOM6

# soca_add_test( NAME enshofx
#                EXE  soca_enshofx.x
#                MPI 3
#                NOTRAPFPE
#                TEST_DEPENDS test_soca_gridgen )

# # Following test does not work yet. IODA crashes
# soca_add_test( NAME 3dvar_zero_ob
#                EXE  soca_var.x
#                NOTRAPFPE
#                TEST_DEPENDS test_soca_static_socaerror_init )

# # Following test is breaking on intel 2022, beyond 1 iteration solver produces NaNs
# soca_add_test( NAME 3dvar_single_ob
#                EXE  soca_var.x
#                NOTRAPFPE
#                TEST_DEPENDS test_soca_gridgen )
```

shlyaeva:
## Description

Removes some commented out tests as there are either no plans to implement features for those tests, or no need for the tests. If/when there is need, the tests can be added back.
Tests removed:
- obslocalization_vertical: we don't have an option to use model depth as vertical coordinate yet, and no plans to implement it.
- enshofx - no need to run this test as I don't think it introduces anything important to test. (we also now have a hack in `AnalysisPostProc` that allows to run executables like this: https://github.com/JCSDA-internal/soca/blob/ae5aabcfea7dbcc9da98e9169621ef8327511280/src/mains/AnalysisPostproc.h#L76-L80
- zero_obs - no need to test, it's tested elsewhere (e.g. oops and fv3-jedi)
- single_obs - I decided no need in this test.

## Issue(s) addressed

Resolves #1228 
## Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have run the unit tests before creating the PR


ctgh (approved):

guillaumevernieres (approved):
:+1:

PR #1243, opened 2026-05-14
Parallel ensemble I/O for State and Increment

### LINKED ISSUE #1125: Completely redo model I/O to parallelize it
## Description
Model I/O is slow, and is especially noticeable when dealing with ensemble output from LETKF. This is due to the fact that all state I/O is currently being done in serial, on PE 0. ... not the most efficient way.

## Solution
I/O needs to be parallelized. There are 2 ways this could be done
1. Use the tiled I/O capabilities of FMS. (This is not preferable because files in the GDAS workflow are not tiled, and I doubt @guillaumevernieres wants to put in `mppnccombine` executable calls in the workflow for each output file)
2. Otherwise, FMS does not have parallel I/O capabilities, we'll need to use direct netcdf and mpi scatter/gather calls.

Assuming we use our own netcdf calls for soca I/O there are several things of varying complexity/craziness we could try
1. parallel netcdf where each PE or a pool of PEs does the I/O. This should perform well, if tuned correctly. Also, I'm not a fan of parallel netcdf I/O because it works best only if the file chunking and underlying filesystem is setup correctly, which they rarely are
2. serial netcdf, but we do some crazy asynchronous stuff in the background. Instead of all IO taking place on PE 0 (as is currently done), the file IO is done round-robin on different PEs. This by itself wouldn't give any speedup, but the state read or write could be done asynchronously, with locks placed around all other state functions that wait for pending asynchronous I/O to finish. The appealing thing about this is that no change would be needed to oops to allow for ensemble I/O done with one file per PE. Also, I have experience implementing this with past versions of LETKF I've worked on. 
3. Add a "parallel ensemble state read/write" set of functions to oops so that a model interface could handle its own mpi scatter/gather then do one-file-per-PE I/O in parallel. I don't feel like touching oops, so no.

I'll probably go with number 2, but we'll see.

travissluka:
## Description

> [!WARNING]
> **Make sure you check out the matching oops branch (JCSDA-internal/oops#3316)!** This soca branch will still compile and run, but it won't go any faster without the oops dispatch hooks.

Replaces the per-member-serial ensemble read/write path with bulk routines that drive concurrent I/O across rotated reader/writer PEs, with optional MPI-async batching for the gather/scatter collectives. Previously the LETKF was doing all ensemble file writes on PE0... ouch. Ensemble member reads were using "strided" read mode (whereby each PE reads its section of a netcdf file), not the worst, but probably not the best.

This PR uses a `oops::StateSet::readEnsemble` / `writeEnsemble` set of methods that let us bypass the per-member loop, doing ensemble read/write in true parallel.

### YAML configuration

The default settings now should do parallel ensemble I/O in what is hopefully the most efficient way on an HPC. But there are some knobs to turn if you need. All knobs live under `geometry.io`.

| key | values | default | what it does |
|---|---|---|---|
| `ensemble read` | `scatter` \| `strided` | `scatter` | Scatter = M reader_pes pull whole-globals and `MPI_Scatterv` to compute groups; strided = each PE reads its compute-domain tile directly. |
| `async mpi` | `true` \| `false` | `true` | Batch `Igatherv` / `Iscatterv` across all (reader/writer, var) pairs so collectives with different roots overlap. Applies to both read and write. |
| `ensemble batch size` | `N >= 0` | `0` | `0` = process all members in one batch (maximum I/O overlap, peak memory ~ all members resident at once). `N > 0` commits the ensemble in waves of at most `N` members to cap peak memory (set `N` ~ number of PEs); batch `1` matches a serial write's memory footprint. |
| `single state read` | `strided` \| `scatter` | `strided` | Per-PE direct compute-tile reads vs single-reader-then-scatter for the single-state (non-ensemble) path. Strided is the right default on parallel filesystems. |

### Memory

Bulk parallel I/O necessarily keeps more in flight than the old serial path: each rotated writer PE stages a member's whole global field (and on read, each reader PE does the same). Two things keep that bounded:

- Collective buffers are freed per-completion (`MPI_Waitany`) as each gather/scatter finishes, rather than holding every member's buffer until one big `Waitall`. On a 1deg 7-member LETKF this cut peak memory ~11% at no batching, for free (I/O overlap unchanged).
- `ensemble batch size` caps how many members are resident per pass. `ensemble batch size: 1` brings the footprint down to the serial path's level; the default single batch trades that memory for maximum overlap. Output is bit-identical across all batch sizes.

### Expected performance impact

Based on tests with a synthetic memory/file-system delay emulating HPC performance... my assistant forecasts an improvement of  1.5–3× on reads, 10–30× on writes... let's see how good that forecast is!

- Parallel ensemble write (now the only write path) is unambiguously faster than the old gather-to-PE0 serial write, no question there
- I'm fairly confident that `ensemble read: scatter` is faster when the number of PEs/nodes is large. But if read times are slower, it might be worth testing putting this back to `strided`, which is the currently implemented method
- `async mpi: true` in theory should go faster when doing ensemble I/O, but the flag is here in case we want to verify
- for `single state read` i'm honestly not sure which would be faster, in theory `strided` should be faster if number of PEs >> the number of lustre OSTs.
- `ensemble batch size: 0` should be the fastest ensemble writes, but also uses the most memory. The knob is there if we need to reduce memory.

## Issue(s) addressed

- Resolves JCSDA-internal/soca#1125

## Dependencies

This PR depends on:
- https://github.com/JCSDA-internal/oops/pull/3316 (adds the `State`/`Increment` `readEnsemble`/`writeEnsemble` dispatch hooks this PR implements)

Merge order: oops#3316 first, then this PR.

build-group=https://github.com/JCSDA-internal/oops/pull/3316

## Impact

Behavior is identical to the legacy path. Expected wall-clock speedup on HPC; no expected impact on downstream science code.

## Checklist

- [ ] I have performed a self-review of my own code

> [!NOTE]
>  Validated bit-identical to the legacy path on the soca EnKF/LETKF ctests. Still need to stress test on an HPC. 
> Did I review my code? Of course not, I figure it's correct judging by the cpu/memory usage, runtime, and the fact the output files are bit identical. Who cares about looking at the code??


shlyaeva:
I've just tested it with our 0.25 deg run (but with 30 members), nice!:
this branch:
```
  0: OOPS_STATS Run end                                  - Runtime:    292.49 sec,  Memory: total:  1563.10 GB, per task: min =     4.44 GB, max =    14.54 GB
```
previous run:
```
  0: OOPS_STATS Run end                                  - Runtime:    537.97 sec,  Memory: total:   908.94 GB, per task: min =     3.09 GB, max =     6.67 GB
```

The memory seems quite a bit larger, I'll look into it later unless @travissluka you have an idea why.

travissluka:
> The memory seems quite a bit larger, I'll look into it later unless @travissluka you have an idea why.

Yeah, some memory increase is inevitable. Each ensemble member requires the extra buffer for the full state on read/write. It's a trade-off. You can set `ensemble batch size` to limit the number of members that are read/written in parallel at a time to cut back on peak memory usage.

That said, I know there are still areas for memory optimization (the async MPI calls have duplicate read/write buffers that I would like to get rid of). But, good to see it's on the right track, thanks for testing !

shlyaeva (approved):
Review solely based on testing and great timing results. I'm not an expert in new or old i/o, and can't review the code changes well. 🎉

PR #1244, opened 2026-05-18
Ensemble postproc: Allow to not have zero-out variables in the list of increment variables

shlyaeva:
## Description

Allows to not specify zero-out variables in the list of increment variables in the ensemble postprocessing app (they don't need to come from backgrounds or analyses since they are zeroed out anyway!)

This would allow us to run recentering app with LETKF, while recentering analyses that don't have currents around the 3DVar analysis that doesn't have currents. (mom6 IAU requires currents in the increments files).

## Manual Testing Instructions (optional)

Specify the currents variables only in `set increment variables to zero` yaml section in the ensemble postprocessing app in gdas.

## Checklist


guillaumevernieres (approved):
:+1:

shlyaeva:
I've pushed a minor update to `State::operator+=` that allows the increment that's added to the state to have more variables than in the state (those are ignored in this case). This is to handle adding the increment with extra (zeroed-out variables) to the background state.

Dooruk (approved):

PR #1246, opened 2026-05-27
Refactored soca -> cice

shlyaeva:
## Description

Sea-ice analysis postprocessing in C++/atlas (`PostProcessIce`): replaces the Fortran `Soca2Cice` variable change with a C++/atlas postprocessor that projects an aggregate SOCA ice analysis onto a per-category CICE restart. The new algorithm is largely based on @DmitryDukhovskoy's Python scripts to insert ice concentration, thickness and snow depth (https://github.com/DmitryDukhovskoy/RTOFS_utilities/blob/gaea_pub/prepare_cice6/insert_iconc_ithkn_cice6_restart.py and https://github.com/DmitryDukhovskoy/RTOFS_utilities/blob/gaea_pub/prepare_cice6/insert_hsnow_cice6_restart.py).  Validated on 7 production GDAS cycles spanning different seasons.

### What's new

- **`soca::PostProcessIce`** in src/soca/PostProcess/,
  with a single public entry point:

  ```cpp
  soca::State PostProcessIce::postprocess(const soca::State & analysis) const;
  ```

  Reads the CICE background restart, applies the per-cell pass, writes the postprocessed restart in update mode, returns an aggregate-ice State `{aice, hice, hsno}` matching what was written. Used by the  standalone `soca_postproc.x` app, by `AnalysisPostproc.h`'s ensemble loop, and by gdasapp's increment handler.

- **Per-cell pass** (see README for details): case dispatch (LAND / ICE2NOICE / NOICE2ICE / ICE2ICE), ITD rebin (on by default), aicen-weighted snow distribution, optional freeboard enforcement, mass-conserving min-vice cleanup. Pure column-physics helpers (BL99 enthalpy, BZ99 salinity profile, thickness-category solver, freeboard) live in `IcePhysics.h/.cc` with their own unit tests in `TestIcePhysics.cc`.

- **New-ice thermo seeding**: cells that transitioned from `bg_aicen=0` to `new_aicen>0` get Tsfcn from the area-weighted mean of a KDTree donor with any ice (global lat/lon tree, donor data gathered once via a sparse halo exchange); sub-surface qice/sice synthesized from CICE physics at the ocean freezing point.

- **`soca_postproc.x`** standalone application
  (src/mains/Postproc.h): takes `background` + `increment`, forms `analysis = bg + incr`, runs `PostProcessIce`. Also accepts an explicit `analysis` block.

- **Dedicated CICE restart writer** on the soca::State side:
  `soca::State::writeCice(cfg)` calls a new Fortran entry
  `soca_fields::write_cice` that wraps the update-mode CICE writer (byte-copy input restart → output, overwrite only modelled variables, ~40 unmodelled CICE vars pass through). 

- **`fields_metadata.yml`**: new `<LEVEL>` placeholder for templated per-layer entries (CICE per-layer restart vars `qice00N`, `sice00N`, `qsno00N`); combines with `<CATEGORY>` to expand `ncat × ice_lev` entries automatically. Per-cat ice `io name` updated to restart naming (`aice<N>`, `vice<N>`, `vsno<N>` - dropping the `_h` suffix which is the CICE history convention; restarts don't carry it).

- **`AnalysisPostproc.h`**: ensemble loop migrated to call `PostProcessIce::postprocess` per member. `Soca2Cice` variable change no longer used here.

### Tests

Four end-to-end tests on the 72×35 grid (one per major code path) plusa  pure unit test on the column physics:

- `test_soca_soca2cice_new` - baseline (rebin on)
- `test_soca_soca2cice_new_freeboard` - freeboard enforcement
- `test_soca_soca2cice_new_seed` - new-ice seeding + thermo
- `test_soca_soca2cice_new_bgfallback` - analysis variables resolver
  (only aice analysed, hice/hsno fall back to background)
- `test_soca_icephysics` - pure C++ unit test of the column-physics helpers
- `test_soca_postprocice_vs_soca2cice` - Fortran-vs-C++ regression
  (`Soca2Cice` output compared bit-wise to `PostProcessIce` output;
  budget = 2 differing cells at 1e-10 tolerance)
- `test_soca_ensanpproc` - ensemble postproc end-to-end

All pass. The legacy Fortran `Soca2Cice` path is still available through `soca_convertstate.x` for the regression test; it can be deprecated in a follow-up PR.

### Configuration

A minimal yaml stanza:

```yaml
postprocess ice:
  ncat: 5
  ice_lev: 7
  sno_lev: 1
  cice restart:
    input:  /path/to/cice_background.res.nc   # also the update-mode template
    output: /path/to/cice_postprocessed.res.nc
```

Full schema with defaults in src/soca/PostProcess/README.md.

## Checklist


shlyaeva:
I'm going to run a few cycles with the ensembles using this postprocessing and report back here. In the meantime, this is mostly ready for review, I don't anticipate major changes. If it looks painful to review, just know that I've been working on this on and off for half a year, and I'm with you 😆 .

shlyaeva:
I realized that I made a mistake in my previous legacy vs this branch vs Python scripts comparison computations (I forgot again that hi_h and hs_h in the history files are effectively volumes, not thicknesses). I have now cleanly reran the comparisons, so I'm removing my previous comment and reposting the updated results here. Note that the results are now consistently good for Dmitry's scripts, without weird artifacts around ice edge in my previous incorrect comparisons. The conclusion is the same: no degradation compared to Dmitry's scripts, significantly better than legacy Fortran soca2cice for ice thickness and snow depth.

## PostProcessIce: production-grid validation

Side-by-side comparison of the new C++/atlas `PostProcessIce` path against the legacy Fortran `Soca2Cice` path and @DmitryDukhovskoy's Python `prepare_cice6` scripts (dd here), on retro + realtime GFSv17 data (0.25deg grid). 

The comparison was done on `20240715`, `20250115`, `20250330`, `20250715`, `20250920`, `20251221`, `20260322`, the results here are just for `20240715`

For each cycle we hold the same DA analysis (from GFSv17 with artificially increased ice thickness and snow depth to test insertion of those variables) and the same CICE background fixed, and compare the resulting CICE restart written by each of the three pipelines.

### Aggregate cell-level RMSE vs analysis

Mask: cells where `analysis.aice > 1e-4`. 

field | new | legacy | dd | background
-- | -- | -- | -- | --
aice | 0.0026 | 0.0042 | 0.0027 | 0.043
hice (m) | 0.0030 | 0.192 | 0.034 | 0.252
hsno (m) | 1.7e-5 | 0.026 | 0.0045 | 0.042

### Per-category ICE2ICE perturbation vs background

Mask: per-cat slots where both `bg.aicen > 1e-4` and `new.aicen > 1e-4`. This measures how much the postprocessing perturbs the background per-cat shape on cells that have ice in both bg and analysis.

| field | new RMSE / max\|d\| | legacy RMSE / max\|d\| | dd RMSE / max\|d\| |
|---|---|---|---|
| aicen | 0.01726 / 0.514 | 0.1714 / 1.000 | 0.01749 / 0.514 |
| vicen | 0.0824 / 0.897 | 0.1464 / 4.094 | 0.08625 / 1.199 |
| Tsfcn | 0.6234 / 1.000 | 7.0742 / 47.409 | 1.0502 / 42.610 |
| sice001 | 0.0000 / 0.000 | 2.0393 / 29.187 | 0.5855 / 23.459 |
| apnd | 0.6811 / 1.000 | 0.0000 / 0.000 | 0.6584 / 1.000 |
| ipnd | 0.0000 / 0.000 | 0.0000 / 0.000 | 0.0000 / 0.000 |

### Plots

Three production-grid comparisons for `20240715` (arctic summer / antarctic winter). Each figure has 3 rows (`new` / `legacy` / `dd`) and 4-5 columns (`background`, `analysis`, `updated`, one or two `updated − analysis` diff
panels at different colour ranges):

- Ice thickness in the Arctic:
hice_arctic


- Snow thickness in the Arctic:
hsno_arctic


- Ice thickness in the Antarctic:
hice_antarctic

All 7 dates exhibit the same qualitative behaviour - the new path is consistently close to the analysis on ice thickness and snow (significantly better than legacy).

shlyaeva:
I pushed a few updates to this branch yesterday that fix some issues with ice thicknesses out of category bounds (thank you @DmitryDukhovskoy for you help in identifying the issues!). I have now successfully run a few cycles of a cycling experiment with an ensemble where the new soca2cice is used for updating CICE restart for the deterministic forecast after 3DVar and for the ensemble forecasts after ensemble recentering.

NickSzapiro-NOAA:
There is a lot here! Some thoughts on reading through, please take as you wish.

Dynamics are missing (velocities and stresses). Weighting donors and ice-->ocn as aice vanishes are some options.
Maybe the thinking is that dynamics spins up quickly (in timesteps?), but there are still those initial timesteps to worry about.

There can be melt ponds and snow on ice in the same cell.

One end-to-end test may be that DA shouldn't change much if the background is close to the analysis. I don't know if you have tests like this, but it's good sanity check. It's one reason I mention an issue of zeroing ponds if snow changes. Some variables can have subtle meanings (https://github.com/CICE-Consortium/CICE/issues/1096#issuecomment-4128163536)

Maybe the self cell should also be included in the neighborhood of donor cells. It wouldn't change anything in the noice2ice branches

shlyaeva:
@NickSzapiro-NOAA thank you for looking through this and for your comments! I'll think about that and get back to you in probably a couple of weeks (code sprinting next week).

DmitryDukhovskoy (commented):
I looked through the changes in the SOCA code. The insertion logic closely follows the Python scripts, except for the redistribution of sea-ice state variables across the ice thickness categories. This difference should have a small impact, provided it does not produce category-mean thicknesses (vicen/aicen) that fall outside the prescribed thickness bounds. I don't see any obvious issues.

mjkagnes123:
Thanks for putting together such a thorough update. The overall workflow was easy to follow, and the validation tables and figures were especially helpful in understanding how the new approach compares with the legacy method and Dmitry's scripts.    

I had two small comments on PostProcessIce.cc.

First, I may be missing something, but a_aice appears to be created as a writable Atlas view of the input "analysis" field and is then clamped in place. Since "analysis" is passed as a const State& , would it be safer to copy aice into a local vector, similar to a_hice_vec and a_hsno_vec, before applying the bounds?  

Also, in the ICE2ICE path,  it looks like "vicen" can be redistributed while the existing "sice" profile remains unchanged. Is the idea that any added ice volume takes on the salinity of that category? I was also wondering whether total salt is conserved through this redistribution.

DmitryDukhovskoy:
> Also, in the ICE2ICE path, it looks like "vicen" can be redistributed while the existing "sice" profile remains unchanged. Is the idea that any added ice volume takes on the salinity of that category? I was also wondering whether total salt is conserved through this redistribution.

Hi @mjkagnes123, the sea ice salinity profile and enthalpy are computed only when the ice column has essentially no existing salt (sice ~0) or enthalpy is not within the correct bounds. This is always the case for newly formed ice (noice --> ice), but it is usually not true for the ice2ice pathway. When sice ~ 0, the model has to estimate the salinity profile, since there is no existing salt distribution to work from. The estimate is based on the BL99 (Bitz & Lipscomb, 1999) approach, as implemented in CICE4. The direct insertion approach does not guarantee conservation of salt, mass, or energy.

guillaumevernieres (approved):
I did not review the code but I'm testing this within the gfs/ufs. It does as advertized!
Thanks @shlyaeva

mjkagnes123 (approved):
I really appreciate the work you did on this. The tables and figures were very useful in showing how this is better than the way of doing things. The improvements over the legacy approach are clear. Dmitrys explanation also helped me understand the part about treating salinity. My comment about  'a_aice' is still something I think could be improved. It is not a big deal and it does not stop this from being approved. I am approving this after I looked at 'PostProcessIce.cc'. The improvements over the legacy approach, like the salinity treatment are good.

Dooruk (approved):
@shlyaeva this is quite a comprehensive overhaul of the current approach and I really appreciate the README. It will take us a while to implement this into our workflow. Since the old approach is still available for a while, I don't want to hold this back.

Also, thank you @mjkagnes123 for your thorough review!

shlyaeva:
> a_aice appears to be created as a writable Atlas view of the input "analysis" field and is then clamped in place. Since "analysis" is passed as a const State& , would it be safer to copy aice into a local vector, similar to a_hice_vec and a_hsno_vec, before applying the bounds?

@mjkagnes123 thank you for catching this, I've addressed it in the latest commit.

shlyaeva:
Thank you everyone for the reviews! I'm going to merge in its current form once the tests pass, and we can improve on it in follow-up PRs as needed.

guillaumevernieres:
It's very possible that I face planted again when modifying the configuration (volume vs thickness) ... But I don't think so. I'm finding a linear growth of seaice volume for both poles when using the new soca2cice. @Dooruk , @mjkagnes123 , have you guys had time to test?

guillaumevernieres:
> It's very possible that I face planted again when modifying the configuration (volume vs thickness) ... But I don't think so. I'm finding a linear growth of seaice volume for both poles when using the new soca2cice. @Dooruk , @mjkagnes123 , have you guys had time to test?

_Edit: It's not as simple as what I wrote above, but I'd be curious to compare notes  once you test this with your system._

Dooruk:
No, we are looking into this now. We didn't address the CICE6 history issue @shlyaeva identified last year so I have to make that change first...

https://github.com/JCSDA-internal/soca/pull/1206

Sorry, this must be buried under some discussion somewhere but what is the purpose of having these following fields in `fields_metadata.yaml`?
`hs_div_aice_h` `hi_div_aice_h`, are these calculated internally somehow?:

https://github.com/JCSDA/soca/blob/3b4b23a0742711469989bef97cd67b77a0cabcdf/test/Data/fields_metadata.yml#L74-L91

shlyaeva:
> Sorry, this must be buried under some discussion somewhere but what is the purpose of having these following fields in fields_metadata.yaml?
hs_div_aice_h hi_div_aice_h, are these calculated internally somehow?

This is for the case where background (model history) files have ice volume, but soca uses ice thickness as a variable. The code will automatically compute ice thickness from ice volume for the backgrounds, and then all the calculation and output in soca would be in ice thickness. So with this updated fields_metadata, we'd read `hi_h` (volume) from the history file, compute ice thickness from it, and output soca analysis and/or increment in ice thickness as `hi_div_aice_h` to distinguish it from the ice volume. soca applications that use the mix of model history files with `hi_h` and soca analysis files with `hi_div_aice_h` would correctly compute ice thickness for the model files and use ice thickness as is for the soca files.

Dooruk:
Thanks @shlyaeva, that is helpful. I searched the whole soca repository for `hs_div_aice_h` but it only points to `fields_metadata.yml`.

Dooruk:
I'm very close. I actually have it running for our workflow but there is one caveat with regards the disparity between the background time vs analysis/increment time. The only difference between the below configurations is how the increment time is set.

Since for `cost type: 3D-FGAT` background is beginning of the window and analysis&increment is the middle I put that in the non-working configuration and i received this error:

`Exception: Assertion failed: validTime() == dx.validTime() in operator+=,`

I mean in terms of our workflow templating I can put in whatever but what I think is the right config nomenclature doesn't work, if that makes sense. For 3DVar cost type this wouldn't have been an issue.

This config doesn't work:

```yaml
geometry:
  mom6_input_nml: soca/input.nml
  fields metadata: soca/fields_metadata.yaml
  geom_grid_file: INPUT/soca_gridspec.nc
background:
  date: '2023-07-02T00:00:00Z'
  read_from_file: 1
  basename: ./
  ocn_filename: MOM6.res.20230702T000000Z.nc
  ice_filename: cice.res.20230702T000000Z.nc
  state variables:
  - sea_ice_area_fraction
  - sea_ice_thickness
  - sea_ice_snow_thickness
  - sea_water_salinity
  - sea_water_potential_temperature
  - sea_surface_height_above_geoid
  - sea_water_cell_thickness
  - ocean_mixed_layer_thickness
  - sea_water_depth
increment:
  date: '2023-07-02T12:00:00Z'
  basename: ./
  ocn_filename: ocn.soca2cice_new.incr.2023-07-02T12:00:00Z.nc
  ice_filename: ice.soca2cice_new.incr.2023-07-02T12:00:00Z.nc
postprocess ice:
  ncat: 5
  ice_lev: 7
  sno_lev: 1
  cice restart:
    input: iced.res.20230702T000000Z.nc
    output: iced.res.20230702T000000Z.nc
  itd:
    category bounds:
    - 0.0
    - 0.6445072
    - 1.391433
    - 2.470179
    - 4.567288
    - 9.333887
```

This one works
```yaml
geometry:
  mom6_input_nml: soca/input.nml
  fields metadata: soca/fields_metadata.yaml
  geom_grid_file: INPUT/soca_gridspec.nc
background:
  date: '2023-07-02T00:00:00Z'
  read_from_file: 1
  basename: ./
  ocn_filename: MOM6.res.20230702T000000Z.nc
  ice_filename: cice.res.20230702T000000Z.nc
  state variables:
  - sea_ice_area_fraction
  - sea_ice_thickness
  - sea_ice_snow_thickness
  - sea_water_salinity
  - sea_water_potential_temperature
  - sea_surface_height_above_geoid
  - sea_water_cell_thickness
  - ocean_mixed_layer_thickness
  - sea_water_depth
increment:
  date: '2023-07-02T00:00:00Z'
  basename: ./
  ocn_filename: ocn.soca2cice_new.incr.2023-07-02T12:00:00Z.nc
  ice_filename: ice.soca2cice_new.incr.2023-07-02T12:00:00Z.nc
postprocess ice:
  ncat: 5
  ice_lev: 7
  sno_lev: 1
  cice restart:
    input: iced.res.20230702T000000Z.nc
    output: iced.res.20230702T000000Z.nc
  itd:
    category bounds:
    - 0.0
    - 0.6445072
    - 1.391433
    - 2.470179
    - 4.567288
    - 9.333887
```

shlyaeva:
@Dooruk that error makes sense. I've just checked our workflow, and we basically trick it into thinking the background and increment are at the same date (the same way you do in the second yaml snippet) since there are no checks on dates in soca file reads. If that's good enough, we can just keep using this yaml trick. Or, we could also add some yaml flag to ignore the date difference, and update the increment time to that of the background if that flag is set. What do you think?

Dooruk:
> @Dooruk that error makes sense. I've just checked our workflow, and we basically trick it into thinking the background and increment are at the same date (the same way you do in the second yaml snippet) since there are no checks on dates in soca file reads. If that's good enough, we can just keep using this yaml trick. Or, we could also add some yaml flag to ignore the date difference, and update the increment time to that of the background if that flag is set. What do you think?

I guess it's fine to keep using this trick if `increment` section with date is solely used for this method. I wasn't sure if it's a bug or expected behavior 😄  I checked `AnalysisPostproc` but that doesn't have increment dates either.

guillaumevernieres:
> > @Dooruk that error makes sense. I've just checked our workflow, and we basically trick it into thinking the background and increment are at the same date (the same way you do in the second yaml snippet) since there are no checks on dates in soca file reads. If that's good enough, we can just keep using this yaml trick. Or, we could also add some yaml flag to ignore the date difference, and update the increment time to that of the background if that flag is set. What do you think?
> 
> I guess it's fine to keep using this trick if `increment` section with date is solely used for this method. I wasn't sure if it's a bug or expected behavior 😄 I checked `AnalysisPostproc` but that doesn't have increment dates either.

It's not a trick, we've assumed that the linear model is the identity so the increment is the same throughout the window.

PR #1247, opened 2026-06-10
Update JEDI CI action with build cache

eap:
Update to jedi-ci v2 with caching

run-ci-on-draft = true

Issue: https://github.com/JCSDA-internal/jedi-ci/issues/77

### DIFF
--- a/.github/workflows/start-jedi-ci.yaml
+++ b/.github/workflows/start-jedi-ci.yaml
@@ -38,7 +38,7 @@ jobs:
           aws-region: us-east-2
 
       - name: Run JEDI CI
-        uses: JCSDA-internal/jedi-ci@develop
+        uses: JCSDA-internal/jedi-ci@v2
         with:
           container_version: 'latest'
           unittest_tag: 'soca'
--- a/.github/workflows/start-jedi-ci.yaml
+++ b/.github/workflows/start-jedi-ci.yaml
@@ -42,8 +42,6 @@ jobs:
         with:
           container_version: 'latest'
           unittest_tag: 'soca'
-          test_dependencies: gsw ioda ioda-data oasim oops saber ufo ufo-data vader jedi-model-data
           test_script: run_tests.sh
-          target_repo_dir: target_repository
           bundle_branch: develop
           jedi_ci_token: ${{ steps.generate-token.outputs.token }}


travissluka (approved):

shlyaeva:
Is the expectation here that only one set of tests would run, and it would contain 3196 tests?

ashley314 (approved):

eap:
@shlyaeva - yes this new behavior is expected.

We still run unit tests first, then integration tests, but they are reported togeather in the same github "Check Run" object. If unit tests fail, the test finishes early and the GitHub Check Run reports failure for the unit tests.

I made this change for two reasons:
* __it matches cdash behavior:__ CI uses the same cdash test identity for both test runs resulting in them being combined in the cdash view. With 2 check-runs in the older version of CI, it was confusing for users reviewing "unit" results of passing tests since you would find combined unit/integration results (which could include failing integration tests even if the unit test reported success, and in all cases would show more tests than expected).
* __cached builds are much faster:__ Part of the reason that I reported two result objects in the old CI was that the 2-stage build was _exceptionally slow_ and I wanted to report _something_ before the whole 2-hour test run finished. Now that all the building is done rapidly, and before any tests run, unit tests are generally quite fast (ufo notwithstanding) so the time delta between unit tests and integration tests is significantly smaller.

shlyaeva:
@eap OK, sounds good! One way to speed things up even more would be to run only this repos tests + everything downstream (no need to run anything upstream; for model interface repos this should save quite a bit of time). But this is already altogether much faster, thank you for doing this!

shlyaeva (approved):

shlyaeva:
@eap one more question: do we run CI with oasim?

eap:
@shlyaeva - yes; here's the configure. Actually, I just added you to the [oasim PR](https://github.com/JCSDA-internal/oasim/pull/9) since I'm short on reviews.

`ecbuild -Wno-dev -DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache -DBUILD_GSIBEC=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCDASH_OVERRIDE_SYSTEM_NAME=intel-Container -DCDASH_OVERRIDE_SITE=AWSBatch -DCDASH_OVERRIDE_GIT_BRANCH=0 -DCTEST_UPDATE_VERSION_ONLY=FALSE -DBUILD_IODA_CONVERTERS=ON -DBUILD_PYIRI=ON -DBUILD_ROPP=ON -DBUILD_RTTOV=ON -DBUILD_OASIM=ON -DCMAKE_DISABLE_FIND_PACKAGE_bufr_query=ON /workdir/bundle`

